PR #66, opened 2022-04-08
New split_history and vector2matrix utils

gaffarelj:
This branch implements the two utilities that @Marchesini010126 suggested in [this issue](https://github.com/tudat-team/tudat-space/issues/65).

I exposed `getIntegratedTypeAndBodyList` from `tudat::propagators` as `tudatpy.kernel.numerical_simulation.get_integrated_type_and_body_list` to make the `split_history` function work (as we need to know what bodies are propagated, and most importantly what type of states). I didn't make docstrings for it, as I don't think it should be too visible to tudatpy users.

The docstrings of the two new functions can be found directly in the function definitions. I will push directly to `tudat-multidoc` to make them active in the API docs when this pull request is approved.

### DIFF
--- a/tudatpy/util/_support.py
+++ b/tudatpy/util/_support.py
@@ -259,10 +259,7 @@ def split_history(state_history: Dict[float, np.array], propagator_settings: pro
     -----------
     state_history_book : Dict[str,[Dict[float, numpy.ndarray]]]
         Dictionnary containing the name of the propagated body as key, and the state history as value.
-    """
-    # Define the state size of the implemented state types.
-    state_type_infos = {1: 6, 2: 7, 3: 1} # Translational (6), Rotational (7), Mass (1)
-    
+    """    
     # Get the propagated state types and names of the propagated bodies from the integrator settings.
     integrated_type_and_body_list = numerical_simulation.get_integrated_type_and_body_list(propagator_settings)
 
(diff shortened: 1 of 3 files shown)


tudatpy/util/_support.py
```diff
@@ -235,4 +237,85 @@ def pareto_optimums(points: list, operator:Union[None,List[Union[min,max]]]=None
...
+    propagator_settings : tudatpy.kernel.numerical_simulation.propagation_setup.propagator.PropagatorSettings
+        Settings used for the propagation.
+
+    Returns
+    -----------
+    state_history_book : Dict[str,[Dict[float, numpy.ndarray]]]
+        Dictionnary containing the name of the propagated body as key, and the state history as value.
+    """
+    # Define the state size of the implemented state types.
+    state_type_infos = {1: 6, 2: 7, 3: 1} # Translational (6), Rotational (7), Mass (1)
```
DominicDirkx, line 264: I propose we replace this with a C++ function. I know for sure that this is implemented somewhere

gaffarelj: Indeed, there is the tudat.propagators.getSingleIntegrationSize function that can be used for this. I will expose it to python and replace the dictionary that I hard-coded into the util.

gaffarelj: I have implemented this change with 3789a15ef98aaabec417fd7ca0a67b832ad62ec9. Do I have the green light to merge this pull request?

PR #77, opened 2022-09-28
Feature/polyhedron

MiguelAvillez:
- Polyhedron gravity and shape model
- Gravitational potential and laplacian as dependent variables
- Update to ellipsoid interface


tudatpy/kernel/expose_numerical_simulation/expose_environment_setup/expose_gravity_field_setup.cpp
```diff
@@ -131,16 +157,63 @@ namespace gravity_field {
               get_docstring("spherical_harmonic").c_str()
               );
 
-        m.def("spherical_harmonic_triaxial_body",
```
DominicDirkx, line 134: I understand your reasoning for modifying this name. But, I would like to keep the old function name as well. Could you re-add the old function name, with a deprecation warning? See custom_ephemeris vs. custom in /expose_numerical_simulation/expose_environment_setup/expose_ephemeris_setup.cpp for an example. The old function name does not need to be added to the API, but it should still continue to exist, allowing old code to continue to be used (for some time)

MiguelAvillez: In the tudat branch I still don't have deprecationWarnings.h file. Should I add the deprecation warning anyway (which won't compile) since it will be fine after merging?

DominicDirkx: I'll try merging your updates with the latest version, to also check for any conflicts, and take care of this along the way :)

DominicDirkx:
I've made a few minor changes to some of the inputs (see tudat pull request and merger) and some small naming updates. It's now merged into develop. If you'd like to make further modifications, let me know!

PR #112, opened 2023-09-28
Creating Ephemerides settings using JPL Horizons

tristandijkstra:
This PR adds an interface to JPL's Horizons System. For now, the vectors API has been implemented, building upon the features of Astroqueries' implementation. Other Horizons features will be added in future. An issue will be opened soon where planned features will be listed.  For now, the new HorizonsQuery (single-body) and HorizonsBatch (multi-body) classes feature:

- Retrieval of Horizons System generated state vectors for any time and any body (including spacecraft) available in the Horizons System.
- Creation of tabulated ephemeris settings for any of these bodies for use in Tudat.
- An extended query option which bypasses Horizon's output limit by automatically creating subqueries.
- Extensive input handling and processing to enable simple addition of future features.

This PR is related to the new MPC interface (https://github.com/tudat-team/tudatpy/pull/110), both included in Tudat's new data submodule. A future PR in the examples repository covering the advanced use of the MPC interface will showcase the features of this PR.


DominicDirkx:
Small comment on this, there's an open pull request:

https://github.com/tudat-team/tudatpy/pull/116

which allows for the merging of native Python and C++ code into a single submodule. Once this is merged and tested, I'd like to extend this pull request with one additional function in this submodule: https://py.api.tudat.space/en/stable/ephemeris.html#functions (named, for instance `jpl_horizons`), which then calls the `HorizonsQuery.create_ephemeris_tabulated` function.

tudatpy/data/horizons.py
```diff
@@ -0,0 +1,962 @@
...
+            res_list.append(res)
+
+        raw = vstack(res_list)
+
+        # retrieve the object's full name from Horizons
+        self._target_full_name = raw["targetname"][0]
+
+        return raw
+
+    def carthesian(
```
DominicDirkx, line 712: -> cartesian :)

DominicDirkx:
This all looks very good! I found one spelling error in a function name, but other than that, I think this can be merged (once #110 is merged)!

When this and #116 are succesfully merged, we can add the function to  https://py.api.tudat.space/en/stable/ephemeris.html that I mentioned in my previous post (but we can merge this PR before doing so).

alopezrivera:
Hi! Nice feature! When performing optimizations with many propagations it would be convenient to avoid re-downloading the JPL Horizons ephemerides as many times.
Would it be possible to serialize/save in some other way the data obtained from JPL Horizons to load it back when running  further simulations using the same ephemerides?

tristandijkstra:
Hi Antonio,

This would indeed be a useful feature. The first thing that comes to mind is saving it to a csv/parquet/etc and then having a query.load method. But this feels a bit cheap. Maybe it would be best to do this at the ephemeris level?

It could work to directly cache start, end, timestep queries, for other types of queries this might be difficult. I'm open to suggestions on the format of saving the cached files, maybe you have some ideas? Also, this would be in a subsequent PR.

alopezrivera:
Hi Tristan, I agree it would be best to do it at the pehemeris level. As to file formats, in ml it's pretty common to pickle data and I really like that way of doing things. I haven't worked with more specific file formats for this case. Maybe @DominicDirkx would have a suggestion?

PR #115, opened 2023-10-03
Semantic dependent variable retrieval: SimulationVariableDict and result2dict

alopezrivera:
# TL;DR

This pull requests implements semantic dependent variable retrieval using dependent variable settings objects, that is, to obtain time histories (`epoch`: `value` maps) of dependent variables with ease, where the `value` of the dependent variable at each epoch will be either numeric, vectorial, a matrix, etc.:

```
total_acceleration_history = dep_vars_dict[
    propagation_setup.dependent_variable.total_acceleration("Delfi-C3")
] 
```

**Subject to approval and review**

- API: parent module and name of `result2dict` function

**Table of contents**

1. Usage
2. In-depth explanation

---

# Usage

This pull request implements the pure Python `SimulationVariableDict` dictionary-like class and `result2dict` function, which takes a `numerical_simulation.SingleArcSimulator` and returns a `SimulationVariableDict`.

Both live in the `tudatpy.utils` module. Importing `result2dict`  is done as follows:

```
from tudatpy.utils import result2dict
```

The goal of `result2dict` and the `SimulationVariableDict` class is to make dependent variable time history retrieval semantic and straight-forward:

```
total_acceleration_history = dep_vars_dict[
    propagation_setup.dependent_variable.total_acceleration("Delfi-C3")
] 
```

See the complete example below. As you can see, we can use as keys either the

* Original dependent variable settings object (`dependent_variables_to_save[0]`),

* Or a newly created dependent variable settings object (`propagation_setup.dependent_variable.total_acceleration("Delfi-C3")`)

```
# General imports
# ...

# Tudat imports
# ...
from tudatpy.utils import result2dict

# Propagation setup
# ...

# Create simulation object and propagate the dynamics
dynamics_simulator = numerical_simulation.create_dynamics_simulator(
    bodies, propagator_settings
)

# Create SimulationVariableDict
dep_vars_dict = result2dict(dynamics_simulator)

# Retrieve the time history (in `dict[epoch: value]` form) of the total acceleration experienced by Delft-C3
total_acceleration_history = dep_vars_dict[
    # This can be done using either the `SingleAccelerationDependentVariableSaveSettings`
    # corresponding to this dependent variable
    dependent_variables_to_save[0]
]
total_acceleration_history = dep_vars_dict[
    # Or a newly created one
    propagation_setup.dependent_variable.total_acceleration("Delfi-C3")
]
```

# In-depth explanation

## `result2dict`

[...]

PR #120, opened 2023-10-29
Dependent variable dictionary - definitive and inside hybrid kernel

alopezrivera:
Pull request building on https://github.com/tudat-team/tudatpy/pull/115 (which can be closed upon merging this one).

## Improvements

**Naming**

Changed to indicate what the class and function do as of today:

- `SimulationVariableDict` -> `DependentVariableDictionary`
- `result2dict` -> `create_simulation_variable_dictionary`

**NumPy array conversion**

New method `DependentVariableDictionary.asarray` implemented, allowing users to easily retrieve the time history of each dependent variable as a NumPy array. Paired with the `DependentVariableDictionary.time_history` attribute this streamlines away the data juggling that was required before to plot dependent variables.

The new method works like this: `dep_vars_dict.asarray(<key>)`, where `<key>` can be either a `SingleDependentVariableSettings` object (or children instance), or the string ID of a `SingleDependentVariableSettings` object. That is, the keys used by this method are the same we use to retrieve dependent variable histories in dictionary form from the `DependentVariableDictionary`.

See the example below as well.

```
# Create semantic dependent variable history
dep_vars_dict = create_dependent_variable_dictionary(dynamics_simulator)

# Retrieve dependent variable history as NumPy array of shape (n, A, B)
a_scalar_dependent_variable = dep_vars_dict.asarray(dependent_variables_to_save[0])

# Plot
plt.plot(
    dep_vars_dict.time_history,
    a_scalar_dependent_variable
)
```

where `(A, B)` is the shape of the dependent variable. The dependent variables are `np.squeeze`d to remove singleton/trivial dimensions, so in the case of a scalar variable `(n, A, B)` shape would simply become `(n,)`.

**Testing** ([dependent_variable_dict](https://github.com/alopezrivera/tudatpy/compare/dependent_variable_dict) ([tudat-team/tudatpy#115](https://github.com/tudat-team/tudatpy/pull/115)))

Test added to the tudatpy tests which tests:

- Retrieval of dependent variables
- `DependentVariableDictionary.asarray` method to obtain NumPy arrays from dependent variable history dictionaries
- Saved histories are equal to those obtained from the `SingleArcSimulationResults.dependent_variable_history`
- Compatibility of scalar dependent variable dictionaries with `tudatpy.util.result2array`
- Dimensions of vectorial dependent variables


DominicDirkx:
Code merged into develop

PR #128, opened 2023-11-08
Kernel autocompletions

alopezrivera:
This pull request implements autocompletions for all members -functions, objects and variables (which are also objects in Python)-  of all modules of the tudat kernel.

To explain how this works, consider a tudat kernel module which is exposed through a Python module that wraps it. The Python module lives inside tudatpy, has the same name as the kernel module, and has inside an `__init__.py` with the following statement: `from tudatpy.kernel.<kernel_module_name>import *` -this allows us to import all members of the kernel module from its Python wrapper-.

For each such exposed kernel module, this pull request does the following:
1. Next to the Python module's `__init__.py`, a new file is created: `_import_all_kernel_members.py`
2. `_import_all_kernel_members.py` explicitly imports all members (functions and objects) of the kernel module, like this: `from tudatpy.kernel.<module_name> import <member 1>, <member 2>, ...`
3. In the module's `__init__.py`, we import everything from `_import_all_kernel_members.py`. This way, all the _explicitly imported_  kernel members are accessible from the Python module wrapping it: autocompletion engines will scan this, and detect all module members for autocompletion

The reason we do this with two files (`__init__.py` and `_import_all_kernel_members.py`) is that this way users **can change the** `__init__.py` (which they may well want to do, as it is the `__init__.py` of a _hybrid_ C++/_Python_ module), and we can **freely automatically generate** the `_import_all_kernel_members.py` **overwriting the existing ones**, as these files will **never be manually changed**.

Basically, this way we allow users to change `__init__.py`, and keep the liberty to blast away old versions of all `_import_all_kernel_members.py`s when the kernel changes.


DominicDirkx:
Hi Antonio, this is great, thanks so much for adding this! As a followup to this, could you document the workflow of this (as well as the hybrid kernel stuff) on our developer docs:

https://tudat-developer.readthedocs.io/en/latest/
source: https://github.com/tudat-team/tudat-developer-docs

PR #129, opened 2023-11-30
Adding NSDC code

FDahmani:
Initial pull request for the NSDC reading code. Please check the function names, whether they are clear and not confusing in the big picture of tudatpy. Make sure to load the necessary kernels, in this case 'jup344.bsp'. You can verify with the following file:
[ji0003.txt](https://github.com/tudat-team/tudatpy/files/13646037/ji0003.txt)
You should get the following results:
Means: [ 4.82986612e-07 -4.81475772e-07] Stddevs: [1.54415899e-06 1.06032470e-06]


DominicDirkx:
@FDahmani  Could you add the script that should be run, from which I would get this output?

PR #132, opened 2023-12-09
exposed clock noise functions

kimonito98:
exposed clock noise functions for tudatpy

### DIFF
--- a/tudatpy/CMakeLists.txt
+++ b/tudatpy/CMakeLists.txt
@@ -60,6 +60,7 @@ YACMA_PYTHON_MODULE(kernel
         kernel/expose_astro/expose_two_body_dynamics.cpp
         kernel/expose_astro/expose_fundamentals.cpp
         kernel/expose_astro/expose_polyhedron_utilities.cpp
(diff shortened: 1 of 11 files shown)


DominicDirkx:
Hey, thanks for adding all this! If I understand correctly, this is to be used with this: https://github.com/tudat-team/tudat/tree/feature/clock_noise_models right?
I think this is also code that can be merged soon, but we need to resolve one thing: the proper inclusion/finding of FFTW3

kimonito98:
Yes indeed (I sent you the messages on slack). Linking fftw3 was quite easy, it requires specifying the path to the library in cmake. Lethal it could be included as a necessary package in the environment.yalm? You know best On 11 Dec 2023, at 11:56, DominicDirkx ***@***.***> wrote:﻿
Hey, thanks for adding all this! If I understand correctly, this is to be used with this: https://github.com/tudat-team/tudat/tree/feature/clock_noise_models right?
I think this is also code that can be merged soon, but we need to resolve one thing: the proper inclusion/finding of FFTW3

PR #134, opened 2023-12-15
Remove deprecated functions

soufyanR1:
They seem to interfere with documentation creation as the parameters do now show up correctly here: https://py.api.tudat.space/en/latest/aerodynamic_coefficients.html#tudatpy.numerical_simulation.environment_setup.aerodynamic_coefficients.constant

The functions for which this happens aligns exactly with the functions that have been deprecated in the code below.


DominicDirkx:
I've changed the target of the merge to the develop branch. There, I think these deprecated functions can now be safely removed.

PR #138, opened 2024-02-15
Remove deprecated (?) direct_tidal_dissipation_acceleration

soufyanR1:


DominicDirkx:
Hi Soufyan, the tidal acceleration one should remain (but we should come up with a different name). The other ones can indeed be removed

PR #143, opened 2024-04-08
Fixed an invalid Name to make the build successful

LeonardoPohl:
I had issues building the whole project with the develop branch and noticed there was a naming issue.


DominicDirkx:
Implemented in a separate push just before this pull request. Nevertheless, thanks for indicating this and creating the fix!

Issue #177, opened 2024-10-04
Implement a nightly bump version automation workflow


niketagrawal:
Similar to https://github.com/tudat-team/tudat/issues/232, we need a GitHub action workflow that triggers nightly at a scheduled time and performs the following actions if there is a new commit(s) on the develop branch of tudatpy in the last 24 hours:

1. Bump the version number, create a tag, and push the tag and the bump version commit to the develop branch of the tudatpy repository
2. Bump the version number, create a tag, and push the tag and the bump version commit to the develop branch of the tudatpy repository

Currently, the above actions happen manually. When the develop branch of tudatpy is found to have reached a staged where a new version can be created, the tag and the version is created manually, and the version and tag is added on the tudatpy-feedstock repository, which then triggers the azure pipelines that creates the conda package. 

###NOTE
This workflow must be present in the master branch because scheduled workflows run only on the default branch, so the PR is opened towards master. This version does not make any changes to the master branch of tudat and tudat-feedstock though. This is controlled by specifying the checkout of develop branch in the workflow. This ensures that the bump version action will be performed on the develop branches.

niketagrawal:
This GitHub Actions workflow will require personal access token to bump version, create a tag, and push the tag to tudatpy-feedstock repository. Below are the steps to create a personal access token to use in a Github action:

1. Go to Github profile -> settings > Developer settings (bottom of the menu on the left corner) > Personal access tokens > tokens (classic).
2. Generate new token -> tokens classic. Create a new token by adding a description, for example,  'nightly bump version workflow'

This token must now be specified to use in the tudatpy repository.

1. Got to tudatpy repository -> settings > secrets and variables > actions > new repository secret
2. Name the secret for example, bump_version_nightly. Go to the personal access token created in previous step and copy the token code and paste it as the value of this secret. 
3. Mention this secret name in the bump version github action workflow.

yiquintero:
**Warnings for this workflow**

From the [GitHub documentation](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule), workflows that run on a trigger of type `schedule` (e.g. every night at 3 am) have the following constraints:

[...]

PR #178, opened 2024-10-04
Nightly bump version workflow

niketagrawal:
This GitHub actions workflow runs daily at 3:00 AM and does the following if there has been new commit(s) on the develop branch of tudatpy in the last 24 hours:

1. Bump dev version: dev field of the version number in the files `version` and `.bumpversion.cfg` will be bumped by 1. This will result in a new commit.
2. Create a tag with the above version, and push the tag and the commit to the tudatpy develop branch
3. Bump dev version in tudat-feedstock: dev field of the version number in recipe/meta.yml and .bumpversion.cfg will be bumped by 1. This will result in a new commit.
4. Create a tag with the above version, and push the tag and the commit to the tudatpy-feedstock develop branch

**This workflow must be present in the master branch because scheduled workflows run only on the default branch, so the PR is opened towards master. This version does not make any changes to the master branch of tudatpy and tudatpy-feedstock though.** This is controlled by specifying the checkout of develop branch in the workflow. This ensures that the bump version action will be performed on the develop branches.

This pull request should be merged after the changes in branch https://github.com/tudat-team/tudatpy-feedstock/tree/19-add-bumpversion-cfg are merged to the develop branch of tudatpy-feedstock, so the bumpversion operation can be performed successfully in tudatpy-feedstock.

### To do
- [ ] Create Personal access token as described [here](https://github.com/tudat-team/tudatpy/issues/177#issuecomment-2393655358) and fill in the secret name in the placeholder in the workflow.

### DIFF
--- a/.github/workflows/bump_dev_version_nightly.yml
+++ b/.github/workflows/bump_dev_version_nightly.yml
@@ -18,7 +18,7 @@ name: Bump dev version in tudatpy and tudatpy-feedstock if new commit in last 24
 
 on:
     schedule:
-        - cron: '0 1 * * *'  # This triggers the workflow daily at 3:00 AM CET. GitHub uses UTC time, so 1:00 AM UTC is 3:00 AM CET.
+        - cron: '0 5 * * *'  # This triggers the workflow daily at 3:00 AM CET. GitHub uses UTC time, so 1:00 AM UTC is 3:00 AM CET.
     workflow_dispatch:       # This facilitates manual triggering of the workflow from the Actions tab of the repository in case the scheduled run fails.
 
 
@@ -91,7 +91,7 @@ jobs:
           uses: actions/checkout@v4
           with:
             repository: tudat-team/tudatpy-feedstock
-            token: ${{ secrets.<add_secret_name> }}
+            token: ${{ secrets.BUMP_VERSION_NIGHTLY }}
             ref: develop
 
         - name: Install bump2version
--- a/.github/workflows/bump_dev_version_nightly.yml
+++ b/.github/workflows/bump_dev_version_nightly.yml
@@ -1,0 +1,116 @@
+# This workflow is triggered at 3:00 AM CET everyday and checks if there has
+# been a commit in the last 24 hours on the develop branch of tudatpy repository.
+
+# If there has been a commit, it bumps the version in the develop branch of
+# the tudatpy repository, creates a tag with the version numbers and pushes
+# both the commit and the tag to the repository, and then bumps the version in
+# the develop branch of the tudatpy-feedstock repository and pushes the commit
+# and the tag to the repository.
+
+# This workflow uses personal access tokens to push the changes to the repository. The personal access token is stored as a secret in the repository.
+
+# This workflow file must be in the master (default branch) of the tudatpy repository because scheduled workflows are only triggered on the default branch.
+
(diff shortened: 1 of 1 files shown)


yiquintero:
**Warnings for this workflow**

From the [GitHub documentation](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule), workflows that run on a trigger of type `schedule` (e.g. every night at 3 am) have the following constraints:

> In a public repository, scheduled workflows are automatically disabled when no repository activity has occurred in 60 days. For information on re-enabling a disabled workflow, see "[Disabling and enabling a workflow](https://docs.github.com/en/enterprise-server@3.12/actions/using-workflows/disabling-and-enabling-a-workflow#enabling-a-workflow)."

This warrants no action from our side, but it's good to keep it in mind. Tudat and tudatpy are both very active repos so I don't expect a problem.

> When the last user to commit to the cron schedule of a workflow is removed from the organization, the scheduled workflow will be disabled. If a user with write permissions to the repository makes a commit that changes the cron schedule, the scheduled workflow will be reactivated. Note that, in this situation, the workflow is not reactivated by any change to the workflow file; you must alter the cron value and commit this change.

Since @niketagrawal and myself will leave the project at some point, it would be best for @DominicDirkx to make a change to the [cron trigger in the workflow file](https://github.com/tudat-team/tudat/blob/58ae2e24ec40f327e073d7c5c8af337bab0e1d11/.github/workflows/bump_dev_version_nightly.yml#L21) to prevent the workflow from being disabled.

.github/workflows/bump_dev_version_nightly.yml
```diff
@@ -0,0 +1,116 @@
...
+# This workflow file must be in the master (default branch) of the tudatpy repository because scheduled workflows are only triggered on the default branch.
+
+# Reference for the logic to check if there has been a commit in the last 24 hours: https://stackoverflow.com/questions/4089430/how-can-i-determine-if-a-git-commit-is-more-recent-than-another-commit
+
+
+name: Bump dev version in tudatpy and tudatpy-feedstock if new commit in last 24 hours
+
+on:
+    schedule:
+        - cron: '0 1 * * *'  # This triggers the workflow daily at 3:00 AM CET. GitHub uses UTC time, so 1:00 AM UTC is 3:00 AM CET.
```
yiquintero, line 21: @DominicDirkx please change this line by altering the time at which the workflow is triggered. For example, change the execution time to 2:00 AM UTC (4am NL time)
`cron: '0 2 * * *'`

.github/workflows/bump_dev_version_nightly.yml
```diff
@@ -0,0 +1,116 @@
...
+          shell: bash
+          run: |
+                sed -i 's/build = ".*" %}/build = "0" %}/' recipe/meta.yaml
+                git add recipe/meta.yaml
+                git commit --amend --no-edit
+
+        - name: Push tag to tudatpy-feedstock
+          run: git push origin --follow-tags
+          env:
+              GITHUB_TOKEN: ${{ secrets.<add_secret_name>}}
```
yiquintero, line 116: @DominicDirkx replace <add_secret_name> with the name of your personal access token. See the PR notes on how to create one.

yiquintero (changes requested):

PR #205, opened 2025-01-15
Added missing docs

DominicDirkx:
Added a number of docs missing for custom state and other propagator settings

### DIFF
--- a/docs/source/propagator.rst
+++ b/docs/source/propagator.rst
@@ -34,6 +34,8 @@ Functions
 
    mass
 
+   custom_state
+
    multitype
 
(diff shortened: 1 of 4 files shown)


DominicDirkx:
@larshinueber @luigigisolfi  Hi, I added two new entries to the API docs (or should have...), but when I build it locally, I get the entry 'No documentation found' in the docs. If you could have a quick look and see if you can easily spot my mistake, that'd be great. Since you have boht worked on adding new docs recently (?), maybe one of you can see what I forgot. Don't spend too much time, this is just a request for a 'sanity check'

PR #218, opened 2025-01-29
Docstring updates

larshinueber:
- Add BodyListSettings docs, some more BodySettings attributes
- Add data docstrings from io module
- Reorder interpolator exposure so return types can be inferred by pybind


DominicDirkx:
Thanks again :) We're really getting there in making the API comprehensive!

PR #221, opened 2025-01-31
Feature/github_action

valeriof7:
Add a new github action that will format the cpp code when a pull request is merged. 

Note that since the project is split into multiple github repositories, we need to include the `.clang-format` setup file in the root of each github repository.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,57 @@
+---
+Language: Cpp
+BasedOnStyle: Google
+IndentWidth: 4
+TabWidth: 4
+UseTab: Never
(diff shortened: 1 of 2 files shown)


DominicDirkx:
@valeriof7 Could you remind me if any action is needed before this PR can be merged (is it tested?) Perhaps there was an action on me, but I forgot :)

valeriof7:
It took me a while to remember. This is for formatting the CPP code of tudatpy using Clang. I don't think we need to merge this at the moment, but you can do it without problem in case. It simply adds the Clang formatter settings file in the directory root.

PR #241, opened 2025-02-19
Minor fixes

larshinueber:
- Move enums and classes to beginning of `time_conversion` module so they can be resolved in function definitions
- Remove duplicate exposure of `keplerian_to_cartesian`
- Add note to `spice.get_body_gravitational_parameter` as distinction to default models
- Fix docstring and argument type of first derivative data in interpolators


DominicDirkx:
Once again, thanks for the modifications and corrections!

PR #243, opened 2025-03-03
Feature: add TNF files reader

valeriof7:
Add the ability to read and process TNF tracking files in TUDAT.

## Details
The TRK-234 (or TNF) file reader functionality is mostly implemented in Python and relies on the [pytrk234](https://github.com/NASA-PDS/PyTrk234) Python library, which must be included in the tudat conda environment. Please note the installation instructions for this library as it differs from the normal pip installation.
Currently only 2-3 way doppler and sequential range are supported. The companion [tudat PR](https://github.com/tudat-team/tudat/pull/294) updates the additional settings for sequential_range and fixes the conversion factor from range unit to meter. Note that this implementation takes into account the current ObservationCollection structure, so a singleObservationSet is created for each range data point.
A unit test for the new module is also included.


DominicDirkx:
Really great to have this! If I understand correctly, the installation requires us to download the pytr234 repo, and then pip install it, is that correct? Ideally, we'd find a way to incorporate this into out environment.yaml files, so the steps a user takes to install don't change. Do you have any ideas on how to do this? If not, go ahead and open a discussion in #development on Slack, maybe some others will

valeriof7:
We would need to run the `git clone' command somehow. I'll do a little research and let you know what I find. I have already posted on slack, but no one has come up with a solution yet.

DominicDirkx:
Could you post a link to this PR there to restart the discussion?

DominicDirkx:
@valeriof7 Could you check with @alfonsoSR (if you hadn't yet) if his changes to the structure are compatible with what you did?

Could you put the list of actions (PRs, updates of docs) needed to merge all TRK234 documentation here?

alfonsoSR:
@DominicDirkx I checked with @valeriof7, and it seems that the changes are already compatible with the new structure

valeriof7:
I confirm that commit [7dbbe6e](https://github.com/tudat-team/tudatpy/pull/243/commits/7dbbe6e38b90d2232e0665e8ec49b2dff2393068) already addressed the integration with the new structure. Also the documentation was already Sphinx compatible. Commit [40ab51f](https://github.com/tudat-team/tudatpy/pull/243/commits/40ab51fd5080aa996dedb65d96de1907c517da96) includes a code snippet with an example.

DominicDirkx:
Perfect! I see there's two merge conflicts, could you resolve these, and (in this PR) post a list of all other actions that need to be taken (modifications to feedstock, updates to tudat-bundle docs) that need to be done in relation to this PR, also indicating whether (i) these changes need to happen before merging this PR to prevent breaking anything (ii) these changes can happen after merging this PR (e.g. updating docs)

DominicDirkx:
Compiling locally without the pytrk234 installed and then running the tests gives:

ImportError while importing test module '/home/dominic/Tudat/tudat-bundle/tudat-bundle/tudatpy/tests/test_data.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/home/dominic/miniconda3/envs/tudat-bundle/lib/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
test_data.py:1: in <module>
    from tudatpy import data
/home/dominic/miniconda3/envs/tudat-bundle/lib/python3.11/site-packages/tudatpy/data/__init__.py:4: in <module>
    from .processTrk234 import Trk234Processor
/home/dominic/miniconda3/envs/tudat-bundle/lib/python3.11/site-packages/tudatpy/data/processTrk234/__init__.py:1: in <module>
    from .processor import Trk234Processor
/home/dominic/miniconda3/envs/tudat-bundle/lib/python3.11/site-packages/tudatpy/data/processTrk234/processor.py:1: in <module>
    import trk234
E   ModuleNotFoundError: No module named 'trk234'

for each test. In other words, the package *needs* to be installed for the library to be used. Since this is the case, we need a way to automatically install the trk234 package, also when using the tudat-bundle

valeriof7:
I tested the updated meta.yaml file using Docker locally. It should work, but I agree that an additional test is always a good idea.
What about `tudat-space`? Isn't that our repository for the documentation?

PR #248, opened 2025-03-07
Feature/radio telescopes exposure and docstrings

luigigisolfi:
This is the fourth of some upcoming pull requests from my side. 
It exposes and documents the following functions: 'evn_stations', 'radio_telescope_stations'.


luigigisolfi:
Hello @DominicDirkx , this Pull Request contains all the changes related to:

- mission data downloader
- new functions exposures (radio telescope stations, evn stations, cartesian to geodetic coords) + related docstrings
- addition of a new horizons wrapper function for azimuth/elevation retrieval

I have reviewed these changes and updated them accordingly. Everything should be fine now, and ready to be merged to developed. 

Ciao
Luigi

tudatpy/data/horizons.py
```diff
@@ -997,6 +997,54 @@ def interpolated_observations(
 
         return res
 
+    def interpolated_az_el(
```
DominicDirkx, line 1000: Could you modify the function name to be 'verbose', something like station_observation_angles. The fact that they're interpolated (or azimuth/elevation) we could put in the docs (?)

DominicDirkx:
Cool, thanks! I've added one small question, other than that this looks ready!

tudatpy/data/horizons.py
luigigisolfi: I called it that way to be consistent with the function named "interpolated_observations" , already in place, that computes Ra & Dec. If you want, I can change both names, but I think that function is already used elsewhere in tudatpy-examples, so we would also need to change its calls :)

PR #258, opened 2025-03-22
Feature/nrlmsise fix

minervino96:
This pull request adds Python-side support for specifying geomagnetic activity conditions when creating an NRLMSISE00 atmospheric model.

### DIFF
--- a/tests/test_nrlmsise00.py
+++ b/tests/test_nrlmsise00.py
@@ -1,0 +1,42 @@
+# Load standard modules
+import os
+import pickle
+import numpy as np
+
+# Load Tudatpy modules
+from tudatpy.kernel import numerical_simulation
+from tudatpy.data import read_solar_activity_data, get_space_weather_path
+from tudatpy.kernel.astro.time_conversion import epoch_from_date_time_iso_string
+
+
+def test_nrlmsise00():
+
+    # Load the validation data obtained with the pymsis library. 
+    # The data contains the generated input data (date [iso string], longitude [deg], latitude [deg], altitude [meters]), 
+    # pymsis resulting densities [kg/m^3], and the corresponding space weather parameters (F10.7, F10.7a, and daily Ap) used 
+    # internally by pymsis for future references.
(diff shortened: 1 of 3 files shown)


minervino96:
### Summary of commit #d36ecab

This latest commit introduces a Python unit test for the NRLMSISE00 atmosphere model with pymsis validation data (in pickle format).

DominicDirkx:
Hi Andrea, could you check which parts of the above two files (with merge conflicts) you changed? It seems to think the whole file is modified, but I think it was just a few lines.

minervino96:
Hi Dominic,

I am not sure why the diff shows the entire file as modified.

For `expose_atmosphere_setup.cpp`, I have only added line 442, adjusted the documentation on line 447, and added documentation on lines 454 and 455.

As for `expose_estimated_parameter_setup.cpp`, I did not make any changes...

larshinueber:
Hi all, I think this is due to the changes made in #266. Had the same issue in #250, where I modified a file that no longer exists after the restructuring. I think the only way to resolve the conflicts is to manually integrate your changes in the new files and then delete the old file.

alfonsoSR:
I think @larshinueber is right. This is the kind of process I would have to go through every time someone pushed to develop while I was working on the new structure of `tudatpy`. I am sorry for the inconvenience, but you cannot make these kinds of changes without entering some sort of "transition period" during which things are a little bit annoying.

DominicDirkx:
I've merged the NRLMSISE update in Tudat, and when running your Python test with:

    NRLMSISE00 = numerical_simulation.environment_setup.atmosphere.NRLMSISE00Atmosphere(solar_weather_data,True,False,True)

(to account for the new inputs) I get a maximum density error of 0.006 (with some entries also at 1E-7). I'll merge the test with the relaxed tolerance, but let's discuss how to get the test data in line with the model (which now expects geodetic latitude and UTC time as input)

DominicDirkx:
This can now be closed, since updates to the NRLMSISE model are tracked in #280 and #283

PR #276, opened 2025-04-26
Make all kernel imports absolute

alfonsoSR:
In the current version of the `develop` branch, all the imports from `kernel` are relative
```python
# In numerical_simulation.environment.__init__.py
from ...kernel.numerical_simulation.environment import *
```
This was an arbitrary choice I made out of habit, but using absolute imports produces the same result
```python
from tudatpy.kernel.numerical_simulation.environment import *
```
Although they are practically equivalent, using absolute imports removes complexity from the stub generation pipeline. For this reason, this pull request makes all the relative kernel imports in tudatpy absolute.

### DIFF
--- a/src/tudatpy/astro/__init__.py
+++ b/src/tudatpy/astro/__init__.py
@@ -1,9 +1,1 @@
-from ..kernel.astro import (
-    element_conversion,
-    frame_conversion,
-    fundamentals,
-    gravitation,
-    polyhedron_utilities,
-    time_conversion,
-    two_body_dynamics,
-)
+from tudatpy.kernel.astro import *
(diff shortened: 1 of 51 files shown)


alfonsoSR:
Build is currently failing for Windows because latest version of `develop` also fails. Issue is unrelated to this branch

DominicDirkx:
This is ready to merge? The Windows build failures are (I think) due to an incompatibility of libraries when building tudat and tudatpy

alfonsoSR:
Yes, it should be ready to merge. The Azure build worked yesterday, none of the modifications I did today are relevant functionality-wise, and all the tests are passing locally. I can trigger another build, but it should be ready to merge.

PR #284, opened 2025-05-05
Backwards compatibility with `sys.path.append` import method

alfonsoSR:
Modified `CMakeLists.txt` with a macro to copy all the Python scripts in tudatpy to the build directory during build, as it was done before the new setup. This makes it possible to import `tudatpy` from a script by manually adding the path to the build at the beginning of it.

NOTE: The `install.py` and `uninstall.py` scripts are still the recommended methods to work with the library. This patch is just meant to ensure backwards compatibility.


alfonsoSR:
My bad, I wanted to merge into `develop`, not `master`

Issue #295, opened 2025-05-12
[F&E]: <Adding get_parameters for interpolators>


JonathanHasASecondName:
### Description:

Interpolators use values to interpolate between, eg. timesteps and temperatures. Outside of this range, extrapolation leads to major deviations. I would like the interpolator object to have a public attribute like get_original_values, that returns the original values that the interpolator interpolates between. This functionality is present in C++, but it is not supported in Python.

### Suggestions for implementation:

Add some public attributes to the python implementation.

PR #300, opened 2025-05-20
Expose dependent and independent variables of the interpolator

larshinueber:
As requested in #295, the dependent and independent variables of the interpolators should be exposed.
This PR adds them as read-only properties to the Interpolator classes.


DominicDirkx:
Again: many thanks!

PR #315, opened 2025-06-23
Small local developments

jo11he:
[ ] expose float version of dependent variable history attribute

[ ] interpolator settings and factory functions for float independent variables (these are in fact not necessary when the interpolator time type in scalarTypes.h is set to double, so could be excluded from PR)

[ ] exposed IAURotationModelSettings class 

With regards to the first two items: they may show in the api docs and I am unsure if we want that since many people using the API docs are unaware of the alternative time type altogether and thus these functions could create confusion.


DominicDirkx:
Merged :)

PR #336, opened 2025-07-11
Corrected exposure names

### LINKED ISSUE #197: Naming conflit in expose observation setup
There is a naming conflict in the `tudatpy/tudatpy/kernel/expose_numerical_simulation/expose_estimation_setup/expose_observation_setup.cpp` file between the observable type objects and the names of the Python functions used to create the associated observation model and define its settings. For example, the object for the dsn n-way range is usually called as follows 

```
tudatpy.numerical_simulation.estimation_setup.observation.dsn_n_way_range
```

While the associated observation settings are (currently) created using

```
tudatpy.numerical_simulation.estimation_setup.observation.dsn_n_way_Range(
                    current_link_definition, light_time_correction_list
                )
```

I think a refactoring of the naming for the observation settings (e.g. using something like create_n_way_range or n_way_range_settings) should be done.

DominicDirkx:
Closes #197

### DIFF
--- a/src/tudatpy/numerical_simulation/estimation_setup/observation/expose_observation.cpp
+++ b/src/tudatpy/numerical_simulation/estimation_setup/observation/expose_observation.cpp
@@ -247,10 +247,10 @@ Examples
             .value( "two_way_instantaneous_doppler_type", tom::ObservableType::two_way_doppler )
             .value( "n_way_averaged_doppler_type", tom::ObservableType::n_way_differenced_range )
             .value( "euler_angle_313_observable_type", tom::ObservableType::euler_angle_313_observable )
-            .value( "dsn_one_way_averaged_doppler", tom::ObservableType::dsn_one_way_averaged_doppler )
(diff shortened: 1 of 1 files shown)


src/tudatpy/numerical_simulation/estimation_setup/observation/expose_observation.cpp
```diff
@@ -1973,7 +1973,7 @@ Examples
            py::arg( "subtract_doppler_signature" ) = true,
            R"doc(No documentation found.)doc" );
 
-    m.def( "dsn_n_way_Range",
+    m.def( "dsn_n_way_nange",
```
larshinueber, line 1976: Hi Dominic, just stumbled across this. I believe this should be `dsn_n_way_range`?

valeriof7:
There is no `dsn_n_way_range_type` in the summer_camp_merge branch. The examples already work with `dsn_n_way_Range` which distinguishes between the two objects.

DominicDirkx:
But, the point is that the dsn_n_way_Range will no longer be used, and be replaced dsn_n_way_range

PR #349, opened 2025-07-18
Drag/Side/Lift direction scaling coefficients

nikemais:
This is the companion PR of #362.


DominicDirkx:
I don't think there is a #362 :)

DominicDirkx:
EDIT: never mind, I think I'm confused :) I see this is a merge into the normal develop branch

PR #350, opened 2025-07-18
Use Time type as standard independent variable of propagation/observation

valeriof7:
The C++ code for the time object is ready for refactoring in the monorepo.


DominicDirkx:
Code part of #351 (closing this sub-PR)

PR #366, opened 2025-07-22
Minor error resolution updates for API docs

larshinueber:
A few minor updates I encountered while working on #365. Together with #365, this should result in an error-free API docs build.

### DIFF
--- a/src/tudatpy/estimation/observable_models_setup/light_time_corrections/expose_light_time_corrections.cpp
+++ b/src/tudatpy/estimation/observable_models_setup/light_time_corrections/expose_light_time_corrections.cpp
@@ -320,63 +320,6 @@ Examples
 
      )doc" );
 
-    m.def(
-            "approximated_second_order_relativistic_light_time_correction",
-            []( const std::vector< std::string >& perturbingBodies ) {
-                // Force bending to always be true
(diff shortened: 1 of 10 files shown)


src/tudatpy/estimation/observable_models_setup/light_time_corrections/expose_light_time_corrections.cpp
```diff
@@ -320,63 +320,6 @@ Examples
 
      )doc" );
 
-    m.def(
-            "approximated_second_order_relativistic_light_time_correction",
```
DominicDirkx, line 324: why is this exposure deleted?

DominicDirkx:
Small question: can this PR be merged with #365 into a single one or is there a good reason to split them?

src/tudatpy/estimation/observable_models_setup/light_time_corrections/expose_light_time_corrections.cpp
larshinueber: Should have mentioned this in the PR body :) The exposure is duplicate, see lines 266ff in the same file

larshinueber:
Perfect, thanks!

PR #380, opened 2025-07-25
Fix/glibc error

larshinueber:
Update OS and conda versions of the readthedocs config to fix the GLIBC not found error described in https://github.com/readthedocs/readthedocs.org/issues/12341.

See https://py.api.tudat.space/en/fix-glibc-error/frame_conversion.html for the docs built from this branch.
PR is opened for the develop branch, to revive the develop/latest builds on RTD.


DominicDirkx:
Wow, that's so awesome! Thanks for fixing this

larshinueber:
Merging, let's see if this also works for develop :)

PR #378, opened 2025-07-24
port forked tudat develop (incl rtg_acceleration feature) to mono repo

jo11he:

PR #388, opened 2025-08-03
Feature/panel unit conversion_10

jo11he:
This is adding an on-the-fly addition to the penelled model loader to allow for user control over unit of input model. 
Feature branch name does not include the _10 extension, but is ported to match the monorepo structure. 
Equivalent commits for tudat/tudatpy develop in the bundle structure are implicit in RTG acceleration model PR (sorry bout that, I am learning!)


nikemais:
From the failing checks, it looks like the exposure of the function was not implemented correctly. I think this PR has only the changes in tudat.

It should be a quick fix!

jo11he:
Now redundant w.r.t. PR https://github.com/tudat-team/tudatpy/pull/378

PR #398, opened 2025-08-19
Make `build.py` compatible with star imports from submodules

alfonsoSR:
The current version of the stub generator in `build.py` only allows star imports from the kernel. Trying to import everything from a python script or a submodule results in an error that prevents stub generation.

We allow kernel star imports because the function of the kernel is to expose public functionality in tudat, but we decided to forbid them on the python side to force developers to be conscious about what they put in the API and what they hide from users.  I still think this is a good idea, but the way in which I implemented it does not allow for star imports from submodules, which is undesirable.

For example: having `tudatpy.estimation.observation_setup` divided into `biases`, `links`, `light_time_corrections`, and `model_settings`, is useful because it keeps the code organized and makes it easier for developers to understand where to find something by following the source tree of the library; but I think a user should be able to find anything related to the setup of observations in the top-level module. A solution would be to have this type of `__init__.py` file in `tudatpy/estimation/observation_setup`:
```python
from tudatpy.kernel.estimation.observation_setup import *    # Exposes top-level C++ functionality
from .biases import *
from .links import *
from .light_time_corrections import *
from .model_settings import *
```
This allows us to keep the low-level subdivision, but hide it from users.

This PR modifies the `build.py` script to make these types of imports technically possible, so that they can be implemented right away if we decide to.

### DIFF
--- a/build.py
+++ b/build.py
@@ -131,6 +131,40 @@ class BuildParser(argparse.ArgumentParser):
         return None
 
 
+def is_star_import(statement: ast.stmt) -> bool:
+    """Check if a statement is a star import
+
+    :param statement: The statement to be checked
+    :return: Whether the statement is a star import or not
+    """
+
+    # All star imports are of type: from X import *
+    if not isinstance(statement, ast.ImportFrom):
+        return False
+
+    # The number of imported elements must be equal to 1 (the star)
+    if len(statement.names) != 1:
+        return False
+
+    # The imported item must be *
+    if statement.names[0].name != "*":
+        return False
+
+    return True
+
+
+def module_has_init_stub(module_stubs_path: Path) -> bool:
+    """Checks if an __init__.pyi stub exists for a module
+
+    :param module_stubs_path: Path to the module directory in the stubs tree
+    :return: Whether __init__.pyi exists in the directory of the module in the stubs tree
+    """
+
+    # Path to __init__.pyi
+    init_stub_path = module_stubs_path / "__init__.pyi"
+    return init_stub_path.exists()
+
+
 class StubGenerator:
 
     # Default indentation length in pybind11-stubgen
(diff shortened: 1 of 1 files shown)


alfonsoSR:
@DominicDirkx 
> For the Pythonically deficient like myself: what does this mean exactly (which code works, and which doesn't right now?)

I didn't express myself super clearly. All the code works, but if you tried to add `from .<submodule> import *` to an `__init__.py`, and then run `build.py` you would get an error during stub generation.

> Does this mean that if you import observation_setup, everything from every underlying submodule (biases, links, etc.) would be automatically imported?

Yes, it would put all the contents of `biases`, `links`... under the `observation_setup` namespace

DominicDirkx:
> Does this mean that if you import observation_setup, everything from every underlying submodule (biases, links, etc.) would be automatically imported?
> 
> Yes, it would put all the contents of biases, links... under the observation_setup namespace

Ok, thanks! Is that desirable (and typical?) behaviour? And (more importantly): why?

> it would put all the contents of biases, links... under the observation_setup namespace

does being 'under' the observation_setup namespace (module?) just mean that the answer to my question was yes, or something else as well?

alfonsoSR:
> Ok, thanks! Is that desirable (and typical?) behaviour? And (more importantly): why?

Yes, it is very common. Some examples: [scipy.integrate](http://github.com/scipy/scipy/tree/main/scipy/integrate), [astropy.coordinates](https://github.com/astropy/astropy/tree/main/astropy/coordinates)

About why hiding the submodules from the user: I am not sure how to put it into words, but it makes things difficult to find. If I want to set up an estimation, I know that I need something for the observations, and something for the observation models, so looking for `estimation.observations` and `estimation.observable_models` is natural. If I type `observations.ObservationCollec` and the autocompletion does not show it, I just assume it might be in another submodule, not that there is a third level of directories.

> does being 'under' the observation_setup namespace (module?) just mean that the answer to my question was yes, or something else as well?

Yes, it just meant yes 😅

DominicDirkx:
Thanks, clear! This sounds like something to discuss at the next developer meeting. You can also post this on Slack to get some input from others already.

alfonsoSR:
@luigigisolfi yes. For example, from `tudatpy.dynamics.environment_setup` you would have access to the contents of atmosphere, shape_deformation, and all of those submodules

PR #402, opened 2025-08-20
Fix: Compatibility with type-annotated `__all__` statements

alfonsoSR:
When looking for __all__ statements, the previous version of the build.py script filters out statements that are not of ast.Assign type. Since `__all__: list = [...]` is of type ast.AnnAssign, the script fails to detect the statements if they are annotated, preventing stub generation.

The bug is fixed in this PR

### DIFF
--- a/build.py
+++ b/build.py
@@ -559,29 +559,34 @@ class StubGenerator:
 
         return None
 
-    def __retrieve_items_in_all(self, statement: ast.Assign) -> list[str]:
+    def __retrieve_items_in_all(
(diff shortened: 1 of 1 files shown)


DominicDirkx:
Thanks! Do you have a concrete example of something that did not work before but works now?

alfonsoSR:
Running `python3 build.py --stubs-only` on the latest version of `tudatpy/develop_10` fails for me. I am using `pybind11-stubgen 2.5.5` and `mypy 1.16.1`. These versions are relevant because the problem occurs during the post-processing of stubs generated by them.

Issue #396, opened 2025-08-18
[BUG]: Setting retransmitted frequency with transponder turnaround


kimonito98:
### Describe the issue:

I am unable to set the downlink frequency for a two-way measurement by means of the turnaround function. I adapt the Delfi-C3 example [https://docs.tudat.space/en/latest/examples/tudatpy-examples/estimation/covariance_estimated_parameters.html](url) to simulate two-way range observables with:

```
# set transmitter frequency station
bodies.get_body("Earth").get_ground_station("TrackingStation").set_transmitting_frequency_calculator(
    numerical_simulation.environment.ConstantTransmittingFrequencyCalculator(8e9) )
# set turnaround spacecraft
bodies.get_body("Delfi-C3").system_models.set_transponder_turnaround_ratio( 
    { (observation.FrequencyBands.ku_band, observation.FrequencyBands.ku_band): 1/1 } )

# Define the link ends for two-way observable
link_ends = dict()
link_ends[observation.transmitter] = observation.body_reference_point_link_end_id("Earth", "TrackingStation")
link_ends[observation.retransmitter] = observation.body_origin_link_end_id("Delfi-C3")
link_ends[observation.receiver] = observation.body_reference_point_link_end_id("Earth", "TrackingStation")
link_definition = observation.LinkDefinition(link_ends)

# Set the observation settings
observation_settings_list = [observation.n_way_range(link_definition,
                             light_time_correction_settings = [observation.jakowski_ionospheric_light_time_correction()] )]
observation_simulation_settings = observation.tabulated_simulation_settings(
    observation.n_way_range_type, link_definition, np.arange(observation_start_epoch, observation_end_epoch, 60.0) )

# Specify the two-way frequencies
n_way_range_ancillary_settings = observation.n_way_range_ancilliary_settings( frequency_bands = [observation.FrequencyBands.ku_band, observation.FrequencyBands.ku_band ] )
observation.add_ancilliary_settings_to_observable([observation_simulation_settings], n_way_range_ancillary_settings, observable_type = observation.n_way_range_type)

# Create the estimator
estimator = numerical_simulation.Estimator(
    bodies,
    parameters_to_estimate,
    observation_settings_list,
    propagator_settings)

# Simulate required observations
simulated_observations = estimation.simulate_observations(
    [observation_simulation_settings],
    estimator.observation_simulators,
    bodies)
```
The final call gives the error:
`RuntimeError: Error when getting double intermediate observation data; could not retrieve type 0`
[...]

PR #404, opened 2025-08-22
Frequency-dependent observation corrections for all observation models

DominicDirkx:
Addressed #396


DominicDirkx:
Unit test of new functionality to be added

PR #423, opened 2025-09-11
WIP: Custom observation viability

larshinueber:
Draft PR to check cross-platform builds and tests

TODO:
- [ ] Add more extensive unit test
- [ ] Add C++ docstrings
- [ ] Add API docstrings, including code example
- [ ] Check implementation of casting to derived class (see comment below)

PR #419, opened 2025-09-03
Switch `-isystem` with `-I` in build flags

suvayu:
## Changes

- `-isystem` marks an include directory as a system include, which breaks conda-free builds, because it breaks the order of header file discovery for dependencies like Eigen.
- Update conda environment to include compilers, and use that for building so that conda builds do not pick system includes.

## Testing

In a fresh conda environment created with the updated environment file, I did two separate builds, one with `build.py` script, the other with the cmake flags from CI.

### Build with `build.py`

```shell
python ./build.py
python ./install.py -e
pytest
```

- The build has errors as shown below, but AFAIU, that is acceptable
  ```
  pybind11_stubgen - [  ERROR] In tudatpy.kernel.trajectory_design.transfer_trajectory.create_transfer_trajectory : Invalid expression 'tudat::simulation_setup::SystemOfBodies'
  pybind11_stubgen - [  ERROR] In tudatpy.kernel.trajectory_design.transfer_trajectory.set_low_thrust_acceleration : Invalid expression 'tudat::simulation_setup::SystemOfBodies'
  pybind11_stubgen - [WARNING] Enum-like str representations were found with no matching mapping to the enum class location.
  ```
- Running `pytest` gives me this failure, but it seems like an unrelated failure.
  ```python
  __________________________________________________ test_compare_mpc_horizons_eph ____________________________________

      def test_compare_mpc_horizons_eph():
          """Compares true observations from BatchMPC to interpolated simulated RA/DEC from JPL Horizons"""
          batch = BatchMPC()
  >       batch.get_observations([433])

  tests/test_tudatpy/test_data_mpc.py:184:
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
  /opt/miniforge3/envs/tudatpy-dev/lib/python3.11/site-packages/tudatpy/data/mpc/mpc.py:874: in get_observations
      self._refresh_metadata()
  /opt/miniforge3/envs/tudatpy-dev/lib/python3.11/site-packages/tudatpy/data/mpc/mpc.py:730: in _refresh_metadata
      self._observatories = list(self._table.observatory.unique())
                                 ^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #432, opened 2025-09-20
Refactor stub generation for python scripts

alfonsoSR:
- Replace `mypy.stubgen` with a custom stub generator: makes stub generation much faster, the implementation is simpler, and it gives us more control over how the stubs are produced.
- Replace the `print` statements in `build.py` with a logger: makes it possible to hide all the `pybind11-stubgen` "errors" from users, while keeping them accessible for us (with the `--debug` flag of `build.py`)
- Expose and document the constructor of `TranslationalStatePropagatorSettings`
- Expose the default constructor of `BodySettings` to be able to create empty settings without adding them to a `BodyListSettings` object
- Fix incorrect indentation in docstrings

All the code is backwards compatible and it does not introduce new dependencies

### DIFF
--- a/build.py
+++ b/build.py
@@ -8,6 +8,40 @@ import ast
 import tempfile
 from dataclasses import dataclass
 from typing import Generator
+import logging
+
+
+class CustomFormatter(logging.Formatter):
+
+    grey = "\x1b[38;20m"
+    blue = "\x1b[34;20m"
+    green = "\x1b[32;20m"
+    yellow = "\x1b[33;20m"
+    red = "\x1b[31;20m"
+    bold_red = "\x1b[31;1m"
+    reset = "\x1b[0m"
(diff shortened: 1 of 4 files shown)


alfonsoSR:
And no one else has tested the changes in `build.py` yet, so let's do some local tests in other computers before merging :)

DominicDirkx:
What about these two: https://github.com/tudat-team/tudatpy/pull/419, https://github.com/tudat-team/tudatpy/pull/423?

alfonsoSR:
Oh, that was by mistake 😅 I deleted the branches that I had created and were already merged into `develop`, and `develop_10` was one of them. I didn't realize that any PR asking to be merged into `develop_10` would be closed. It should be possible to

- Change the target branch of these PRs to `develop`
- Recover `develop_10` by pushing it again or from the settings section of the repo

Sorry for that :(

PR #439, opened 2025-09-26
Fix/kernel imports

DominicDirkx:


DominicDirkx:
This is now building successfully on Azure, thanks for the quick fix!!

PR #459, opened 2025-10-03
Expose constructors of BodySettings and TranslationalStatePropagatorSettings

alfonsoSR:
- Exposed default constructor of BodySettings to allow creating empty body settings without need for a BodyListSettings object
- Exposed constructor of TranslationalStatePropagatorSettings to allow bypassing factory function

Don't trust the commit history, I am reorganizing code that was spread over multiple branches. The only changes are the ones I list above.

### DIFF
--- a/src/tudatpy/dynamics/environment_setup/expose_environment_setup.cpp
+++ b/src/tudatpy/dynamics/environment_setup/expose_environment_setup.cpp
@@ -101,6 +101,7 @@ void expose_environment_setup( py::module& m )
          :class:`BodyListSettings` object.
 
       )doc" )
+            .def( py::init<>( ) )
             .def_readwrite( "constant_mass", &tss::BodySettings::constantMass, R"doc(
 
          Mass that gets assigned to the vehicle. This mass does *not* automatically define a gravity field
@@ -232,8 +233,7 @@ void expose_environment_setup( py::module& m )
 
          :type: RadiationSourceModelSettings
       )doc" )
-            .def_readwrite(
-                    "vehicle_shape_settings", &tss::BodySettings::bodyExteriorPanelSettings_, R"doc(
+            .def_readwrite( "vehicle_shape_settings", &tss::BodySettings::bodyExteriorPanelSettings_, R"doc(
 
          Object that defines the settings of an exterior panelled vehicle shape that is to be created. A variable of this type is typically
          assigned by using a function from the :ref:`vehicle_systems` module.
@@ -252,8 +252,7 @@ void expose_environment_setup( py::module& m )
 
       )doc" );
 
-    py::class_< tss::BodyListSettings, std::shared_ptr< tss::BodyListSettings > >(
-            m, "BodyListSettings", R"doc(
+    py::class_< tss::BodyListSettings, std::shared_ptr< tss::BodyListSettings > >( m, "BodyListSettings", R"doc(
 
          Class for defining settings for the creation of a system of bodies.
 
@@ -308,8 +307,7 @@ void expose_environment_setup( py::module& m )
 
      )doc" )
             .def( "add_settings",
-                  py::overload_cast< std::shared_ptr< tss::BodySettings >, const std::string >(
-                          &tss::BodyListSettings::addSettings ),
+                  py::overload_cast< std::shared_ptr< tss::BodySettings >, const std::string >( &tss::BodyListSettings::addSettings ),
                   py::arg( "settings_to_add" ),
                   py::arg( "body_name" ),
                   R"doc(
@@ -363,8 +361,7 @@ void expose_environment_setup( py::module& m )
 
          :type: str
       )doc" )
-            .def_property_readonly(
-                    "frame_orientation", &tss::BodyListSettings::getFrameOrientation, R"doc(
+            .def_property_readonly( "frame_orientation", &tss::BodyListSettings::getFrameOrientation, R"doc(
 
          **read-only**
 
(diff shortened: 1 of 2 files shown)


DominicDirkx:
See comment in #460 :)

DominicDirkx:
There also seems to be a lot of code reformatting, which makes it difficult to review the PR since it's not clear which changes are intentional and which are from the auto-formatting (?)

DominicDirkx:
What is the reason to add code to bypass the factory function? My worry is that exposing this will give the user two ways to do the same things, and having both in the API will lead to confusion on what the recommended/correct workflow is.

alfonsoSR:
From my point of view, factory functions (unless they are static methods of the type they create) make code more difficult to understand, because they hide the type of the object they are creating. I see the point of using them when they add extra value with respect to the constructor (e.g. by processing or checking the inputs), but otherwise they just hide information from the user.

Also, although it is not the case for `propagator.translational`, our factory functions tend to have a base class as return type, instead of the derived class they are actually returning, so they tend to mess up syntax highlighting and force you to write unnecessary assertions (equivalent to `dynamic_cast`) just to avoid syntax warnings.

<details><summary>Example: Negative effect on syntax highlighting</summary>
<p>

A good example is `create_dynamics_simulator`, which just calls the constructor of the right type of propagator based on the type of your settings, so it does not add extra value in terms of processing the inputs you give it before using them.

```C++
template< typename StateScalarType = double, typename TimeType = double >
std::shared_ptr< propagators::DynamicsSimulator< StateScalarType, TimeType > > createDynamicsSimulator(
        const simulation_setup::SystemOfBodies& bodies,
        const std::shared_ptr< propagators::PropagatorSettings< StateScalarType > > propagatorSettings,
        const bool areEquationsOfMotionToBeIntegrated = true )
{
    if( std::dynamic_pointer_cast< propagators::SingleArcPropagatorSettings< StateScalarType, TimeType > >( propagatorSettings ) !=
        nullptr )
    {
        return std::make_shared< propagators::SingleArcDynamicsSimulator< StateScalarType, TimeType > >(
                bodies,
                std::dynamic_pointer_cast< propagators::SingleArcPropagatorSettings< StateScalarType, TimeType > >( propagatorSettings ),
                areEquationsOfMotionToBeIntegrated,
                propagators::PredefinedSingleArcStateDerivativeModels< StateScalarType, TimeType >( ) );
    }
    else if( std::dynamic_pointer_cast< propagators::MultiArcPropagatorSettings< StateScalarType, TimeType > >( propagatorSettings ) !=
             nullptr )
    {
        return std::make_shared< propagators::MultiArcDynamicsSimulator< StateScalarType, TimeType > >(
                bodies,
                std::dynamic_pointer_cast< propagators::MultiArcPropagatorSettings< StateScalarType, TimeType > >( propagatorSettings ),
                areEquationsOfMotionToBeIntegrated );
    }
    else if( std::dynamic_pointer_cast< propagators::HybridArcPropagatorSettings< StateScalarType > >( propagatorSettings ) != nullptr )
    {
        return std::make_shared< propagators::HybridArcDynamicsSimulator< StateScalarType, TimeType > >(
                bodies,
                std::dynamic_pointer_cast< propagators::HybridArcPropagatorSettings< StateScalarType, TimeType > >( propagatorSettings ),
                areEquationsOfMotionToBeIntegrated,
                false );
    }
    else
    {
        throw std::runtime_error( "Error when creating variational equations solver, dynamics type not recognized" );
    }
}
```
In order to get the right syntax highlighting for the return type of this function, I need to add unnecessary assertions to the code to do the equivalent of a `dynamic_cast`. 

```python
        simulator = tsim.create_dynamics_simulator(
            bodies=bodies,
            propagator_settings=propagator,
            simulate_dynamics_on_creation=True,
        )
        assert isinstance(simulator, tsim.SingleArcSimulator)
        simulation = simulator.propagation_results
        assert isinstance(simulation, tprop.SingleArcSimulationResults)
```

</p>
</details> 

And another reason is that I struggle to find factory functions (unless they are static methods), because I don't have any natural route that brings me to them.  I give an example of what I mean below to avoid cluttering this

<details><summary>Details: Factory functions are difficult to find</summary>
<p>

Say that I want to call `create_dynamics_simulator`. I check the docstring or the API and I see that the first argument is of type `SystemOfBodies`, so I go to the API, I search for `SystemOfBodies`, and I see that it is in `dynamics.environment`. I import the class and check the arguments of its constructor to see what I need to be able to initialize it, but it doesn't have a constructor. I check the docstring of the class, but it does not mention any factory functions, so I am stuck.

For this particular case, the factory function is mentioned in the docstring of `create_dynamics_simulator`, but what if I want to create a system of bodies to pass it to an observation model? We could try to take care of mentioning the factory functions in the docstring of the classes, but why rely on that if you can call the constructor of the class directly? 

The reason why I say that static methods are fine is that they are always with the type itself, so there is a natural way to find them without having to rely on docstrings.

</p>
</details> 

I remember @valeriof7 having a similar point of view from the presentation he made some months ago, and I've also discussed this with @larshinueber  not too long ago :)

I see your point about consistency, though. Maybe we could ask other people for input on slack, and if there is interest we could discuss this during a developer meeting.

DominicDirkx:
I think this is a much longer discussion than one to have in this thread, and if we make any modifications here I think it's good to do it in a library-wide manner, rather than a couple of (arbitrary?) classes/functions. A few thoughts:

* I think there's an important distinction between things like the BodySettings and BodyListSettings that don't have any derived classes/different types, and propagator/acceleration/... settings. For the former case, I see few (dis)advantages either way, and it largely ends up being a matter of aesthetics/habits. 
* For things with derived classes, it's a very different situation. Pre-Python we used the classes, and it was very difficult forf people to find how to create a settings object. For instance for accelerations: you want a point-mass gravity or aerodynamic? You create a base class pointer with the right enum as input. You want a spherical harmonics acceleration? You create the specific derived class pointer with max. degree/order as input. In the current setup, all such types are handled equally. Having them as factory functions may not be what people are used to, but with proper documentation (and some time getting used to the library) I don't see it as a big problem
* Having the factory functions return the derived classes when relevant is fine, I did not realize the issue with autocompletion when using the base class. I think doing this is also a good exercise to ensure that all our derived classes are actually exposed
* @alfonsoSR The issue you mention in navigating the documentation is indeed an important one. There's an open issue to start each submodule page of the environment setup with a block like this (https://py.api.tudat.space/en/latest/dynamics/environment_setup/atmosphere.html) relating what is here to the rest of the module. Adding additional links in the documentation (for instance for the SystemOfBodies class a link to the user guide where its use and creation is explained)

Since this needs a broader discussion and/or broader development, I'm closing this PR now, with points to be discussed at the next meeting

Issue #202, opened 2024-12-13
Error rendering docstrings


yiquintero:
## Description

The docstrings for the functions listed below are incorrectly rendered by Sphinx. The issue occurs when building the documentation locally using tudat-multidoc or the new method introduced as part of the [2024 updates to the tudatpy API documentation workflow](https://github.com/tudat-team/tudatpy/wiki/2024-Updates-to-tudatpy-API-Documentation-Workflow). The docstrings are also rendered incorrectly on the [API Reference website](https://py.api.tudat.space/en/latest/).

### Ephemeris
1. interpolated_spice
2. tabulated_from_existing

### Estimation
<details>
<summary> List of Functions </summary>

1. add_dependent_variable
2. compatible_dependent_variable_settings
3. compatible_dependent_variables_list
4. concatenated_dependent_variable
5. dependent_variable
6. dependent_variable_history
7. dependent_variable_history_per_set
8. filter_observations
9. get_bodies_in_link_ends
10. get_computed_observations
11. get_concatenated_computed_observations
12. get_concatenated_float_observation_times
13. get_concatenated_observation_times
14. get_concatenated_observations
15. get_concatenated_observations_and_times
16. get_concatenated_residuals
17. get_concatenated_weights
18. get_mean_residuals
19. get_observable_types
20. get_observation_times
21. get_observations
22. get_observations_and_times
23. get_reference_points_in_link_ends
24. get_residuals
25. get_rms_residuals
26. get_single_observation_sets
27. get_time_bounds_list
28. get_time_bounds_per_set
29. get_weights
30. set_constant_weight (x2)
31. set_reference_point (x2)
32. set_reference_points
33. set_tabulated_weights
34. split_observation_sets
35. EstimationInput constructor

</details>

### Observation
<details>
<summary> List of Functions </summary>

1. two_way_doppler_averaged
2. one_way_range
3. n_way_range
4. two_way_range
5. angular position
6. relative_angular_position
7. one_way_doppler_instantaneous
8. one way doppler averaged
9. n_way_doppler_averaged
10. n_way_doppler_averaged_from_one_way_links

</details>

### Propagator
<details>
<summary> List of Functions </summary>

1. hybrid_arc
2. translational
3. rotational
4. mass
5. multitype
6. multi_arc

</details>

larshinueber:
Hi all,

[...]

PR #461, opened 2025-10-03
Refactor stub generator

### LINKED ISSUE #277: [BUG]: Exposing functions with `nullptr` as default argument
### Describe the issue:

Autogenerated stubs reveal that the way in which we expose functions with `nullptr` as default argument results into incorrect type annotations.

If the argument `x: SomeType` of a function is `None` by default, the correct annotation for the function is `x: SomeType | None = None`. Our current approach to exposing these types of functions produces `x: SomeType = None`, which results into a type-hinting error from mypy.

- Checkout the `feature/stubs` branch of `tudat-bundle`
- Checkout the `feature/stubs` branch of `tudatpy`, and the `develop` branch of `tudat`
- Run `python3 build.py` to build the library with stubs.
- Go to `tudat-bundle/build/tudatpy-stubs/astro/element_conversion/extension.pyi` and look for the `mean_to_eccentric_anomaly` function. You will see that the last argument is annotated as `root_finder: root_finders.RootFinderCore=None`

### OS and Python version:

OS: macOS Intel
Python version: 3.11

### Expected behavior:

The `root_finder` parameter of the `mean_to_eccentric_anomaly` function should be annotated as `root_finder: root_finders.RootFinderCore | None = None` or equivalent.

### Current behavior:

This is the current annotation in my code base

```python
def mean_to_eccentric_anomaly(eccentricity: float, mean_anomaly: float, use_default_initial_guess: bool=True, non_default_initial_guess: float=..., root_finder: root_finders.RootFinderCore=None) -> float:
    """Convert mean to eccentric anomaly.
    
    
    Parameters
    ----------
    eccentricity : float
        Value of the orbital eccentricity
    mean_anomaly : float
        Value of the mean anomaly
    use_default_initial_guess : bool, default = True
        Boolean to determine whether the user-defined initial guess is used for conversion, or an automatically generated one.
    non_default_initial_guess : float, default = NaN
        User-defined initial guess for conversion, to be used only if ``use_default_initial_guess`` is set to ``False``.
    root_finder : RootFinder, default = None
        User-defined root finder, overriding default root-finding algorithm for conversion (default is used if this input is left empty)
    Returns
    -------
    float
        Value of the eccentric anomaly"""
```

### Additional context:

From [pybind11 documentation](https://pybind11.readthedocs.io/en/stable/advanced/functions.html#default-arguments-revisited:~:text=args%20and%20**kwargs-,Default%20arguments%20revisited,-Keyword%2Donly%20arguments):

> Sometimes it may be necessary to pass a null pointer value as a default argument. In this case, remember to cast it to the underlying type in question, like so:
> 
> ```python
>  py::class_<MyClass>("MyClass")
>     .def("myFunction", py::arg("arg") = static_cast<SomeType *>(nullptr));
> ```

This might be of help in fixing the issue

### LINKED ISSUE #288: [BUG]: Missing types during stub generation
### Describe the issue:

The current version of the stubs includes some functions with missing type information. The most common reason for this is a function having an argument of type `A` before the exposure for class `A` is declared.

On the `main` branch of `tudat-bundle`, run `python build.py`. Multiple warnings such as the one shown here appear:

```log
pybind11_stubgen - [  ERROR] In tudatpy.kernel.numerical_simulation.propagation_setup.propagator.custom_state : Invalid expression 'tudat::propagators::SingleDependentVariableSaveSettings'
```

### OS and Python version:

OS: macOS Intel

### Expected behavior:

Expected function signature
```python
    def add_single_control_surface(
        self,
        control_surface_settings: ControlSurfaceIncrementAerodynamicCoefficientSettings, 
        control_surface_name: str
    ) -> None:
```

### Current behavior:

Current function signature
```python
    def add_single_control_surface(
        self, control_surface_settings: ..., control_surface_name: str
    ) -> None:
```

### Additional context:

Some of these warnings can be fixed by reordering the exposures, but one ends up running into situations where compiling `A` before `B` gives some warnings, and compiling `B` before `A` gives others. A robust solution for this issue would be to modify the exposure files to declare classes and enums before they are used for type annotation, as described in https://github.com/isl-org/Open3D/pull/6868.

### LINKED ISSUE #368: [F&E]: Module level docstrings through stubs
### Description:

As discussed in #342 (quoting @larshinueber):

> I dont know if it's possible to make the module level docstrings also available through the stubs (maybe @alfonsoSR knows?), otherwise I think some duplication would be worth it (at least in the basic functions that a new user of the module would probably look at, e.g. for the biases: absolute_bias, relative_bias, combined_bias).

Answer of @alfonsoSR 

> I missed this, sorry! It is possible to add module level docstrings to the stubs, but I missed that when implementing the stub generator. If we open an issue I can try to make a bit of time to fix it :)

Issue hereby opened :)

_No response_

### Additional context:

_No response_

alfonsoSR:
# Rationale

The goal of this PR is to replace `mypy.stubgen` and `pybind11-stubgen` with custom stub generators for Python and C++ extensions.

## Bad features of current stub generator

- `pybind11-stubgen` is not particularly well maintained, it does not have documentation, it is not good at processing enumerations, and it produces rough stubs that require significant post-processing.
- `mypy.stubgen` is very slow, and the stubs it produces also require a lot of post-processing

## Nice features of the new stub generators

- We can discourage users from using deprecated functions/classes/modules by not generating stubs for them. This means that deprecated code still works, but it no longer gets syntax highlighting, so new people will not find it.
- We can raise custom warnings to help us find code that does not follow some rule. For example, we can raise a warning when a function does not have documentation, or when it has C++ types in its signature.
- We can use a logger to control what gets printed to the terminal while generating the stubs. For example, we can use `log.warning` for warnings that matter to users, and `log.debug` for information that is only relevant to developers, and then decide to only print developer information if we call the `build.py` script with the `--debug` flag.
- Instead of having to define post-processing functions to fix the stubs produced by `pybind11-stubgen`, we can adapt the stub generators to our needs and simplify the implementation.

# TODO

- [x] Fix exposure of enumerations
- [ ] Try to fix representation of default arguments: #202 
- [ ] Incorporate forgotten fix from tudat-bundle: https://github.com/tudat-team/tudat-bundle/issues/40


DominicDirkx:
@alfonsoSR Could you close this PR and replace it with an issue (linking to the branch you had set up)?

PR #489, opened 2025-10-09
Feature/rtg acceleration parameter across multi arcs

jo11he:
This is my implementation for allowing the creation of individual acceleration models per arc on multi-arc setups, while all of the rtg acceleration models reference the same set of rtg model parameters


src/tudatpy/dynamics/propagation_setup/propagator/expose_propagator.cpp
```diff
@@ -832,6 +844,9 @@ Enumeration of available integrated state types.
                   &tp::TranslationalStatePropagatorSettings< STATE_SCALAR_TYPE, TIME_TYPE >::resetAccelerationModelsMap,
                   py::arg( "new_acceleration_settings" ),
                   py::arg( "bodies" ) );
+            //.def( "reset_initial_states",
```
DominicDirkx: Was this removed in relation to https://github.com/tudat-team/tudatpy/issues/493?

Issue #504, opened 2025-10-14
[F&E]: 80cols parser (addition to mpc.py)


luigigisolfi:
### Description:

Title is descriptive. This is a sub-issue of tudat-team/tudatpy-examples#125

Issue #503, opened 2025-10-14
[F&E]: Made Horizons Wrapper more robust


luigigisolfi:
### Description:

The handling of time scales within the `ephemerides()` function of the Horizons wrapper seems very delicate and intricate, both because it was using non-tudat time conversions, and because it was going back and forth between tdb and utc, leading to confusion. 

### Suggestions for implementation:

Use of `time_representation` within the `ephemerides()` function will allow for more robust and clearer data processing. Addition of a new column to the output table  - corresponding to utc times - is recommended and will be added.

luigigisolfi:
Yes, can be closed

PR #505, opened 2025-10-14
Feature/80col parser (closes sub-issues #504 & #505)

luigigisolfi:
This PR solves (sub)issue #504 by adding an 80 columns format parser, based on the `object_type` (can be `comet`, `minor_planet`, `natural_satellite`). 
It returns an `astropy` table ready to be validated through the `from_astropy()` method, and ready to be converted into an observation collection via the `to_tudat()` method.

It also solves (sub)issue #503 by making the `ephemeris()` function of the Horizons Wrapper more robust.

### DIFF
--- a/tests/test_tudatpy/test_data_mpc.py
+++ b/tests/test_tudatpy/test_data_mpc.py
@@ -305,5 +305,3 @@ def test_parse_80cols_file():
 
     tol = 5e-5 # not completely sure why some are zero and some are not.
     assert not (diff_seconds > tol).any()
-
-test_80cols_line_parser()


DominicDirkx:
Thanks for the thorough update of the mpc code and time conversions! This resolves a lot of small ongoing issues :)

I won't be able to look at this before going on holiday, but I recal @larshinueber also working on the time conversions recently. And, since I'm no Python programmer, could one of you (@valeriof7, @larshinueber, @alfonsoSR) do a check of the Python code?

I see one time conversion unit test is failing:

2025-10-16T04:35:18.2512140Z unknown location(0): [4;31;49mfatal error: in "test_date_time/testTimePointConversions": std::runtime_error: Failed to convert date to time_t using mktime for date: 1969-12-31 23:59:59.000000000000000[0;39;49m
2025-10-16T04:35:18.2513068Z /home/runner/work/tudatpy/tudatpy/tests/test_tudat/src/astro/basic_astro/unitTestDateTime.cpp(262): [1;36;49mlast checkpoint[0;39;49m

 which is an edge case of a time comversion very close to 1-1-1970

And, now that the C++ code is made more robust for time scale conversions in teh more distant past, also important to extend the unit test for this

luigigisolfi:
@DominicDirkx question: was this running on linux?

DominicDirkx:
@luigigisolfi Looks like the tests pass now! (the failing Windows test is not related to this PR)

@alfonsoSR If you could have a look over the Python code when you have the time (luigi pushed the modifications he mentioned) and make any comments you think are critical, we can get this merged (possibly with an iteration). If you see points that you think are good to address but are not directly relevant to this PR, I'd recommend we spin those off into a separate issue/PR

src/tudatpy/data/mpc/mpc.py
```diff
@@ -218,6 +217,8 @@ def get_biases_EFCC18(
     # ideally nside should be retrieved from the load_bias_file function
     hp_obj = HEALPix(nside=nside)
 
+    print(Quantity(RA, unit=u.rad))
```
alfonsoSR: Are these print statements intentional?

src/tudatpy/data/mpc/mpc.py
```diff
@@ -805,83 +805,116 @@ def _apply_EFCC18(
...
                 obs = (
                     obs.assign(
-                        epochJ2000secondsTDB=lambda x: (
-                            Time(x.epoch, format="jd", scale="utc").tdb.value
-                            - 2451545.0
-                        )
-                        * 86400
+                        RA=lambda x: (np.radians(x.RA)  + np.pi ) % (2 * np.pi) - np.pi,
+                        DEC=lambda x: np.radians(x.DEC),
+                        epochUTC=lambda x: Time(x.epoch, format="jd").to_datetime()
```
alfonsoSR: Why not use our own function to get the datetime object?

src/tudatpy/data/mpc/mpc.py
```diff
@@ -805,83 +805,116 @@ def _apply_EFCC18(
...
-                else:
-                    obs.loc[:, "number"] = obs.number.astype(str)
+                    observation_types_to_drop = ["x", "X", "V", "v", "W", "w", "R", "r", "Q", "q", "O"]
+                    obs = obs.query("note2 not in @observation_types_to_drop")
 
+                MPC_parser = MPC80ColsParser()
+                try:
+                    parsed_value = MPC_parser.parse_packed_permanent_designation(str(obs.number[0]))['number']
+                    obs.loc[:, "number"] = str(int(parsed_value))
+                except:
```
alfonsoSR: Catch a specific exception. If you use a generic `except` there is a chance that the `try` part will fail for some other reason that will go undetected.

This used to be an `if...else` block. Is there a reason to prefer a `try...except` now?

src/tudatpy/data/mpc/mpc.py
```diff
@@ -892,81 +925,90 @@ def _add_table(self, table: pd.DataFrame, in_degrees: bool = True):
...
-        # check if all mandatory names are present
-        if not set(self._req_cols).issubset(set(table.colnames)):
-            txt = f"Table must include a set of mandatory columns: {self._req_cols}"
-            raise ValueError(txt)
+        # Get column names depending on table type
+        if isinstance(table, (astropy.table.QTable, astropy.table.Table)):
+            colnames = table.colnames
+        elif isinstance(table, pd.DataFrame):
+            colnames = table.columns
+        else:
```
alfonsoSR: Shouldn't this fail based on the type annotation in the docstring?

src/tudatpy/data/mpc/mpc.py
```diff
@@ -1623,3 +1666,301 @@ def observatories_table(
...
+            suffix_val = int(packed[1:])
+            return str(prefix_val * 10000 + suffix_val)
+
+        # Case 3: Numbers < 100,000 (starts with a digit)
+        return str(int(packed))
+
+    def _unpack_provisional_minor_planet(self, packed: str) -> str:
+        """Unpacks a 7-char provisional minor planet designation."""
+        # Case 1: Survey designations (e.g., PLS2040 for 2040 P-L)
+        survey_map = {'PLS': 'P-L', 'T1S': 'T-1', 'T2S': 'T-2', 'T3S': 'T-3'}
```
alfonsoSR: Make global and document with docstring

src/tudatpy/data/mpc/mpc.py
```diff
@@ -1623,3 +1666,301 @@ def observatories_table(
...
+        if packed[0:3] in survey_map:
+            number = int(packed[3:])
+            survey_type = survey_map[packed[0:3]]
+            return f"{number} {survey_type}"
+
+        # Case 2: Standard provisional designations (e.g., K07Tf8A for 2007 TA418)
+        century_map = {'I': '18', 'J': '19', 'K': '20'}
```
alfonsoSR: Make global and document with docstring

src/tudatpy/data/mpc/mpc.py
```diff
@@ -1623,3 +1666,301 @@ def observatories_table(
         if not include_positions:
             temp = temp.loc[:, ["Code", "Name", "count"]]
         return temp
+
+class MPC80ColsParser:
+    """
+    A parser to unpack provisional and permanent designations from the MPC format.
```
alfonsoSR: I would add a reference to the document/website where they explain the format of these files. Otherwise, although the code is well documented, it can be a bit difficult to interpret.

src/tudatpy/data/mpc/mpc.py
```diff
@@ -1623,3 +1666,301 @@ def observatories_table(
...
+            cycle = val1 * 10 + val2
+
+        if cycle == 0:
+            return f"{year} {half_month_1}{half_month_2}"
+        else:
+            return f"{year} {half_month_1}{half_month_2}{cycle}"
+
+    def _unpack_provisional_comet_or_satellite(self, packed: str) -> str:
+        """Unpacks a 7-char provisional comet/satellite designation."""
+        century_map = {'I': '18', 'J': '19', 'K': '20'}
```
alfonsoSR: Duplicated. Make global and add docstring

src/tudatpy/data/mpc/mpc.py
```diff
@@ -1623,3 +1666,301 @@ def observatories_table(
...
+        # Rule for Interstellar Objects (e.g., 0002I)
+        elif re.match(r"^\d{4}I$", packed_perm_num):
+            ident_data['type'] = 'Interstellar Object'
+            number = int(packed_perm_num[0:4])
+            ident_data['name'] = f"{number}I"
+            ident_data['number'] = str(number)
+            ident_data['comettype'] = 'I'
+
+        # Rule for Minor Planets (e.g., 00433, A0345, D4341, ~000z)
+        elif packed_perm_num:
```
alfonsoSR: Shouldn't this be an `else`?

src/tudatpy/data/mpc/mpc.py
```diff
@@ -805,83 +805,116 @@ def _apply_EFCC18(
...
+                # obs['epochUTC'] = [DateTime.to(dt_obj) for dt_obj in dt_objects]
+
                 obs = (
                     obs.assign(
-                        epochJ2000secondsTDB=lambda x: (
-                            Time(x.epoch, format="jd", scale="utc").tdb.value
-                            - 2451545.0
-                        )
-                        * 86400
+                        RA=lambda x: (np.radians(x.RA)  + np.pi ) % (2 * np.pi) - np.pi,
```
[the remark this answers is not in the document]
luigigisolfi: It is was not a bug, but tudat always return Ra and Dec normalized to (-pi, pi), so we should be consistent imo

valeriof7:
Sure, I don't see any problems. However, I suggest addressing all the minor comments, merging the pull request, and opening an issue about the refactoring. Up to you.

luigigisolfi:
@DominicDirkx, please refer to [this comment](https://github.com/tudat-team/tudatpy/pull/505#discussion_r2472495088)

DominicDirkx:
Thanks! I missed that. For the Pythonically retarded (like myself), why is this:

> From the way it is implemented, this should be a (hidden) submodule instead of a class. 

true?

DominicDirkx:
> If we choose to move the implementation of the big changes to another PR, I think it is important to at least do something about https://github.com/tudat-team/tudatpy/pull/505#discussion_r2472544010 and https://github.com/tudat-team/tudatpy/pull/505#discussion_r2472531592

Agreed, and both cases don't seem to be too big to tackle

alfonsoSR:
>  For the Pythonically retarded (like myself), why is this:
> 
> > From the way it is implemented, this should be a (hidden) submodule instead of a class.
> 
> true?

@DominicDirkx My main reasons are that the state of the object cannot be changed through inputs or calls to methods, and that the methods of the class do not make use of it. The class is just being used to group a set of independent functions that share a certain purpose, and that is what a module is meant to be used for :)

DominicDirkx:
Then if we do it it's good to do it now indeed :) For my understanding, how would this change the user interface?

luigigisolfi:
@alfonsoSR , @DominicDirkx 
I am working on it, cause it is annoying to have it pending :D 
I think I have a good grasp on the suggestions, so we can probably skip the meeting on Thursday an iterate on this PR instead. Stay tuned 😆

luigigisolfi:
Hi, I have applied almost all the suggestions. Here's what I **did not** do:

> As a very nice to have, but not needed, feature. I would create a dataclass to hold the contents of and MPC data file, which would replace the Table returned by the parse_80cols_file function. The function could then be a from_80cols_file static method of the dataclass, which could also be augmented with a from_ades method. Data structures tend to make the code easier to read and write compared to dictionaries.

luigigisolfi:
As I understand it: by substituting the current Table object with a dataclass, the creation and validation of such object will still be performed internally, so nothing should change from a user's perspective. @alfonsoSR is this how you meant it?

alfonsoSR:
Sorry, I kept remembering and forgetting to answer this during the whole day :)

Replacing the Table with a data structure would change the API if we expect users to interact with the 80col parser. I guess that's not the case, but since we don't have a well defined distinction between public and private in tudatpy it basically depends on what people end up doing with this function.

Anyway, I think that if we don't  wait for too long before we change to a data structure or we don't do it at all what we have right now should be good to merge.

alfonsoSR:
I didn't check the new code, btw. I can have a look tomorrow just to be sure :)

DominicDirkx:
Sounds good! @alfonsoSR If you could give the modifications made by @luigigisolfi a quick look some time in the coming days that'd be great!

src/tudatpy/util/_support.py
```diff
@@ -363,3 +363,29 @@ def vector2matrix(flat_matrix: np.ndarray):
         Rotation matrix (3x3 orthogonal matrix).
     """
     return flat_matrix.reshape(3, 3)
+
+def to_roman(num: int) -> str:
```
alfonsoSR: Can we give this a more descriptive name? It makes sense to use names as `to_x` or `from_x` in methods, because you have an object `y` of type `Y`, and you call `y.to_x()`. Aside from being more descriptive, a name like `transform_integer_to_roman_number` would be better aligned with the general verbosity level of tudat

src/tudatpy/data/mpc/parser_80col/parsers.py
```diff
@@ -0,0 +1,180 @@
...
+
+    # Rule for Minor Planets (e.g., 00433, A0345, D4341, ~000z)
+    elif packed_perm_num:
+        ident_data['type'] = 'Minor Planet'
+        ident_data['number'] = unpackers.unpack_permanent_minor_planet(packed_perm_num)
+
+    return ident_data
+
+
+def parse_80cols_identification_fields(line: str) -> dict:
```
alfonsoSR: Specify type of keys and values of the dictionary: `dict[str, str]` instead of `dict`. Otherwise, if your dictionary is `some_dict` and you write `x = some_dict["key"]`, the syntax highlighter will not be able to know the type of `x`

src/tudatpy/data/mpc/parser_80col/parsers.py
```diff
@@ -0,0 +1,180 @@
...
+    """
+    parsed_observations = []
+    with open(filename, 'r') as f:
+        for line in f:
+            if len(line) < 80:
+                continue
+
+            try:
+                ident_data = parse_80cols_identification_fields(line)
+            except (ValueError, IndexError):
```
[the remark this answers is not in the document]
DominicDirkx: That makes sense as explanation. For empty/invalid data, returning an empty data structure (or something similar) would be much preferred over an exception @luigigisolfi Could you make this modification?

src/tudatpy/data/mpc/parser_80col/parsers.py
```diff
@@ -0,0 +1,180 @@
+import re
+import datetime
+import numpy as np
+from astropy.table import Table
+from tudatpy.astro.time_representation import DateTime
+
+# Import the refactored unpacker functions and constants
+from . import unpackers
+
+def parse_packed_permanent_designation(packed_perm_num: str) -> dict:
```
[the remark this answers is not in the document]
luigigisolfi: Added

src/tudatpy/data/mpc/parser_80col/parsers.py
```diff
@@ -0,0 +1,180 @@
...
+                    "DEC": dec_rad,
+                    "observatory": line[77:80].strip(),
+                    "magnitude": float(mag_str) if mag_str else None,
+                    "band": line[70].strip() or None,
+                    "note1": line[13].strip() or None,
+                    "note2": line[14].strip() or None,
+                    "catalog": None
+                }
+                parsed_observations.append(final_data)
+            except (ValueError, IndexError):
```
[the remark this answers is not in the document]
luigigisolfi: see comment above about `pydantic`

src/tudatpy/data/mpc/mpc.py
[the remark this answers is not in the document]
luigigisolfi: **NOT IN THIS PR**

luigigisolfi:
Hi @alfonsoSR , @DominicDirkx , I added some of the requested modifications. Other might require a little bit more thinking. I again suggest to merge this PR, open an issue to address what's left, and solve that in the upcoming weeks.

DominicDirkx:
Sounds good to me, could you check if the examples PR is up to date? And, after this is merges, could you open an issue with the open points?

@alfonsoSR This is a nice example of your proposed develop workflow, not requiring interfaces to be stable at first :)

alfonsoSR:
I am sorry to be annoying, but I don't see this as a change of interface. From my point of view, it is pushing bug-prone code to be fixed in the future instead of doing it right away.

The idea of using a dataclass that automatically checks the input on construction requires proper rules to decide if a line is valid or not, which is what the code is replacing with the use of `try:...except: continue`. There is no need for data structures to make this safe to use, it is just a matter of figuring out what rules do desirable lines follow, write a couple of `if...else` conditions to check if they are respected, and skipping the line otherwise.

For example: the `parse_80cols_identification_fields` function takes a line as a string, and directly attempts `packed_perm_num = line[0:5].strip()`, which will raise an `IndexError` if the line is empty or too short. You have a check on the length of the line right before calling this function, so I am not sure why the `except IndexError` is needed, or why wouldn't we move the check on the length of the line inside the function directly, and return with `None` if the line is invalid.

In any case, there is no need to refactor the code to use dataclasses to solve this, and I keep thinking that this kind of usage of `try:...except: continue` is dangerous, and should not be replaced before the code is merged.

Issue #492, opened 2025-10-10
[F&E]: Safer Multi-Arc Ephemeris Interpolation


jo11he:
### Description:

I am setting up a multi-arc ephemeris for a body: a multi-flyby trajectory where the state history dictionary contains data for few hours per arc. There are months/years between these arcs, yet the ephemeris will happily interpolate between these pockets of validity. 

This is potentially dangerous at the least a warning should be issued when this is happening. 
Other solutions could be to throw an exception when interpolation between these arcs is requested.

The most advanced and user friendly version could include setting a "default" ephemeris in the multi-arc ephemeris creation, that the interpolator falls back to when queried between arcs (similar to the "layering" of loaded spice kernels).

DominicDirkx:
Fix merged!

PR #512, opened 2025-10-29
Added improved multi-arc support

DominicDirkx:
Addresses #492


DominicDirkx:
Point of discussion: The code modifies the multi-arc ephemeris in its behaviour when a state is requested outside of the bounds of any of the arcs (for instance when modelling flybys by JUICE or Cassini of moons, the propagated state history will consist of non-contiguous arcs). At present, the arc preceding an 'gap' in the coverage is simply extrapolated which leads to erroneous results. The code is now modified with two different options:

1. throw an error when out of bounds
2. call a default ephemeris (user-provided) that fills in the gaps

now, this seems to work fine (although a proper unit test is to be added). **My question is on how to deal with the boundaries on a tabulated ephemeris, for which there are different options.**

As you know, the tabulated ephemerides that come from a propagtion are generated using a Lagrange interpolator. This Lagrange interpolator works well, except at the boundaries of the domain. My question is: when checking if an epoch is 'in' an arc for the multi-arc ephemeris, should we consider the boundaries of this domain as valid or not? Depending on how the Lagrange interpolator is set up, it can be set to throw an error, a warning or use a cubic spline for the boundaries. It would be easiest to just not permit the boundary regions to be interpolated in the multi-arc ephemeris, but this could override the user behaviour (if they choose the cubic spline at the boundaries)

DominicDirkx:
> but this could override the user behaviour (if they choose the cubic spline at the boundaries)
> 
> couldn't we check the user settings and decide accordingly?

@valeriof7 We could, absolutely, and this would be the most consistent. But the dependence between interpolator settings and ephemeris behaviour may also be a little confusing to the user. But I think this is the best solution.

DominicDirkx:
I think I agree that this is the most suitable option. But, from a user perspective there is a confusing aspect to this. This is the input type (in C++) for the different ephemerides:

const std::map< double, std::shared_ptr< Ephemeris > >& singleArcEphemerides,

where the key is the epoch at which the arc for the associated Ephemeris starts. But now, if the 'earliest valid time' for the Ephemeris is larger than its corresponding arc start time (key), the multi-arc ephemeris will return an error in an interval _after_ the arc start time provided by the user. This is quite counter-intuitive, since you'd expect the ephemeris to be used to the epochs after the provided arc start time.

But since in most cases the multi-arc ephemeris is created automatically, I think that with good documentation this is ok.

jo11he:
I suspect this PR could have caused an error with the synchronous rotation model, which I reported here:

https://github.com/orgs/tudat-team/discussions/117

Issue #521, opened 2025-11-09
[BUG]: Misspelled attributes in tudatpy.kernel.dynamics.environment.GroundStationState


PascalChristiaanse:
### Describe the issue:

Attributes
- cartesian_positon_at_reference_epoch 
- spherical_positon_at_reference_epoch 
of class tudatpy.kernel.dynamics.environment.GroundStationState
in python exposure
have misspelled positon, should be position

<img width="1273" height="359" alt="Image" src="https://github.com/user-attachments/assets/a5705144-c728-4549-9760-d872f9cbeccf" />

### How to reproduce the issue:

any SystemOfBodies instance WITH a ground station registered
station = SystemOfBodies.bodies.get("Earth").get_ground_station(ground_station_name)
station.station_state.cartesian_positon_at_reference_epoch 
---------------------------------^^^^^----------------------

Open any debugger and you'll see the misspelled variable. It extends only to the python exposure layer, below that everything is fine

### OS and Python version:

Ubuntu 24.04 running from windows 11 Version	10.0.26200 Build 26200 using WSL2 

### Expected behavior:

cartesian_positon_at_reference_epoch and spherical_positon_at_reference_epoch  should have position spelled correctly

### Current behavior:

Position is spelled positon

### Additional context:

Reporting this to fix it myself as a first contributino

valeriof7:
Thank you for creating the issue.

Fixed in PR #522

PR #522, opened 2025-11-11
Fix "position" typo

valeriof7:
Close issue #521


DominicDirkx:
Technically not backwards compatible, but sufficiently niche to not really care :)

Issue #530, opened 2025-11-13
[F&E]: Add spacetrack Tudatpy Unit Tests


luigigisolfi:
### Description:

Title is descriptive.

Issue #529, opened 2025-11-13
[F&E]: Implementing Smarter API Calls to Space-Track.org


luigigisolfi:
### Description:

The current implementation of SpaceTrack (ST) query does not account for the [SpaceTrack API Usage guidelines ](https://www.space-track.org/documentation). This granted me a suspension of my ST account because of exceeded query limit :D 

This issue is also connected to issue ###, which is about adding tudatpy tests for `spacetrack`. `discos` will undergo the same fate, see issues ### and ###. 

### Suggestions for implementation:

In order to avoid this, a refactoring of t`udatpy/data/spacetrack/spacetrack.py` will be implemented, that uses a cached system that
1) optimizes (reduces) API calls
2) upon object query, it checks whether a TLE corresponding to that object is already present in the system cache. 

> Note:
These changes will introduce the following dependency:  [`requests_cache`](https://pypi.org/project/requests-cache/).

PR #533, opened 2025-11-13
Implementing Smarter API Calls to Space-Track.org

luigigisolfi:
This PR addresses #529 by implementing Smarter API calls to Space-Track.org. 

It also addressed #530 and implements the solution outlined in #643 for unit tests using remote connection (--remote-data flag, adding conftest.py)


luigigisolfi:
@minervino96 @valeriof7 @DominicDirkx  I think this is now in line with what we wanted. Data is saved to `get_resources_path/tle_data` and the query checks for TLE existence in that folder before hitting the SpaceTrack API. 

If you could add a disclaimer anyway (to say: hey, be careful with how many times you call the SpaceTrack API), where would you put it? I was thinking we can put it both in the Jupyter Notebook and in the API(?)

I will write unit tests in an upcoming PR

luigigisolfi:
> @luigigisolfi Thanks, Luigi! Out of curiosity, was my approach of any use for it?

@minervino96  Yes Thanks! I basically used the idea of adding date-tags to the downloaded TLEs, as you did. However, functionality needed to be adapted to the existing code, also to take care of different downloading methods. For instance we have two calls, one querying the full history of TLEs from GP, for a specific object in descending/ascending order (you can also select the number **n** of TLEs you want for a specific (list of) object(s). If **n = 1**, keeps the latest only.), the other one allows to specify a date range inputs are `norad_id`, `start_date`, `end_date`. The latter almost mirrors what you provided, but the "filtering" is done on the queried URL directly, using:

`spacetrack-url/.../.../EPOCH/<start_date>end_date/.../format/json`
instead of 
`spacetrack-url/.../.../orderby/EPOCH%20asc/format/json`

src/tudatpy/util/_support.py
```diff
@@ -2,11 +2,51 @@
 from tudatpy.math import interpolators
 # from tudatpy.dynamics.propagation_setup import propagator
 import os
-from typing import Union, TYPE_CHECKING, Callable
+from typing import Union, Callable
+
```
DominicDirkx: Do these belong here? I though these would enter the FOTOS repo?

DominicDirkx:
> Data is saved to get_resources_path/tle_data

We may need to create an empty directory in the resources, I'm not sure if this will work on all systems if the directory does not exist

environment.yaml
```diff
@@ -28,6 +28,8 @@ dependencies:
   - geopandas # for fancy ground track visualizations (kosmos example)
   - requests
   - spiceypy # for automatic data downloader and kernel converter
+  # Smart API requests
+  - requests_cache
```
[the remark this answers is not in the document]
luigigisolfi: No. Removed

src/tudatpy/data/spacetrack/spacetrack.py
```diff
@@ -1,385 +1,701 @@
...
 import getpass
 import os
+import requests
 from collections import defaultdict
 import numpy as np
 from tudatpy.dynamics import environment, environment_setup, propagation_setup
 from datetime import datetime, timedelta
-import requests
 import math
+from tudatpy.data import get_resource_path
```
luigigisolfi, line 10: Added this because we want the data to be stored in a persisent tudat-related folder. Let's discuss whether we should move it to a different folder, or call it in a different way

src/tudatpy/data/spacetrack/spacetrack.py
```diff
@@ -1,385 +1,701 @@
...
+
+        self.password: str = password
+        """ Space-Track.org username """
+
+        self.spacetrack_url: str = 'https://www.space-track.org'
+        """ Space-Track.org url """
+
+        # Setup local TLE data folder
+        self.tle_data_folder = tle_data_folder
+        if not os.path.exists(self.tle_data_folder):
```
luigigisolfi: if the `tle_data` folder does not exist, it creates it

src/tudatpy/data/spacetrack/spacetrack.py
```diff
@@ -1,385 +1,701 @@
...
+            if response.status_code == 200:
+                print("Login successful.")
+            else:
+                print(f"Login failed with status code: {response.status_code}")
+        except requests.exceptions.RequestException as e:
+            print(f"Login failed: {e}")
+            raise
+
+    def _get_json_and_save(self, url: str, json_name: str) -> dict | list | None:
+        """
```
luigigisolfi, line 127: here, the docstring is explanatory :)

src/tudatpy/data/spacetrack/spacetrack.py
```diff
@@ -1,385 +1,701 @@
...
+        end_date : str
+            The end date of the range (YYYY-MM-DD).
+
+        Returns
+        -------
+        list[dict] | None
+            A list of TLE dictionaries covering the requested range, or None if retrieval fails.
+        """
+        # 1. Smart Range Check
+        local_tles = self._retrieve_local_tle_data(norad_id, start_date, end_date)
```
luigigisolfi: attempts local tle retrieval for the specified time range

src/tudatpy/data/spacetrack/spacetrack.py
```diff
@@ -1,385 +1,701 @@
...
+            self.spacetrack_url,
+            f"basicspacedata/query/class/gp/NORAD_CAT_ID/{norad_id}/EPOCH/{start_date}--{end_date}/orderby/EPOCH asc/format/json"
+        )
+
+        filename = f"tle_{norad_id}_{start_date}_{end_date}.json"
+
+        # Use centralized saver (which handles saving to folder)
+        # Note: We pass the URL. Since we know the file didn't exist (from step 1 check),
+        # _get_json_and_save will fetch and save it.
+        return self._get_json_and_save(url, filename)
```
luigigisolfi: if no local file matching time range is found in `tle_data` folder, hit API and save the downloaded file inside `tle_data`, so that the next time we query that object for those time ranges, it will be found in local.

src/tudatpy/data/spacetrack/spacetrack.py
```diff
@@ -1,385 +1,701 @@
...
+
+    #####################################
+    # --- AVAILABLE DOWNLOAD METHODS ---#
+    #####################################
+
+    def latest_on_orbit(self) -> dict | list | None:
```
luigigisolfi: all available download methods now make use of this "smarter" retrieval: first check local, and hit API if requested json is not found.

luigigisolfi:
This overview explains how to work with the data retrieved by the SpaceTrackQuery class.

**1. Interacting with Downloaded Files**
All data retrieved from Space-Track is cached locally in the directory specified by `tle_data_folder` (which defaults to `.../tudatpy/data/tle_data`).

**File Structure**: Cached JSON files use a specific internal structure containing both metadata and the orbital records:
{"last_api_hit": "<ISO datetime>", "data": [...]}.

**Loading Data**: Users can interact with these files using standard Python json tools or through the `OMMUtils` utility class:

**Extraction**: `OMMUtils.get_tles()` can extract TLE line pairs directly from a loaded OMM dictionary or list.

**Conversion**: `OMMUtils.get_tudat_keplerian_element_set() `converts a single record from the data list into SI units (meters and radians) formatted for Tudat.

**Maintenance**: The `clean_tle_file() `method can be used to manually remove duplicate entries from a specific cache file in-place, keeping only the record with the latest `CREATION_DATE` for any given epoch.

**2. Merging Data with update_existing**
Most download methods (such as `latest_on_orbit`, `get_tles_by_norad_ids`, and `filtered_by_oe_dict`) include an update_existing parameter.

- When `update_existing=False` (Default): The class operates in "overwrite mode." If a cache file already exists for that query, it is deleted and replaced entirely by the new API results.

- When `update_existing=True`: The class merges new API results into the existing local file.

- Deduplication Logic: During a merge, the system identifies unique records using a combined key of `NORAD_CAT_ID` and `EPOCH`. If the same record exists in both the local cache and the new API data, the record with the later `CREATION_DATE` is kept. The final merged list is then sorted by `EPOCH` and `NORAD_CAT_ID`.

**3. Managing API Hits and Cooldowns**
To prevent redundant server requests and hammering the API, the `get_tles_for_date_range` method implements a smart fetch logic.

- 1.5-Hour Cooldown: The class enforces a 1.5-hour cooldown period between API hits for a specific satellite.
    Cooldown Logic: 
- 1.  The system first checks if the local cache already fully covers the requested start and end dates.
- 2.  If data is missing but the last_api_hit timestamp in the file indicates a fetch occurred less than 1.5 hours ago, the API call is skipped even if gaps remain in the data.

- override_last_api_hit: Setting this parameter to True bypasses the 1.5-hour timer entirely, forcing the system to query the API regardless of how recently it was last accessed. This is useful if you know new data has just been released or if you need to fill a specific gap immediately.


A little snippet that creates a user-custom catalogue of the first 5 and 10 polar and equatorial satellites (total of 15 satellites, in descending epoch order). This code creates a json file containing a list of 15 OMMs and you can set TLE Ephemeris to retrieve the following ground plot (plots not validated yet)


```
# 1. Initialize SpaceTrack
spacetrack_query = SpaceTrackQuery(username='username',
                                   password='password')

# custom catalog name
custom_catalog_file = "my_polar_and_equatorial_satellites.json"

# --- STEP 1 & 2: Fetch Data for Polar Satellites ---
spacetrack_query.filtered_by_oe_dict(
    filter_oe_dict={'INCLINATION': (97.0, 99.0)},
    limit=5,
    output_file=custom_catalog_file,
    update_existing=False
)

# --- STEP 1 & 2: Fetch Data for Equatorial Satellites, Use update_existing = True ---
spacetrack_query.filtered_by_oe_dict(
    filter_oe_dict={'INCLINATION': (-10,10)},
    limit=10,
    output_file=custom_catalog_file,
    update_existing=True
)
```
Screenshot 2026-02-18 at 13 36 34


@minervino96, @DominicDirkx

DominicDirkx:
@luigigisolfi This is now ready for review right?

luigigisolfi:
> @luigigisolfi This is now ready for review right?

Yes

src/tudatpy/data/spacetrack/spacetrack.py
```diff
@@ -1,398 +1,700 @@
...
+        parts = ["basicspacedata/query", f"class/{tle_class}", f"NORAD_CAT_ID/{id_string}"]
+        if use_orderby:
+            parts.append(f"orderby/{orderby}")
+        parts.extend([f"limit/{limit_per_object}", "format/json"])
+
+        json_data = self._get_json_and_save(self._build_url("/".join(parts)), json_name, merge=update_existing)
+
+        if json_data and (history or limit_per_object > 1):
+            filepath = os.path.join(self.tle_data_folder, json_name)
+            filtered = OMMUtils.filter_tles_keep_latest_creation_from_json(filepath)
```
DominicDirkx: This function no longer exists, and the code will crash when it gets here (?) Any idea why this is not covered by the tests?

src/tudatpy/data/spacetrack/spacetrack.py
```diff
@@ -1,398 +1,700 @@
...
+            data = content["data"]
+            metadata = {k: v for k, v in content.items() if k != "data"}
+            is_dict = True
+        else:
+            print(f"Skipping {filepath}: unknown format.")
+            return
+
+        initial_count = len(data)
+        unique: dict[str, dict] = {}
+        for entry in data:
```
DominicDirkx, line 685: Not sure how this works, but won't this remove a TLE if an existing one with the same epoch but for a different object exists?

DominicDirkx:
I added some comments into the code (looks like nothing major). I haven't checked the unit tests yet, will do so later. Also check if the .rst file is still consistent with the modifications

src/tudatpy/data/spacetrack/spacetrack.py
luigigisolfi: Hello! That function has been superseded by `_save_unique_sorted`, I had forgotten to remove that bit. That part was not covered by the tests because there is no need anymore, since the logic for filtering is now different, so the if can be removed at once. I will push the changes

src/tudatpy/data/spacetrack/spacetrack.py
luigigisolfi: Thanks for spotting this bug when dealing with mixed (multi-satellite) jsons. Fixed

src/tudatpy/data/spacetrack/spacetrack.py
```diff
@@ -264,8 +266,8 @@ def get_tles_for_date_range(
         if needs_fetch:
             print(f"Fetching {norad_id} from API ({start_date} -- {end_date})...")
             url = self._build_url(
-                f"basicspacedata/query/class/gp/NORAD_CAT_ID/{norad_id}"
-                f"/EPOCH/>{start_date}<{end_date}/orderby/EPOCH asc/format/json"
+                f"basicspacedata/query/class/gp_history/NORAD_CAT_ID/{norad_id}"
+                f"/EPOCH/{start_date}--{end_date}/orderby/EPOCH asc/format/json"
```
luigigisolfi, line 270: I noticed a bug here while playing around with the code. It was originally 'gp' but it should be 'gp_history'

src/tudatpy/data/spacetrack/spacetrack.py
```diff
@@ -556,14 +556,19 @@ def get_tles(json_data: list[dict] | dict) -> dict[str, tuple[str, str]]:
...
+            pass
+
+        final_dict = defaultdict(list)
+
+        for entry in json_data:
+            final_dict[entry["NORAD_CAT_ID"]].append(
+                (entry["TLE_LINE1"], entry["TLE_LINE2"])
+            )
+
+        return final_dict
```
luigigisolfi, line 569: While playing with the code I noticed the older version was buggy when dealing with multiple satellites and multiple reference epochs. Fixed.

src/tudatpy/data/spacetrack/spacetrack.py
```diff
@@ -234,8 +234,6 @@ def get_tles_for_date_range(
         req_start = datetime.strptime(start_date, "%Y-%m-%d")
         req_end = datetime.strptime(end_date, "%Y-%m-%d") + timedelta(days=1, microseconds=-1)
 
-        print(req_start, req_end)
-
```
luigigisolfi, line 238: Removed unwanted debug print statement

luigigisolfi (commented):
@DominicDirkx  I  addressed both your comments and fixed some undesired behavior I found while playing with the code.

PR #536, opened 2025-11-17
Feature/arcwise aerodynamic scaling parameters

jo11he:
Implementation of aerodynamic scaling parameters for arc-wise estimation.

Incl. Unit tests for parameter partial evaluation (caveat!) and estimation in the loop.
Incl. refactoring of aerodynamic partials class, for consistency between constant_drag_coefficient and xxx_component_scaling parameters, plus inclusion of arc_wise_constant_drag_coefficient parameter in the unit tests mentioned above.

caveat: the numerical evaluation of the arc_wise_constant_drag_coefficient (test reference value in unit tests for parameter partial evaluation is numerically correct, but always assigns to column one of partials matrix)

### DIFF
--- a/src/tudatpy/estimation/observations_setup/observations_wrapper/expose_observations_wrapper.cpp
+++ b/src/tudatpy/estimation/observations_setup/observations_wrapper/expose_observations_wrapper.cpp
@@ -430,7 +430,12 @@ void expose_observations_wrapper( py::module& m )
         )doc" );
 
     m.def( "create_pseudo_observations_and_models",
-           &tss::simulatePseudoObservations< TIME_TYPE, STATE_SCALAR_TYPE >,
+    py::overload_cast<  const tss::SystemOfBodies &,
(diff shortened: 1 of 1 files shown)


include/tudat/simulation/estimation_setup/fitOrbitToEphemeris.h
```diff
@@ -81,6 +81,46 @@ simulatePseudoObservations( const SystemOfBodies& bodies,
     return std::make_pair( observationModelSettingsList, observationCollection );
 }
 
+
+template< typename TimeType = double, typename StateScalarType = double >
```
DominicDirkx: Not related tot this PR (?) but very nice to have :)

DominicDirkx:
The build fails because the added simulatePseudoObservations function is an overload, and the Python exposure of this function needs to be told which function to bind (using py::overload_cast). Could you modify it?

Issue #524, opened 2025-11-11
[COLLECTION]: Collect and fix missing API entries in RST files and C++ docstrings


valeriof7:
### Describe the issue:

This issue tracks an ongoing effort to audit and synchronize our Python API documentation (.rst files) with the C++ bindings.

While working with the documentation, I discovered that a significant number of functions, classes, and enumerations are properly exposed in our C++ (`expose_*.cpp`) files but were never added to the corresponding .rst files. This means they are completely invisible in the final Sphinx-generated API documentation, despite being functional in the Python module.

As part of this investigation, I am also systematically collecting a list of all missing docstrings in the C++ exposure files themselves.

This is a "work in progress" issue to track the overall effort. I will be submitting PRs to fix this with module by module commits.

So far, the following modules (and all applicable submodules) have been checked:

Checked and Fixed:
- data
- astro
- constants

Just Checked:
- dynamics.environment
- dynamics.environment_setup

I will continue this process for the remaining modules and update this issue accordingly.

# List 
Below is a list of missing entries, divided by module and category (docstring or rst):

### Missing RST Entries
* **Classes:**
    * `ConstantEphemeris`
    * `KeplerEphemeris`
    * `MultiArcEphemeris`
    * `TabulatedEphemeris`
    * `Tle`
    * `TleEphemeris`
    * `IonosphereModel`
    * `AerodynamicCoefficientGenerator36`
    * `ControlSurfaceIncrementAerodynamicInterface`
    * `CustomControlSurfaceIncrementAerodynamicInterface`
    * `LongitudeLibrationCalculator`
    * `DirectLongitudeLibrationCalculator`
    * `SynchronousRotationalEphemeris`
    * `AerodynamicAngleRotationalEphemeris`
    * `DirectionBasedRotationalEphemeris`
    * `InertialBodyFixedDirectionCalculator`
    * `CustomBodyFixedDirectionCalculator`
    * `PolyhedronGravityField`
    * `CannonballRadiationPressureTargetModel`
    * `TimingSystem`
* **Functions:**
    * `get_default_local_inclination_mach_points`
    * `get_default_local_inclination_angle_of_attack_points`
    * `get_default_local_inclination_sideslip_angle_points`
    * `get_local_inclination_total_vehicle_area`
    * `get_local_inclination_mesh`

### Missing CPP Docstrings
* **Classes (and their methods):**
    * `ConstantEphemeris`
    * `KeplerEphemeris`
[...]

PR #545, opened 2025-11-20
Doc/fix api

valeriof7:
Related to WIP Issue #524


DominicDirkx:
@valeriof7 I added one very minor clarification to one of the new pieces of text. I think these modifications are great, and I'd propose we merge them already, and add new PRs to incrementally address other points in #524

PR #550, opened 2025-11-25
set c version to C++23

PascalChristiaanse:
# Update the default C++ version from 14 to 23
Like was suggested in discussions/Will the internal C++ standard be updated, a PR to see if the CI/CD pipeline will accept this. 
Didnt test on my machine as it is very slow...

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -160,4 +160,4 @@ jobs:
       working-directory: "${{ github.workspace }}/build"
       # Execute tests defined by the CMake configuration. Note that --build-config is needed because the default Windows generator is a multi-config generator (Visual Studio generator).
       # See https://cmake.org/cmake/help/latest/manual/ctest.1.html for more detail
-      run: ctest --build-config ${{ matrix.build_type }} --output-on-failure -j1
+      run: ctest --build-config ${{ matrix.build_type }} --output-on-failure -j4


PascalChristiaanse:
For some reason macos was still failing, it apparently doesnt like forward declarations when unique_ptr has to implicitly deconstruct its target. This move circumvents the forward declaration so it _should_ build now

PascalChristiaanse:
Unsure of what is happening, macos seems to quit after some hours, ubuntu broke again?
Is this possibly a memory issue? on my machine building with 16 cores takes ~48gb of memory so naively 3 should take 9 gb.

PascalChristiaanse:
Not sure what happened in commit #[caab7e4](https://github.com/tudat-team/tudatpy/pull/550/commits/caab7e463484f2a1226f54961e2e1043d2a7a86e) , i didnt intend to merge these branches in this direction, but somehow it was bidirectional...

DominicDirkx:
Not sure either, but I see quite some changes here that I don't think belong in this PR (?) Could you revert to the original state? The macOS build has an issue with RAM usage. Once you've reverted the PR to what it should be I'll tweak a couple of things to reduce the RAM usage and test this.

PascalChristiaanse:
So im fairly sure its in the correct state right now, if you look at the files changed tab, everything should be correct. i reverted my mistake in [d0a91f0](https://github.com/tudat-team/tudatpy/pull/550/commits/d0a91f052a6a082d159900c703a7db10515c5dc4), but i can open a new branch and cherrypick if you'd like?

Issue #539, opened 2025-11-18
[F&E]: Making 80col parser more robust


luigigisolfi:
### Description:

The 80col parser functionality will soon undergo the changes listed below.  

- in [src/tudatpy/data/mpc/parser_80col/parsers.py](https://github.com/tudat-team/tudatpy/pull/505/files/f54a28b1c1007262b3858a59cab0f4092dde949d..5e0bac28a300e6d184c5e22a8b9b422096648000#diff-925d44ad603bdfe017fdc0d756c3aeadc518f5c246081c78287699c8f137f01c): change try/except logic to something more robust. See: https://github.com/tudat-team/tudatpy/pull/505#discussion_r2513699221, https://github.com/tudat-team/tudatpy/pull/505#discussion_r2513711794 and https://github.com/tudat-team/tudatpy/pull/505#issuecomment-3532282706
- make all from_x BatchMPC objects static, see: https://github.com/tudat-team/tudatpy/pull/505#discussion_r2514863955
- create a dataclass to hold the contents of an MPC data file, see: https://github.com/tudat-team/tudatpy/pull/505#discussion_r2472495088:~:text=the%20mpc%20module.-,As%20a%20very%20nice%20to%20have%2C%20but%20not%20needed%2C%20feature.%20I,make%20the%20code%20easier%20to%20read%20and%20write%20compared%20to%20dictionaries.,-Member 

### Additional context:

This is issue was opened to solve comments on [this PR](https://github.com/tudat-team/tudatpy/pull/505#discussion_r2472495088), while also allowing the Atlas 3I example to be merged while still relevant.

PR #552, opened 2025-11-26
Fix/80col parser

luigigisolfi:
This PR will address issue #539.


luigigisolfi:
@alfonsoSR , @DominicDirkx 

Hopefully this is a better version of the parser. A likely non-exhaustive list of changes:

1) the parser is now faster as it makes use of pandas dataframes to perform validation checks on an array of lines, instead of looping over each line of a file
2) the parser is able to accept either 
- a string (corresponding to a path to an 80 col file), 
Namely
```
    file = tmp_path / "test_obs.txt"
   parsed_table = parse_80cols_file(str(file))
```
- a list of strings (the lines themselves)
Namely
```
    # 1. Setup Data: Real observation lines
    line_atlas = '0003I         C2025 07 03.94821918 01 44.983-18 39 35.17         18.0 GW#0JSN598'

    # Eros: Observation ('S') line followed by Parallax ('s') line
    line_eros_valid = '00433         S2021 06 07.42640918 08 15.401-41 22 02.35         12.0 V      500'
    line_eros_skip = '00433         s2021 06 07.4264091 -198301.940 +198171.039 +56287.9850   ~6oMXC57'

    line_charon = 'D4341J79M00A*4A1979 06 25.66181 20 27 06.64 -15 37 11.5          19.0   M4986413'
    line_2025FA22 = '     K25F22A  C2025 10 13.24277 00 20 45.76 +25 53 06.1          18.3 RrET147718'
    
    lines_to_parse = [line_atlas, line_eros_valid, line_eros_skip, line_charon, line_2025FA22]

    # 2. Parse
    parsed_table = parse_80cols_data(lines_to_parse)
```

3) code does not skip invalid lines, it rather fails if any are detected. Unless they belong to `observation_types_to_drop`. As a consequence, try/except blocks were removed.
4) Unit tests were added (including the integration tests comparing BatchMPC output to 80col parser output)
5) Failure handling was improved
6) Docstrings were improved (also adding links to relevant MPC format standards)

src/tudatpy/data/mpc/mpc.py
```diff
@@ -960,13 +960,45 @@ def get_observations(
...
+                            valid_designations = obs['desig'].dropna().astype(str).replace('<NA>', np.nan).dropna()
+                            potential_id = valid_designations.iloc[0]
+
+                        # We allow alphanumeric strings now (to support packed numbers like D4341)
+                        # We only pad if it is a short digit string (e.g. '1' -> '00001')
+                        # Packed strings are always 5 chars long, so they won't be affected by zfill(5)
+                        if len(potential_id) < 5:
+                            potential_id = potential_id.zfill(5)
+
+                        try:
```
alfonsoSR: How does this `try...except` sit within the logic of the function? Is it necessary?

alfonsoSR (commented):
The code looks much better now, and the unit tests seem pretty exhaustive, which is also good. Thanks for taking the time to make all these modifications!

I just made two comments about the `try...except` blocks in `mpc.py`. If you could write down a situation in which not having each of them would be problematic, it would help me understand why they are there :)

DominicDirkx:
(updated code to latest develop)

DominicDirkx:
Is this now ready to merge?

alfonsoSR:
I just checked the reasoning to keep the last bare exception again, and I don't think it is a good one.

Right now, we use the function to find the numerical ID of an object from its `potential_id`, and we assign the output to `identifier`. Assuming that the function can only fail if the object does not have a number, the we assign `identifier = potential_id` in case of an exception.

This behavior should also be achievable by making the function return `potential_id` if the object does not have a number, which should be doable without relying on exceptions. This would remove the last `try...except`, and allow the code to fail if an exception is raised without affecting the behavior of the program.

Aside from this, the code should be ready to merge

DominicDirkx:
If you'd like to make this final modification: could you make the changes to the code?

DominicDirkx:
@alfonsoSR Could you make the modifications you'd like to see? If I recall correctly, that was the workflow we agreed on for this PR

alfonsoSR:
Hi! Sorry about the long delay. I have to focus on something else right now, so I would merge it as it is to avoid leaving it stale :)

Issue #551, opened 2025-11-26
[BUG]: Test Failures (tudat and tudatpy)


minervino96:
### Describe the issue:

On the latest version (1.0.0.dev32), two tudat tests fail (see logs attached):
140 - test_orbit_determination_MultiArcStateEstimation
171 - test_rotational_dynamics_partials_TorquePartials

Also, on the tudatpy side, there are a couple of failures in the `tle_downloding` test, but I think those are known/expected, but sharing them below just in case:

```
No object found satisfying all user-defined filters. Exiting...

INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/Users/ama/Downloads/tudatpy/.pixi/envs/default/lib/python3.13/site-packages/_pytest/main.py", line 289, in wrap_session
INTERNALERROR>     session.exitstatus = doit(config, session) or 0
INTERNALERROR>                          ~~~~^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/Users/ama/Downloads/tudatpy/.pixi/envs/default/lib/python3.13/site-packages/_pytest/main.py", line 342, in _main
INTERNALERROR>     config.hook.pytest_collection(session=session)
INTERNALERROR>     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/Users/ama/Downloads/tudatpy/.pixi/envs/default/lib/python3.13/site-packages/pluggy/_hooks.py", line 512, in __call__
INTERNALERROR>     return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
INTERNALERROR>            ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/Users/ama/Downloads/tudatpy/.pixi/envs/default/lib/python3.13/site-packages/pluggy/_manager.py", line 120, in _hookexec
INTERNALERROR>     return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
INTERNALERROR>            ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #553, opened 2025-11-26
Fix/remove tle downloading tests

luigigisolfi:
This PR addresses the issue with the pytests failure mentioned in #551.


luigigisolfi:
> This looks like an update of the tudatpy-examples submodule (?) I just updated the examples submodule to the latest commit here: https://github.com/tudat-team/tudatpy-examples (which is what it should be, unfortunately submodules can only be linked to a commit, not a branch).

Thanks! How does this reflect on the tudatpy repo though? I still see the examples submodule pointing at the outdated branch

PR #555, opened 2025-12-02
Feature/iau pole libration parameters

jo11he:
PR introduces pole libration amplitudes of the IAU rotation model as estimatable parameters.
Unit test included and passing.


DominicDirkx:
Thanks for adding!

PR #560, opened 2025-12-04
Feature/exponential atmosphere parameters

jo11he:
implements estimatable parameters of the exponential atmosphere model (via aerodynamic acceleration partials)

### DIFF
--- a/src/tudat/astro/orbit_determination/acceleration_partials/aerodynamicAccelerationPartial.cpp
+++ b/src/tudat/astro/orbit_determination/acceleration_partials/aerodynamicAccelerationPartial.cpp
@@ -158,23 +158,29 @@ void AerodynamicAccelerationPartial::computeAccelerationPartialWrtArcWiseAerodyn
  */
 void AerodynamicAccelerationPartial::computeAccelerationPartialWrtCurrentDensity( Eigen::MatrixXd& accelerationPartial )
 {
+    // Full computation for (near-)zero density
+    if( flightConditions_->getCurrentDensity( ) < 1.0E-30 )
+    {
+        Eigen::Vector3d currentForceCoefficients = flightConditions_->getAerodynamicCoefficientInterface( )->getCurrentForceCoefficients(  );
+        std::pair< reference_frames::AerodynamicsReferenceFrames, double > coefficientsFrameAndSign =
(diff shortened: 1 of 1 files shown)


include/tudat/astro/orbit_determination/estimatable_parameters/exponentialAtmosphereParameter.h
```diff
@@ -0,0 +1,153 @@
...
+     *  \param exponentialAtmosphereModel Class defining properties of exponentialAtmosphereModel used for body.
+     *  \param associatedBody Body associated with atmosphere mdoel.
+     */
+    ExponentialAtmosphereParameter( const std::shared_ptr< aerodynamics::ExponentialAtmosphere >& exponentialAtmosphereModel,
+                                    const estimatable_parameters::EstimatebleParametersEnum parameterType,
+                                    const std::string& associatedBody ):
+        EstimatableParameter< double >( parameterType, associatedBody ), exponentialAtmosphereModel_( exponentialAtmosphereModel ),
+        parameterSize_( 1 )
+    {
+        if( ( parameterType != exponential_atmosphere_base_density ) && ( parameterType != exponential_atmosphere_scale_height ) )
```
jo11he: I chose to handle both parameters through the same EstimatableParameter class, Settings class etc, only the factory function for the parameter settings object differs. I thought that was neat, but maybe not desired for other reasons?

src/tudat/astro/orbit_determination/acceleration_partials/aerodynamicAccelerationPartial.cpp
```diff
@@ -146,6 +146,73 @@ void AerodynamicAccelerationPartial::computeAccelerationPartialWrtArcWiseAerodyn
...
+
+//! Function to compute the partial derivative of the acceleration w.r.t. the current atmospheric density
+/*!
+ * Function to compute the partial derivative of the acceleration w.r.t. the current atmospheric density
+ * \param accelerationPartial Derivative of acceleration by reference.
+ */
+void AerodynamicAccelerationPartial::computeAccelerationPartialWrtCurrentDensity( Eigen::MatrixXd& accelerationPartial )
+{
+
+    Eigen::Quaterniond rotationToInertialFrame = flightConditions_->getAerodynamicAngleCalculator( )->getRotationQuaternionBetweenFrames(
```
DominicDirkx: This will only work correctly if the force coefficients are in the aerodynamic frame. I'd recommend adding a function to the flight conditions to get the coefficients in teh inertial frame (since it knows which frame the original coefficients are in)

Issue #382, opened 2025-07-25
[F&E]: Mars Climate Database (MCD) Exposure


nikemais:
### Description:

Exposure of the Fortran MCD source code to retrieve a list of atmospheric variables to be used for different acceleration models. 

### Suggestions for implementation:

The current working implementation is used to retrieve density values inside a new atmospheric model.

The procedure has many steps (in the future could be automated):

- install NetCDF-c (https://github.com/Unidata/netcdf-c): required to build the fortran version
- install NetCDF-fortran (https://github.com/Unidata/netcdf-fortran): required by the Fortran MCD source code to access the scientific data
- retrieve the source code (MCD_6.1/mcd/MCD.f90), the header file with the exposure of `call_mcd` ( MCD_6.1/mcd/interfaces/c_interfaces/mcd.h), and the data folder (MCD_6.1/data) with the desired addons. All of this can be obtained from https://www-mars.lmd.jussieu.fr/MCD_pro/mcd_pro.html upon request (which is immediately accepted)

The current implementation is based on the old repo structure:
- MCD.f90 lives in a directory called Fortran next to Tudat and Tudatpy, which must be compiled first by adding it in `tudat-bundle/CMakeLists.txt` before `add_subdirectory(tudat)`
- in `tudat/cmake_modules/TudatLinkLibraries.cmake` add before `list(APPEND Tudat_PROPAGATION_LIBRARIES ...)` the line: `list(APPEND TUDAT_EXTERNAL_INTERFACE_LIBRARIES fortran_lib)` 
- the exposure and wrappers are in `tudat/include/tudat/interface` and can be customised (currently 86 variables can be extracted under different settings)
- the current simple implementation only retrieves density and is used in the new `marsClimateDatabaseAtmosphericModel`, the exposure requires the user to specify some parameters and the location of the data folder

### Additional context:

I am not sure of the implications of copyright for the source code, which can be retrieved upon request (which is immediately approved). Nevertheless, I think it is important to double-check or ask directly to the developers of MCD, as having this in Tudat could allow the use of many cool parameters in multiple models!

DominicDirkx:
This is really cool (@valeriof7 , @RivaAlkahal )! Since the MCD code is only available upon request, we can't have it as a 'standard' dependency, but we can write the CMakeLists for Tudat (in the final version) such that it enables the MCD support if it finds the package (I think)

nikemais:
[...]

PR #561, opened 2025-12-04
Feature: Mars Climate Database Atmosphere models

valeriof7:
PR for issue #382


DominicDirkx:
Great to add this! A couple of questions:
* Are any additional package dependencies needed except for mcd? I see netcdf, fortran compilers, and maybe other things(?)
* Why does this PR  say it consists of a total of 6306 commits? The changed files seem fine, so seems to be some git-related messiness
* Any idea how to test this in our CI? We could put mcd in a private repo, and give the test access to the private repo to clone it as part of the test (?) Is the implementation platform-independent?

valeriof7:
In order:
- I have summarized everything in issue #382 . In brief, the dependencies are Fortran and NetCDF-Fortran.
- Yes, that's exactly what happened. I committed the source code and data of MCD; therefore, I used Git filter-repo to remove the files from the commits. I think they were somehow modified in the hash. I'd need help in case this needs to be fixed.
- Not really, to be honest, because many steps have to be performed manually. The dependencies listed above are platform dependent, and the MCD source code and data have to be requested by each user. As far as my cmake knowledge goes, the implementation should be platform independent.

valeriof7:
I know for sure on linux/macOs, can't confirm on Windows

DominicDirkx:
(updated branch to latest develop)

valeriof7:
I know that Nicoló is working on a CI workflow to install the necessary dependencies. Should we wait to test issue #419?

DominicDirkx:
You can go ahead and test 419 with this branch, if it's succesful in the current setup we can merge it. The automated testing will be a separate point to handle

nikemais:
Since more models in the future will rely on MCD (for example, PRP),  I think MCD should have its own class and updater, which gets called by the  EnvironmentUpdater. This way, only one call of MCD (with the correct variable keys set during setup) has to be made at each step, and then the models can share and access the requested variables

DominicDirkx:
Agreed, to be added in a next PR (?)

nikemais:
Yes, we can merge this. I will create the issue for the next PR with the list of features

valeriof7:
Before the PR, someone else needs to take a detailed look at the implemented caching logic and the unit tests. The former is based on Section 5.5 of the [UserGuide](https://www-mars.lmd.jussieu.fr/mars/info_web/user_manual_6.1.pdf), and the latter is based on the ten unit tests provided in the MCD software directory `MCD_6.1/mcd/testcase`.

Regarding the last comment, as I mentioned to Nicoló offline, the new implementation of the call should be easy to transfer to a dedicated class that wraps the call to `mcd.f90`.

nikemais:
I will have a look at the caching logic and unit tests (especially to understand the minimal amount of data they can run on)

nikemais:
I had some time to review the caching logic, and everything looks good. 

I'd like to express a small concern regarding the input altitude. In Tudat, the computation depends on the type of shape model chosen for Mars, which defaults to a spherical model. By setting MCD's zkey=3 (altitude above local surface, computed with precise models) and by not performing a conversion on the input altitude (for example, to an oblate-spherical altitude), it can lead to noticeable differences. To solve this, I suggest the use of zkey=1 (radial distance from the centre).

Aside from this, I think this could be merged! The next PR will focus on CI testing for MCD, followed by restructuring.

DominicDirkx:
Great, and agreed on the default setting

valeriof7:
I think I chose this option for two reasons:
- First, if you set the distance from the center, you lose the ability to model the body's shape as you wish. What do you mean by "not performing a conversion"? As far as I know, the flight condition object knows which shape model to use for geodetic altitude computation, so everything should be consistent.
- Second, to replicate the provided test cases, which use zkey=3. If we change that, we need to remember to update the unit tests as well.

DominicDirkx:
Having radial distance to center of Mars is more robust, I think. Presumably MCD has an internal model for the Martian shape (since it can work with distance to center of Mars). Any shape model we use will be inconsistent with this model (even if only slightly). To make sure the atmosphere model produces the correct density at a given state, the distance to Mars center of mass is better. It would be good to make this user-configurable, but this can be done in a later step.

What would updating the unit tests entail? Are you reusing tests from the MCD library that have zkey=3? If so, then let's add an option now to make this settable in our atmosphere model. Then the tests can use zkey=3, but the default can go to zkey=1

Also, I see there are many changes to code files that are unrelated to this issue. Could you modify the PR so that the changes are only those that address #382 ? Thanks!

valeriof7:
It sounds like a good plan! 

Regarding the PR, all the additional files come from the merge I did of the develop branch before the PR. Specifically, all the changes come from correcting the misspelled ancillary word. How is this possible?

DominicDirkx:
Great! Which one of you will make the changes?

For the branches, not sure what happened but you probably branched off of your local branch with changes rather than the develop branch. With a little luck this can be merged soon: https://github.com/tudat-team/tudatpy/pull/606 and there will be no more inconsistencies

nikemais:
I have been working on the next PR, reusing and refactoring this code (essentially incorporating the changes from this one). Tomorrow or early next week, I plan to create the draft, and then we can decide how to move forward (after #606 is merged, anyway).

DominicDirkx:
@nikemais So, shall we put this proposed modification of the zkey in the next PR that you're now working on?

nikemais:
Yes exactly. It would also supersede this PR, as the structure is different but with the same features. Ideally, we could also close this one if we are happy with the new structure (I want to stress that the work was not in vain and has been reused). However, it's nothing a merge cannot resolve, so merging this one into develop first is also an option.

valeriof7:
I'm not following. Nicoló, could you please explain in detail what you're doing and how? Is it this #586? If so, please merge mine first and then modify.

Also, remember to use the board and move the issue to "In Progress." Otherwise, it is difficult to follow what is being worked on and by whom.

nikemais:
Yes, it's for #586. We can definitely merge this first.

In my PR, mcdAtmosphereModel remains similar and requires a MarsClimateDatabase object, which comprises the MCD exposure and caching logic. The idea is that MCD becomes an attribute (a new climateModel variable) of the planet Body object so that all the models that require it can access the same one, which gets updated only once. More info and a proper explanation will be in the PR.

DominicDirkx:
@valeriof7 Your changes to spelling of ancillary settings (and some other auto-formatting?) are leading to a lot of merge conflicts. Could you remove these changes from the branch, or resolve the merge conclifts in favor of the develop branch?

valeriof7:
Easier than expected. Now the branch is ready to be merged before Nicoló's PR.

DominicDirkx:
Thanks!

PR #565, opened 2025-12-07
Fix constant aerodynamic coefficient

### LINKED ISSUE #486: [BUG]: Constant drag coefficient formulation
### Describe the issue:

In `constantDragCoefficient.h` the function `setParameterValue( )` modifies the first entry of the coefficient vector. 

This however, has the intended effect only if the original coefficient set is in the aerodynamic frame.

A possible solution would require changing the class structure to allow for `aerodynamicAcceleration` to be stored, from which to retrieve the rotarion matrix between the defined frame and the aerodynamic one, so that the function can actually modify the vector entry related to drag.

`constantDragCoefficient.h` row 63

### OS:

OS : Linux

### Expected behavior:

_No response_

### Current behavior:

_No response_

### Additional context:

_No response_

nikemais:
This PR closes #486.

The fix involves rotating the set of coefficients specified by the user in the chosen frame to the aerodynamic frame, where the first entry corresponds to the drag coefficient. This allows for the correct estimation of Cd.

For simplicity, for the arcwise parameter, the fix is to throw an error in case the chosen frame is different from the default negative aerodynamic frame.

### DIFF
--- a/include/tudat/astro/aerodynamics/aerodynamicAcceleration.h
+++ b/include/tudat/astro/aerodynamics/aerodynamicAcceleration.h
@@ -96,8 +96,6 @@ public:
         if( !( this->currentTime_ == currentTime ) )
         {
             currentTime_ = currentTime;
-            coefficientInterface_->setRotationToAerodynamicFrame( flightConditions_->getAerodynamicAngleCalculator()->getRotationQuaternionBetweenFrames(
-                aerodynamicCompleteCoefficientFrame_, reference_frames::aerodynamic_frame ) );
             currentForceCoefficients_ = coefficientInterface_->getCurrentForceCoefficients( );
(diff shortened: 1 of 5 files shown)


nikemais:
I was about to comment on this as well, I'm not sure either. On top of that, the arc-wise coefficient cannot be solved the same way, so I opted for a simpler warning.

To solve the issue nicely (and to keep it consistent with, for example, the scaling coefficients), I could rewrite everything to have `std::vector< std::shared_ptr< aerodynamics::AerodynamicAcceleration > >` as a member in ConstantDragCoefficient, this way, we could access both AerodynamicAngleCalculator and AerodynamicCoefficientInterface. 

This would definitely be more complex but nicer overall (fixing arc-wise parameter as well). Let me know if you think this is worth doing!

Issue #591, opened 2025-12-23
[BUG]: Arc-wise aerodynamic scaling factors missing from position partial creation


valeriof7:
### Describe the issue:

When attempting to estimate arc-wise drag and lift component scaling factors in orbit determination, the following runtime error occurs:

```
RuntimeError: Parameter 59 not implemented when making position partial
```

This error is thrown because the arc-wise aerodynamic scaling factor parameters are not handled in the switch statement in createCartesianStatePartials.cpp.

The new estimable parameters have been added in PR #536. @jo11he, please let me know if I missed something or if this is the case for you.ll.

### How to reproduce the issue:

1. Set up an estimation problem with arc-wise aerodynamic scaling factors as parameters to estimate
2. Specifically, use any of:
   - `arc_wise_drag_component_scaling_factor`
   - `arc_wise_side_component_scaling_factor`
   - `arc_wise_lift_component_scaling_factor`
3. Attempt to run the estimation
4. The error is triggered when creating Cartesian state partials w.r.t. parameters

### OS and Python version:

Linux, python 3.11.14

### Expected behavior:

The arc-wise aerodynamic scaling factors should be handled similarly to their constant (non-arc-wise) counterparts, which already exist in the code with `break` statements for the double parameters (since they don't require position partials, only acceleration partials).

### Current behavior:

The code throws a runtime error because the arc-wise versions are missing from the switch statement in the `createCartesianStatePartialsWrtParameter` function for vector parameters.

### Additional context:

the arc-wise versions (`arc_wise_drag_component_scaling_factor`, `arc_wise_side_component_scaling_factor`, `arc_wise_lift_component_scaling_factor`) are missing from the switch statement of vector parameters, causing the default case to throw the error.

Note: The acceleration partials ARE correctly implemented in aerodynamicAccelerationPartial.cpp at lines 480-488.

## Fix:
Add the missing arc-wise cases to the switch statement in createCartesianStatePartials.cpp around line 350 (after the `full_acceleration_scaling_factor` case):

```cpp
case estimatable_parameters::arc_wise_drag_component_scaling_factor:
    break;
case estimatable_parameters::arc_wise_side_component_scaling_factor:
    break;
case estimatable_parameters::arc_wise_lift_component_scaling_factor:
    break;
```

This fix has been tested and resolves the runtime error, allowing arc-wise aerodynamic scaling factors to be estimated successfully. Fix in PR #592

jo11he:
[...]

PR #592, opened 2025-12-23
Bug/arcwise aero parameters

valeriof7:
PR for issue #591


DominicDirkx:
@valeriof7  Thanks for addressing this! This PR seems to have a lot of changes to the examples directory, could you modify it so that it only modifies the code to address #591 ?

DominicDirkx:
I had a look at the code, and streamlined the creation of these partials so that not all parameters have to be handled in the switch statement, only the (few) ones where the direct cartesian state partial is non-zero. I've opened #596 to replace this one.

Thanks for spotting the issue and identifying the location of the error!

Issue #593, opened 2025-12-28
[BUG]: DISCOS api is using outdated (broken) scheme


PascalChristiaanse:
### Describe the issue:

Related to #532. Python api suggests that lookup is by norad ID, instead it is by Discos id. This is caused by a new version of the discos API which uses a different query scheme.

### How to reproduce the issue:

Queries follow:
 response = requests.get(
            f'{self.url}/api/objects/{norad_id}',
            headers=self.headers
        )
an example could be the CAPSTONE satellite with norad id 52914 which can be looked up like (according to the api)
https://discosweb.esoc.esa.int/api/objects/52914
This will instead return the object "Titan 3C fragmentation debris". 

### OS and Python version:

Ubuntu 24.04 LTS, python 3.11.14

### Expected behavior:

should return according to norad id (as python describes)

### Current behavior:

returns according to discos id

### Additional context:

the fix is very easy:
Requests should follow the new forma: query a list of objects with filter applied. only return the first object from the resulting json data
example: 
CAPSTONE: https://discosweb.esoc.esa.int/api/objects?filter=eq(satno,52914)

/tudatpy/data/discos/discos.py:
line 28:9 to 32:66
response = requests.get(
            f"{self.url}/api/objects?filter=eq(satno,{norad_id})", headers=self.headers
        )
        if response.ok:
            attributes = response.json()['data'][0]['attributes']

DominicDirkx:
@luigigisolfi This sounds like something for you

luigigisolfi:
I implemented a fix by allowing for both discos_id and norad_id queries, while keeping backwards compatibility. Docstrings have been updated. 
Unit tests will follow :)

Now you can use it like:

```
    # Default behavior 
    iss_data = discos.query_object(25544) #this is understood as a norad_id (satno)

    # We pass the flag is_discos_id=True
    some_obj = discos.query_object(25544, is_discos_id=True) 
```

PR #594, opened 2026-01-05
Fix Discos Query

luigigisolfi:
Solves #593 by adding a flag to the `sat_id` type one wants to query (now both 'discos_ids' or 'norad_ids'=='satno') can be queried. Flag defaults to `norad_id`.


DominicDirkx:
Can be merged once a unit test is added (?)

luigigisolfi:
I added some simple unit tests. 

What they test (and comments on how they test it):

- `test_query_norad_success`: tests Successful NORAD ID query (tested by mocking a successful discos API request using the `objects?filter=eq(satno, norad_id)` mode)
- `test_query_discos_success`:  tests Successful DISCOS ID query (tested by mocking a successful discos API request using the `objects/discos_id` mode)
- `test_query_not_found`: tests invalid satellite norad and discos id (mocks empty API return)
- `test_api_error`: tests failure against API failure (Mocks API failure or bad request)

> Note
Mocks (from unittest.mock import MagicMock, patch) are needed because we do not want to hit the API when running unit tests.

DominicDirkx:
Thanks! To run these tests you need a Discos login right?

luigigisolfi:
> Thanks! To run these tests you need a Discos login right?

You dont, because I mocked the API query. I am simulating an API response using Mock objects. 
It does require some care and extra work to create such mock objects, but it avoids hitting the API and allows everyone to run the tests. 
If we like this solution, this is how I am planning to implement the spacetrack tests as well

DominicDirkx:
Nice! But this will only allow for partial testing right (?)

Does this require any new conda package dependencies?

luigigisolfi:
> Nice! But this will only allow for partial testing right (?)

What it tests: 
1) correct url creation
2) correct values retrieval upon successful API response. 
3) returns None if bad API response or invalid id

My reasoning is: since the url is checked, the only way the query can fail is by getting an authentication error (wrong or non existing credentials, restricted use) or connection failure. But this is not tested, as this would test the API connection, not the functionality itself.

In nominal, successful API responses, the two types of outputs we get from the API are the two I hardcoded (one querying by `norad_id` and one querying by `discos_id`, and all other objects will come with the same data structure upon successful response). So I'd say we are testing what we should. Does it make sense?

> Does this require any new conda package dependencies?

Does not require conda package because I am using unittest, which comes with the standard python library

DominicDirkx:
Thanks, that's clear. Don't get me wrong, I think this is a good way to test it without going through the hassle of setting up a github token that can access discos (is this even possible and allowed?). But this:

> In nominal, successful API responses, the two types of outputs we get from the API are the two I hardcoded 

is the one thing that is not tested, if these things that should be the same actually will be the same. But I think this is a very minor thing. Since the Github CI doesn't run the Python tests, I'll build this locally and merge it if it's succesful

PR #596, opened 2026-01-05
Simplified partial creation switch

DominicDirkx:
Addresses #591; replaces PR #592


jo11he:
Thank you Dominic for this fix, this makes more sense to me now!

PR #598, opened 2026-01-07
Feature/differenced frequency of arrival

PascalChristiaanse:
Feature/Differenced Frequency of Arrival

This PR introduces two new observation models:
- one way doppler measured frequency
- one way differenced frequency of arrival

The latter is an aggregation of the former, taking the difference between two of these models. This observable is produced in certain VLBI setups (such as by the University of Tasmania)

Please feel free to provide feedback! (It'll take a few weeks before i have time to implement suggestions, internship reports/exam week etc...)

The one way frequency of arrival model currently uses an aliased settings object inherited from the one_way_doppler model. Designwise im not sure what would be the best choice, for now i made the specialization to keep it more verbose as to whats happening. 
I also have some doubts about the normalization factor in include/tudat/simulation/estimation_setup/createPositionPartialScaling.h (note the comment). I saw a few times that a 1/c factor is applied, but doing so gives the wrong result in the partial. The way its set up now converges with both simulated and actual data when tracking Danuri (Capstone works in simulations, currently i dont have enough data for an actual OD)


PascalChristiaanse:
The unit test _should_ work and should pass, but building with tests takes literal hours on my machine (its has been running for 3 hours now, at 74 percent...)

PascalChristiaanse:
Still need to update documentation, holding of on committing this though as I noticed theres been an update to the naming scheme. Im also leaning towards desolving the specialized settings object

DominicDirkx:
Nice! I don't see any note in the code, could you repeat it in a comment? I see that you have a unit test for the new differenced frequency observation model, could you also add one for the new one-way frequency model, and the new partials? For the partials test, there is a (kind of) standard way to test new observation model partials, but it's a little old/messy, so maybe easier to have a chat.

Really cool the model works for orbit estimation from real data! Would love to see more :)

DominicDirkx:
@PascalChristiaanse Small reminder :)

PascalChristiaanse:
@DominicDirkx Any idea why these tests are failing? it seems unrelated to my changes.

DominicDirkx:
These failing tests:

	137 - test_observation_models_DifferencedFrequencyOfArrivalObservationModel (Failed)
	192 - test_observation_partials_TimeBiasPartials (Failed)
	195 - test_observation_partials_OneWayDopplerMeasuredFrequencyPartials (Failed)
	196 - test_observation_partials_DifferencedFrequencyOfArrivalPartials (Failed)

do seem conceptually relate to your implementation, so it seems likely that your modifications are the cause of this. Are they failing for you locally?

PascalChristiaanse:
@DominicDirkx previous test fail was my silly mistake; this time it seems like Mac just failed. Ubuntu and Windows are passing all, as it also is on my machine. 

Im guessing we restart?

DominicDirkx:
@PascalChristiaanse Will be really cool to get this merged into Tudat! To help me out in reviewing it, could you summarize your testing strategy (e.g. what exactly is check to pass the unit tests of the new observation models, partial tests are pretty clear)

tests/test_tudat/src/astro/observation_models/unitTestDifferencedFrequencyOfArrivalObservationModel.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+    std::shared_ptr< ObservationModelSettings > frequencyObservableSettings2 =
+            oneWayDopplerMeasuredFrequencySettings( linkEnds2, lightTimeCorrectionSettings );
+
+    for( int testTimeScale = 0; testTimeScale < 1; testTimeScale++ )
+    {
+        std::shared_ptr< ObservationModelSettings > differencedObservableSettings = differencedFrequencyOfArrivalObservationSettings(
+                differencedLinkEnds,
+                lightTimeCorrectionSettings,
+                ( testTimeScale == 0 ) ? basic_astrodynamics::tdb_scale : basic_astrodynamics::utc_scale,
```
DominicDirkx, line 111: why only test for TDB-referenced observations?

tests/test_tudat/src/astro/observation_models/unitTestDifferencedFrequencyOfArrivalObservationModel.cpp
```diff
@@ -0,0 +1,168 @@
...
+                linkEndTimesDifferenced.at( 3 ), linkEndTimesRange2.at( 1 ), ( 4.0 * std::numeric_limits< double >::epsilon( ) ) );
+
+        TUDAT_CHECK_MATRIX_CLOSE_FRACTION(
+                linkEndStatesDifferenced.at( 2 ), linkEndStatesRange2.at( 0 ), ( 4.0 * std::numeric_limits< double >::epsilon( ) ) );
+        TUDAT_CHECK_MATRIX_CLOSE_FRACTION( linkEndStatesDifferenced.at( 3 ), linkEndStatesRange2.at( 1 ), 2.0E-14 );
+
+        std::cout << ( firstFrequency - secondFrequency ) << " " << differencedFrequencyOfArrival << std::endl;
+        std::cout << ( firstFrequency - secondFrequency ) - differencedFrequencyOfArrival << std::endl;
+
+        if( testTimeScale == 0 )
```
DominicDirkx, line 166: See comment above: why only test for TDB-referenced observations?

DominicDirkx:
I added a bunch of comments, but nothing major (main thing is the absence of UTC-referenced data, but on second thought that is not relevant here, I had it confused with other functionality; for now, make sure an error gets thrown if the time scale gets set to UTC)

PascalChristiaanse:
**Testing strategy summary:**

**One-Way Doppler Measured Frequency (`unitTestOneWayDopplerMeasuredFrequencyObservationModel.cpp`) — 4 tests:**

1. **Core observable verification** — Creates both a measured-frequency model and a plain one-way Doppler model (normalized by c). Computes both at the same epoch and verifies that `f_rx = f_tx × (1 + D)` holds to machine precision (~4ε). Also checks that link-end times and states are consistent between both models.

2. **Reference link-end validation** — Verifies that calling the model with `transmitter` as the reference link end throws a `std::runtime_error` (only `receiver` is valid for this observable).

3. **Observation biases** — Applies combined absolute (1 kHz) and relative (1 ppm) biases and verifies `biased = absoluteBias + (1 + relativeBias) × ideal` to ~1e-15 precision.

4. **Multi-epoch consistency** — Repeats the core `f_rx = f_tx × (1 + D)` check at 5 different epochs spread across a 7-day simulation window, confirming the relationship holds regardless of geometry.

All tests use an Earth-Mars link (DSS-13 → Mars) with X-band (8.4 GHz) constant transmitter frequency and first-order relativistic light-time corrections (Sun perturbation).

**Differenced Frequency of Arrival (`unitTestDifferencedFrequencyOfArrivalObservationModel.cpp`) — 1 test:**

Sets up two independent one-way measured-frequency models (DSS-13←Mars and DSS-43←Mars) and a differenced model (DSS-13, DSS-43 ← Mars). At a single epoch, it verifies:
- Link-end times/states from the differenced model match those from the individual sub-models (transmitter/receiver pairs align)
- The differenced observable equals `f_rx1 - f_rx2` to ~5e-13 precision

**Partial tests** are registered in CMakeLists but follow the standard Tudat pattern of comparing analytical partials against numerical finite differences.

Both models currently only support TDB-referenced observations. I've now added validation at the settings level so that requesting UTC throws a clear error immediately at construction (previously this was only caught at model creation for DFOA and not at all for OWDMF).

PascalChristiaanse:
@DominicDirkx only took 3 months but I think this is finally ready to merge?

DominicDirkx:
@PascalChristiaanse Awesome! Looking close mergable :) There are a bunch of comments from me and (mostly) copilot left above, many of which seem valid (although most are easy to address). Could you modify the code, or specify why the point is not valid, for these?

PascalChristiaanse:
@DominicDirkx The test that is failing on mac is unrelated to the model, other than that i think its done?
UTC support has been added, i think correctly? I think its best if you check #[b82254d](https://github.com/tudat-team/tudatpy/pull/598/commits/b82254d6ba019132953004906f487fa6cc2eb332)[b82254d](https://github.com/tudat-team/tudatpy/pull/598/commits/b82254d6ba019132953004906f487fa6cc2eb332) to make sure. 

i fixed the timescale issues chatgpt pointed out too, that should now all be consistent

DominicDirkx:
Indeed the Mac failure is not an issue introduced here. 

A few small points on the code:

OneWayDifferencedFrequencyOfArrivalObservationModel does not need groundStationStates, only the constituent OneWayDopplerMeasuredFrequencyObservationModel do. In the OneWayDifferencedFrequencyOfArrivalObservationModel constructor, it is important to check if the time scales for the OneWayDifferencedFrequencyOfArrivalObservationModel is the same as that for the two OneWayDopplerMeasuredFrequencyObservationModel objects

Similarly, it should be checked if the OneWayDopplerMeasuredFrequencyObservationModel and underlying OneWayDopplerObservationModel are in the same time scale.

This call:

ObservationScalarType transmittedFrequency =
                frequencyInterpolator_->template getTemplatedCurrentFrequency< ObservationScalarType, TimeType >( transmitterFrequencyTime );


has the input time in the wrong time scale. Right now, it is *always* in TDB, but it should always be in UTC

include/tudat/astro/observation_models/oneWayDopplerMeasuredFrequencyObservationModel.h
```diff
@@ -0,0 +1,265 @@
...
+        Eigen::Vector3d nominalTransmittingStationState = ( stationStates_.count( transmitter ) == 0 )
+                ? transmitterPosition
+                : stationStates_.at( transmitter )->getNominalCartesianPosition( );
+
+        TimeType transmitterFrequencyTime = transmitterTime;
+        if( observableTimeScale_ == basic_astrodynamics::utc_scale )
+        {
+            transmitterFrequencyTime = timeScaleConverter_->template getCurrentTime< TimeType >(
+                    basic_astrodynamics::tdb_scale, basic_astrodynamics::utc_scale, transmitterTime, nominalTransmittingStationState );
+        }
```
PascalChristiaanse, line 213: @DominicDirkx: if the time scale is UTC, we assume the timescale is already correct for the frequency interpolator? This should be changed to only run the conversion if the timescale is not UTC but TDB?

DominicDirkx:
And merged!

Issue #507, opened 2025-10-15
[F&E]: Time Conversion Issues for Legacy MPC Observations


luigigisolfi:
### Description:

Functions from the **`DateTime`** module do not work before a given date (1958-09-18 19:31:01). 
This is not a problem for most applications, but it arises whenever we want to MPC-query objects like **Eros**, for which observations are available starting from 1893-10-29.

For context, I would like to change the following code:

```
                # convert JD to J2000 and UTC, convert deg to rad
                obs = (
                    obs.assign(
                        epochJ2000secondsTDB=lambda x: (
                            Time(x.epoch, format="jd", scale="utc").tdb.value
                            - 2451545.0
                        )
                        * 86400
                    )
                    .assign(RA=lambda x: np.radians(x.RA))
                    .assign(DEC=lambda x: np.radians(x.DEC))
                    .assign(epochUTC=lambda x: Time(x.epoch, format="jd").to_datetime())
                )
```

to the following (faster and more robust, leveraging tudatpy's DateTime)

```
                dt_objects = [DateTime.from_julian_day(jd) for jd in obs.epoch]

                # Convert DateTime objects to epochJ2000secondsTDB
                obs['epochJ2000secondsTDB'] = [
                    time_scale_converter.convert_time(
                        input_scale=time_representation.utc_scale,
                        output_scale=time_representation.tdb_scale,
                        input_value=dt_obj.epoch()
                    ) for dt_obj in dt_objects
                ]

                # Convert DateTime objects to Python datetime objects for epochUTC
                obs['epochUTC'] = [DateTime.to_python_datetime(dt_obj) for dt_obj in dt_objects]
```

However, this is not possible because it would break Eros's estimation example, since the current way to query an object from the MPC only allows date filtering AFTER querying all observations for the object itself. In other words, the mpc wrapper will ask for all dates and only then one could perform: `batch.filter(start_date,end_date)`

This breaks the workflow of the `get_observations()` function in `mpc.py`.

luigigisolfi:
The current sofa wrapper handles conversions flawlessly but, unlike the previous (`astropy`) time conversion functions used in the mpc wrapper, it throws an error if a date is far in the past (in this case, very far = 1893-10-29, date of the first Eros observations):

```
An error occurred while retrieving observations of MPC: 433
Provided caledar date cannot properly give Delta AT1893 10 29
```

**CORRESPONDING SOURCE CODE:**

```
> Get number of leap seconds and check feasibility of calculation
[...]

Issue #566, opened 2025-12-08
[BUG]: Remove python datetime objects from mpc.py and parsers.py


luigigisolfi:
### Describe the issue:

Both files mentioned above use python datetime objects.
To solve issues #507 and #501 , it is sufficient to avoid the creation of  such objects

The idea is to remove this dependency by using tudat's DateTime instead.
This will include some refactoring of both files, especially mpc.py. 

The codes are now relying on:
`datetime(year, month, day, hour, second, fraction of second)`

these needs to be changed to:
`DateTime(year, month, day, hour, decimal_second)`

for objects created using `datetime.timedelta`, we need to go through some conversions. For instance:
`obs_time_utc = datetime(year, month, day, hour, second, fraction of second) + timedelta(days=day_frac)`

becomes something like:
```
obs_time_utc_seconds = DateTime(year, month, day, hour, decimal_second).to_epoch() + wanted_number_of_seconds
obs_time_utc = DateTime.from_epoch(obs_time_utc_seconds)
``

### How to reproduce the issue:

The following historical date 
`DateTime(1800,1,1,1).to_python_datetime() `

(present - for instance - in some MPC observations of Eros)
cannot be converted to a python datetime and results in:

```
>>> DateTime(1800, 1, 1, 1).to_python_datetime()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError:  Date 1800-01-01 01:00:00.000 is out of range for conversion to time point. Lower limit (in seconds from J2000) is: -3155716800.000000, upper limit: 9222425308854.775391
```


### OS and Python version:

OS: MacOS ARM64, Python 3.11.13

DominicDirkx:
@luigigisolfi To implement/test this, could you add a specific script or piece of code that right now does not work, but which this solving issue would resolve

PR #604, opened 2026-01-12
Remove python datetime depedency from mpc.py

luigigisolfi:
This PR:

1) implements a solution to #507 by removing all python datetime depedencies from `mpc.py`. 
2) fixes some previous bugs in `get_weights_VFCC17()  `


DominicDirkx:
Thanks! Issue #507 got superseded by #566,can you confirm that your modifications address the point there? 

How have you tested the update to the code? Can you confirm that it now works well for older data (e.g. where the time conversion was originally failing)?

luigigisolfi:
This PR removes **all** `python datetime` usages form `mpc.py`. 
Users will still be able to filter observations using `python datetime` objects (using the `filter()` method). 
These will be then internally converted to `iso_strings`, then to Tudatpy DateTime objects.

Checks performed: 

- updated versions (will add those to a different `tudatpy-examples` PR) of `improved_estimation_with_mpc` and `estimation_with_mpc` run as expected
- updated python unit tests (within this PR) run as expected 

This PR solves issue #566, although with a slightly different solution (which was more straightforward imo).

DominicDirkx:
Nice! Since the PR is no longer a draft, I assume it's finished as far as you're concerned?

In the unit test, could you add a check that would have failed for the current tudatpy version, but that will work after this PR (for instance the processing of old data). I had a look at the diff, and it didn't look like this was added (but I could be mistaken)

DominicDirkx:
> These will be then internally converted to iso_strings, then to Tudatpy DateTime objects.

That's nice! If I remember correctly, the C++ type that the datetime automatically converts to using pybind11 was the problem (?) By converting to Tudatpy DateTime this is nicely fixed with no impact on user interfaces

luigigisolfi:
> Nice! Since the PR is no longer a draft, I assume it's finished as far as you're concerned?

Yes, and I have added a unit test now, showcasing one example of code that would have failed before my changes

> If I remember correctly, the C++ type that the datetime automatically converts to using pybind11 was the problem (?) By converting to Tudatpy DateTime this is nicely fixed with no impact on user interfaces

Yes and correct :)

luigigisolfi:
@DominicDirkx what about this PR?

DominicDirkx:
As a check: the examples all run properly after these changes, and the results don't change (appreciably) right?

Could you add a quick top-level overview of which (if any?) changes are functional (conversions UTC <-> TDB removed/added) and which ones are done in a different manner (computations done through Tudat's DateTime rather than Python's datetime

luigigisolfi:
> As a check: the examples all run properly after these changes, and the results don't change (appreciably) right?

Examples results are very similar to the existing ones. 

> Could you add a quick top-level overview of which (if any?) changes are functional (conversions UTC <-> TDB removed/added) and which ones are done in a different manner (computations done through Tudat's DateTime rather than Python's datetime

## Functional changes:

- In `get_biases_EFCC18`:
```
-    epochs_years = (epochs.tt.value - J2000_jd) / 365.25
+    epochs_years = [time_representation.seconds_since_epoch_to_julian_years_since_epoch(epoch_tdb) for epoch_tdb in epochJ2000secondsTDB]
```

- in `_add_table`:
```
-        obs = obs.assign(
-            epochJ2000secondsTDB=lambda x: (
-                Time(x.epoch, format="jd", scale="utc").tdb.value - 2451545.0
-            )
-            * 86400
-        ).assign(epochUTC=lambda x: Time(x.epoch, format="jd").to_datetime())
+        obs = self._add_time_columns(obs)
         if in_degrees:
-            obs = obs.assign(RA=lambda x: np.radians(x.RA)).assign(
+            obs = obs.assign(RA=lambda x: (np.radians(x.RA)  + np.pi ) % (2 * np.pi) - np.pi).assign(
                 DEC=lambda x: np.radians(x.DEC)
             )
```
- `_add_time_columns` method was created, in order to allow for these (-) lines to be replaced with the (+) lines
```
-                obs = (
-                    obs.assign(
-                        epochJ2000secondsTDB=lambda x: (
-                            Time(x.epoch, format="jd", scale="utc").tdb.value
-                            - 2451545.0
-                        )
-                        * 86400
-                    )
-                    .assign(RA=lambda x: np.radians(x.RA))
-                    .assign(DEC=lambda x: np.radians(x.DEC))
-                    .assign(epochUTC=lambda x: Time(x.epoch, format="jd").to_datetime())
+                obs = self._add_time_columns(obs)
+                obs = obs.assign(
+                    RA=lambda x: (np.radians(x.RA) + np.pi) % (2 * np.pi) - np.pi,
+                    DEC=lambda x: np.radians(x.DEC)
                 )
```
## Computations done through Tudat's DateTime:

- Python datetime is never used internally. It is only used for user-inputs in the `filter()` function, but these get soon converted to Tudat DateTimes (in TDB) via:
                    
```
                    user_epoch = datetime.datetime(date) # python datetime
                    user_epoch_iso_string = epoch_start.isoformat(sep=' ') #iso string
                    user_start_utc = DateTime.from_iso_string(epoch_start_iso_string).to_epoch() # convert to UTC float 
                    epoch_start_tdb = time_scale_converter.convert_time(
                        input_scale=time_representation.utc_scale,
                        output_scale=time_representation.tdb_scale,
                        input_value=epoch_start_utc) # convert to TDB float
```


- in `_add_time_columns()`:  `time_representation.default_time_scale_converter()` is used to convert mpc-retrieved UTC times into TDB floats and add the corresponding TDB epoch column;
- in `filter()`: dates used to be filtered based on UTC min and UTC max dates (given in `datetime` and user-defined). The bounds can still be defined as `datetime` objects by the user, but now we use the `time_representation.default_time_scale_converter()` to internally convert the min/max UTC bounds to min/max TDB bounds, and filtering on the TDB column instead of the UTC column (although the table still contains both columns for ease of the users);
- in `plot_observations_sky()`: we use the `time_representation.default_time_scale_converter()` to go from TDB to UTC so as to be able to plot nice UTC labels.

src/tudatpy/data/mpc/mpc.py
```diff
@@ -248,11 +248,11 @@ def get_biases_EFCC18(
 
 def get_weights_VFCC17(
         MPC_codes: pd.Series | list | np.ndarray | None = None,
-        epochUTC: pd.Series | list | np.ndarray | None = None, 
-        observation_type: pd.Series | list | np.ndarray | None = None, 
-        observatory: pd.Series | list | np.ndarray | None = None, 
-        star_catalog: pd.Series | list | np.ndarray | None = None, 
-        mpc_table: pd.DataFrame | None = None, 
+        epoch: pd.Series | list | np.ndarray | None = None,
```
DominicDirkx, line 251: Is the change from epochUTC to epoch only a matter of terminology, or is the meaning changed (if there is no specifier for the time scale, it should normally be TDB, not UTC)

src/tudatpy/data/mpc/mpc.py
```diff
@@ -394,9 +388,9 @@ def get_weights_VFCC17(
     # TABLE 5: Non-CCD residuals
     # Conditions for Table 5
     # 1890-1-1 and 1950-1-1
-    pre_1890 = table.epochJD <= 2411368.0
-    between_1890_1950 = (table.epochJD > 2411368.0) & (table.epochJD <= 2433282.0)
-    after_1950 = table.epochJD > 2433282.0
+    pre_1890 = table.epoch <= 2411368.0
```
DominicDirkx: The change here seems inconsistent (in naming at least) the epoch in Julian day is change to just 'epoch' which should be TDB (?)

src/tudatpy/data/mpc/mpc.py
luigigisolfi: the table has three time columns

`epoch = UTC JD, epochJD2000TDB = float TDB, epochUTC = float UTC`

the old version of get_weights_VFCC17 was using epochUTC there. Since i moved away from using epochUTC there, I changed the name to epoch to reflect JD in UTC, and be consistent with the table denominations. So UTC. Should we move it to epochJDUTC or something similar?

src/tudatpy/data/mpc/mpc.py
luigigisolfi: (Citing comment on line 251) which explains why this is not an issue


"> the table has three time columns
> epoch = UTC JD, epochJD2000TDB = float TDB, epochUTC = float UTC
> 
> the old version of get_weights_VFCC17 was using epochUTC there. Since i moved away from using epochUTC there, I changed the name to epoch to reflect JD in UTC, and be consistent with the table denominations. So UTC. Should we move it to epochJDUTC or something similar?"

src/tudatpy/data/mpc/mpc.py
DominicDirkx: The term 'epoch' comes directly from teh MPC right? While in 'Tudat-speak' the term is a little confusing, I think it's good to keep the dataframe a direct representation of the MPC where we can.

For the other two:

epochJD2000TDB Is this the epoch in seconds since J2000 or Julian days since J2000? I'm assuming epochUTC is seconds since J2000

luigigisolfi: It comes directly from the astroquery wrapper query, yes. 
sorry, there was a typo: it's epochJ2000TDB (without a D), which is in seconds (float) since J2000.

DominicDirkx: If it doesn't make the user interface (too) backwards-incompatible, I'd recommend making explicit that the other two are in seconds (J2000 will always be the reference epoch and need not be named in the variable I think). So, we would have:

* epoch (UTC, Julian days)
* epoch_seconds_UTC (UTC, seconds)
* epoch_seconds_TDB (TDB, seconds)

That way, the original 'non-tudatness' of the variable name 'epoch' is translated into the other variable names in a kind of logical manner (specifying that the unit is seconds, since the 'default' now becomes days)

luigigisolfi:
Overview of the latest changes:

1) as suggested by @DominicDirkx : now the table supports "`epoch` [JD, to stay consistent with MPC columns denomination]", "`epoch_seconds_TDB`", "`epoch_seconds_UTC`"
2) some of these entries were also being used within `horizons.py` and `horizons_wrapper.py`, as well as in `parsers.py `and `test_data_mpc.py` and `test_data_80col_parser.py`, hence the corresponding commits. 
3) I pulled from develop 

> NOTES (Edit)
>The unit tests run fine.
> Tudatpy-examples run fine.

 
```

DominicDirkx:
Thanks!

PR #616, opened 2026-01-20
Fix bug in createCompressedDopplerCollection with low compression ratios

### LINKED ISSUE #615: [BUG]: `create_compressed_doppler_collection()` produces empty observation collection for compression_ratio ≤ observation spacing
### Describe the issue:

When using `create_compressed_doppler_collection()` to compress DSN Doppler observations from IFMS files, certain combinations of `compression_ratio` and `min_observations` result in an observation collection with an **empty** `link_definitions_per_observable` dictionary. This then leads to a **KeyError** when attempting to access the Doppler link ends for residual computation.

In my case, the IFMS file contains Doppler data sampled every **10 seconds** over **6 hours**, with **no gaps**. However, when using a `compression_ratio` of **10 or lower**, the number of compressed observations becomes **0**, and any later access to `link_definitions_per_observable[...]` triggers a KeyError.

Comparison script:
https://github.com/Markus-Reichel/rosetta/blob/424d952c94142a5c7ba092eb985e6d206dfdf062/doppler_data_analysis/rosetta_residuals/doppler_compression_comparison.py

The script loops through different `compression_ratio` and `min_observations` configurations. For all `compression_ratio <= 10`, a KeyError occurs due to the compressed observation collection having (see https://github.com/Markus-Reichel/rosetta/blob/424d952c94142a5c7ba092eb985e6d206dfdf062/doppler_data_analysis/rosetta_residuals/doppler_compression_comparison.py#L361C9-L366C1):
- `len(compressed_observations) == 0`, and
- `compressed_observations.concatenated_observations.size == 0`, and
- `len(compressed_observations.link_definitions_per_observable) == 0`

The script requires additional input files to run, which are not in the github repo! 

### OS and Python version:

OS: macOS 26.1
Python: 3.11.13

### Expected behavior:

The function should produce compressed observations regardless of the compression ratio, as long as there are sufficient observations.

### Current behavior:

- compression_ratio >= 11: Works correctly
 - compression_ratio <= 10: Returns empty observation collection, causing KeyError

### Additional context:

I traced the “0 compressed observations” behavior to two likely issues in `processOdfFile.h` that can combine to produce an empty compressed observation collection (and therefore an empty `link_definitions_per_observable`, causing a KeyError).

### Suspected bug #1: `compression_ratio` is treated as a *time gap* (seconds) for arc splitting
In `processOdfFile.h` (around lines **1430–1434**), `compression_ratio` appears to be passed into `time_interval_splitter` as a **maximum time interval** (seconds) used to split arcs.

My understanding is that `compression_ratio` is intended to be *count-based* (e.g., “compress/average N observations together”), not a time-gap threshold.

Example: 10-second sampled data, `compression_ratio = 5`
- The code interprets this as: “split arcs at gaps > 5 seconds”
- Consecutive observations are ~10 seconds apart
- Therefore every consecutive pair triggers a split
- Result: **each observation becomes its own 1-observation arc**

This matches the arc splitting logic (e.g., `singleObservationSet.h` around line **1287**) where the condition becomes `10.0 > 5.0` → `true` → split at every step.

### Suspected bug #2: strict inequality in the compression check (`>` vs `>=`)
In `processOdfFile.h` (around line **1360**), the compression block is guarded by:

`if (originalObservations.size() - i > compressionRatio)`

This uses `>` instead of `>=`, meaning it requires **MORE than** `compressionRatio` observations remaining to produce any output. If the intention is to compress *exactly* `compressionRatio` observations (or at least that many), then `>=` would be expected.

Combined with Bug #1 (1-observation arcs), this condition will always fail, and compression produces **no output**.

### Why `min_observations = 1` still fails
Even if `min_observations = 1`, compression can still yield **0 compressed observations**. For:
- `compression_ratio = 5`
- `min_observations = 1`
- 10-second sampling (continuous 6-hour arc)

**Step 1: Arc splitting** (`processOdfFile.h:1430–1434`, approx.)
- `compression_ratio` is effectively used as `maxTimeInterval = 5.0 s`
- consecutive observations are ~10 s apart
- check (e.g. `singleObservationSet.h:1287`): `10.0 > 5.0` → `true`
- result: **new arc created at every observation** → each arc has 1 observation

**Step 2: Minimum filter**
- each arc has 1 observation
- check (e.g. `singleObservationSet.h:1336`): `1 >= 1` → `true`
- result: **all 1-observation arcs pass the filter**

**Step 3: Compression ** (`processOdfFile.h:1358–1360`)
- loop effectively becomes: `for (i = 0; i < 1; i += 5)`
- guard check becomes: `1 - 0 > 5` → `false`
- compression block is skipped
- result: **0 compressed observations per arc**

**Step 4: Empty set removal** (`processOdfFile.h:1458–1459`)
- all observation sets have 0 compressed observations
- all sets removed
- result: empty observation collection → later access to `link_definitions_per_observable` triggers **KeyError**

### Suggestion
- If `compression_ratio` is meant to be count-based, it likely should not be used as a time-gap parameter for `time_interval_splitter` (or that logic should use a separate, explicitly time-based parameter).
- Changing the condition `>` to `>=` would allow exact-sized compression windows to produce output.

### LINKED ISSUE #617: [BUG]: `create_compressed_doppler_collection`  discards most observations due to overly strict time tolerance
### Describe the issue:

 The `create_compressed_doppler_collection` function loses a significant portion of observations during compression. The data loss becomes larger with higher compression ratios.

 1. Load Doppler observations (e.g., from IFMS files) with ~2000 data points
 2. Call `create_compressed_doppler_collection` with compression ratio 6
 3. Observe that far fewer compressed points are produced than expected

### OS and Python version:

OS: macOS 26.1
Python: 3.11.13

### Expected behavior:

 - 2019 data points with compression ratio 6 → ~336 compressed points (2019/6)
 - 2019 data points with compression ratio 60 → ~33 compressed points (2019/60)

### Current behavior:

 - 2019 data points with compression ratio 6 → 300 compressed points (11% loss)
 - 2019 data points with compression ratio 60 → 3 compressed points (91% loss)

### Additional context:

The time gap tolerance check in `compressDopplerData()` (processOdfFile.h:1368-1369) is too strict:
```

 if( ( originalObservationTimesUtc.at( i + j ) - originalObservationTimesUtc.at( i + j - 1 ) - currentCompressionTime ) <
     10.0 * std::numeric_limits< double >::epsilon( ) * static_cast< double >( originalObservationTimesUtc.at( i + j ) ) )

```
 Issues:
 1. Tolerance is ~1 microsecond and might be too strict
 2. Missing absolute value. only catches gaps larger than expected, not smaller
 3. Data loss. Any observation with timing variations > 1μs is discarded via `skipObservation = true`

Markus-Reichel:
## Summary
-  Fixed  #615 
 - Fixed off-by-one error in boundary check that skipped last batch of observations
 - Added `maxArcGap` parameter to decouple arc splitting from compression ratio
 - Updated Python bindings with new `max_arc_gap` parameter (default: 300.0s)

 ## Problem
 When using low compression ratios (e.g., 5) with data sampled at longer intervals (e.g., 10s), the function produced no compressed
 observations because:
 1. The arc splitter used `compressionRatio` as the gap threshold (5s < 10s interval → split every observation into separate arc)
 2. Boundary check `>` instead of `>=` skipped the last batch

 ## Solution
 - Change boundary check from `>` to `>=` to include last batch
 - Add explicit `maxArcGap` parameter (default 300s) for arc splitting
 - This decouples arc splitting behavior from compression ratio

## Additional Fixes:
- Fixed #617

### DIFF
--- a/include/tudat/simulation/estimation_setup/processOdfFile.h
+++ b/include/tudat/simulation/estimation_setup/processOdfFile.h
@@ -1357,7 +1357,7 @@ std::shared_ptr< observation_models::SingleObservationSet< ObservationScalarType
 
     for( unsigned int i = 0; i < originalObservations.size( ); i += compressionRatio )
     {
-        if( originalObservations.size( ) - i > compressionRatio )
+        if( originalObservations.size( ) - i >= compressionRatio )
         {
             Eigen::Matrix< ObservationScalarType, Eigen::Dynamic, 1 > newObservable = originalObservations.at( i );
             TimeType newTime = originalObservationTimesUtc.at( i );
@@ -1424,13 +1424,13 @@ template< typename ObservationScalarType = double, typename TimeType = double >
 std::shared_ptr< observation_models::ObservationCollection< ObservationScalarType, TimeType > > createCompressedDopplerCollection(
         const std::shared_ptr< observation_models::ObservationCollection< ObservationScalarType, TimeType > > originalDopplerData,
         const unsigned int compressionRatio,
-        const unsigned int minNumberObservations = 10 )
+        const unsigned int minNumberObservations = 10,
+        const double maxArcGap = 300.0 )
 {
     // Split Doppler observation sets into arcs
-    double compressionRatioFloat = mathematical_constants::getFloatingInteger< double >( compressionRatio );
     std::shared_ptr< observation_models::ObservationCollection< ObservationScalarType, TimeType > > compressedData =
             splitObservationSets( originalDopplerData,
-                                  observationSetSplitter( time_interval_splitter, compressionRatioFloat, minNumberObservations ),
+                                  observationSetSplitter( time_interval_splitter, maxArcGap, minNumberObservations ),
                                   observationParser( dsn_n_way_averaged_doppler ) );
 
     std::map< LinkEnds, std::vector< std::shared_ptr< observation_models::SingleObservationSet< ObservationScalarType, TimeType > > > >
--- a/include/tudat/simulation/estimation_setup/processOdfFile.h
+++ b/include/tudat/simulation/estimation_setup/processOdfFile.h
@@ -1365,8 +1365,8 @@ std::shared_ptr< observation_models::SingleObservationSet< ObservationScalarType
             bool skipObservation = false;
             for( unsigned int j = 1; ( j < compressionRatio && !skipObservation ); j++ )
             {
-                if( ( originalObservationTimesUtc.at( i + j ) - originalObservationTimesUtc.at( i + j - 1 ) - currentCompressionTime ) <
-                    10.0 * std::numeric_limits< double >::epsilon( ) * static_cast< double >( originalObservationTimesUtc.at( i + j ) ) )
+                if( std::fabs( static_cast< double >( originalObservationTimesUtc.at( i + j ) - originalObservationTimesUtc.at( i + j - 1 ) ) - currentCompressionTime ) <
+                    0.01 )
                 {
                     newObservable += originalObservations.at( i + j );
                     newTime += originalObservationTimesUtc.at( i + j );
--- a/src/tudatpy/estimation/observations_setup/observations_wrapper/expose_observations_wrapper.cpp
+++ b/src/tudatpy/estimation/observations_setup/observations_wrapper/expose_observations_wrapper.cpp
@@ -381,6 +381,7 @@ void expose_observations_wrapper( py::module &m )
            py::arg( "original_observation_collection" ),
            py::arg( "compression_ratio" ),
            py::arg( "minimum_number_of_observations" ) = 10,
+           py::arg( "max_arc_gap" ) = 300.0,
            R"doc(
         Create a compressed Doppler observation collection.
 
@@ -394,6 +395,8 @@ void expose_observations_wrapper( py::module &m )
             The number of observations to average into a single compressed observation.
         minimum_number_of_observations : int, optional
             The minimum number of observations required in an arc to be considered for compression, by default 10.
+        max_arc_gap : float, optional
+            Maximum time gap (in seconds) between consecutive observations before splitting into a new arc, by default 300.0.
 
         Returns
         -------


Markus-Reichel:
**Additional fix:** Fixed #617 

  While testing the initial fix, I discovered a second issue causing data loss during compression.

  **Problem:** The time gap tolerance check used an epsilon-based threshold (~1 microsecond), which is too strict. Any timing jitter > 1μs caused entire observations to be discarded.

  **Impact:**
  - Compression ratio 6: ~11% data loss (300 vs expected 336 points)
  - Compression ratio 60: ~91% data loss (3 vs expected 33 points)

  **Fix:** Replaced the epsilon-based tolerance with a practical 10ms tolerance using `std::fabs()`, consistent with the approach used in processTrackingTxtFile.h:475-476.

  **// Before**
  `if( ( time_diff - currentCompressionTime ) < 10.0 * epsilon * time )`

  **// After (practical tolerance)**
`  if( std::fabs( static_cast<double>(time_diff) - currentCompressionTime ) < 0.01 )`

Markus-Reichel:
Fixed: #618 

Rename misleading `minNumberObservations` parameter to `minObservationsPerArc` in Doppler compression and observation set splitting functions.

  The parameter name `minNumberObservations` was confusing because it suggested controlling the minimum number of observations used per compressed point. In reality, it controls the minimum number of observations required per arc after time-gap splitting. Arcs with fewer observations are discarded.

  This rename clarifies the parameter's actual purpose:
  - `compressionRatio`: number of observations averaged into each compressed point
  - `minObservationsPerArc`: minimum arc size to keep after splitting (filters out short arcs)

  **Changes**

  C++ API:
  - `ObservationSetSplitterBase`: renamed member `minNumberObservations_` → `minObservationsPerArc_`
  - `ObservationSetSplitterBase`: renamed getter `getMinNumberObservations()` → `getMinObservationsPerArc()`
  - `observationSetSplitter()` factory functions: renamed parameter
  - `createCompressedDopplerCollection()`: renamed parameter

  Python API:
  - `create_compressed_doppler_collection()`: renamed `minimum_number_of_observations` → `min_observations_per_arc`
  - `observation_set_splitter()`: renamed `min_number_observations` → `min_observations_per_arc`
  - Updated docstrings to clarify parameter behavior

include/tudat/simulation/estimation_setup/observationsProcessing.h
```diff
@@ -216,8 +216,8 @@ enum ObservationSetSplitterType { time_tags_splitter, time_interval_splitter, ti
 
 struct ObservationSetSplitterBase {
 public:
-    ObservationSetSplitterBase( const ObservationSetSplitterType splitterType, const int minNumberObservations = 0 ):
-        splitterType_( splitterType ), minNumberObservations_( minNumberObservations )
+    ObservationSetSplitterBase( const ObservationSetSplitterType splitterType, const int minObservationsPerArc = 0 ):
+        splitterType_( splitterType ), minObservationsPerArc_( minObservationsPerArc )
```
DominicDirkx: Let's leave the name here the same as what it was. This is in the base class, for which there's no need to change the name, since the term 'arc' has no set meaning here

DominicDirkx:
@jo11he  WHen you have the time, could you check if your Cassini pre-fit residuals are unaffected by these changes?

include/tudat/simulation/estimation_setup/observationsProcessing.h
Markus-Reichel: In general or only in the two lines you highlighted?

DominicDirkx: In any place that is not specific to the application you're using these splitters for. Changing the name from minNumberObservations to minObservationsPerArc makes it seem application specific (an application in which the term 'arc' has a meaning)

DominicDirkx:
@Markus-Reichel I tested your code with our existing GRAIL example, and it worked without a problem. If you can (i) remove the renaming (ii) make the default input equivalent (as much as possible) with the original ones, we can merge this!

@jo11he @valeriof7 @alfonsoSR  This functionality should not impact your work for Cassini, MRO or Mars Express, but please verify to make sure

Markus-Reichel:
@DominicDirkx I moved the branch pointer back to the second commit, so the renaming is excluded. 
Regarding the default values: I introduced the `maxArcGap` parameter, which replaces the `compressionRatio` being passed to `time_interval_splitter()`, which was the bug that I fixed. And `compressionRatio` did not have a default value. 
So it should be ready to merge now.

jo11he:
I checked, I do not see any impact on the Cassini observation modeling. 

If I may pitch in on the _minNumberObservations_ discussion: I agree with Markus that we can be more explicit on the naming of this parameter, but I think the _per_arc_ choice is misleading. In the end, I think the unit that is discarded if this criterion is not met is the given observation _set_, which does not necessarily correspond to any physical "arc" (which is also not a clearly defined term). I see e.g. in the case of the Cassini Doppler data, that my smallest "physical" group is a tracking _pass_ (which is the name it is referred to by the PDS documentation, maybe what @Markus-Reichel calls _arc_), which is a more or less continuously tracked interval of tracking by the same linkdef. In some cases this corresponds to a single observation set, but in most cases it is composed of multiple such sets (presumably because of small variations in the ancillary settings).

Markus-Reichel:
@jo11he thats exactly what I was referring to. Thank you for clarifying.

PR #619, opened 2026-01-22
Add MPC stations from Project Pluto, assuming no motion.

luigigisolfi:
This PR adds Tudat support for optical observatories [listed on Project Pluto](https://www.projectpluto.com/mpc_stat.txt).
Project Pluto provides geodetic positions and parallax measurements (`rho_cos_phi`, `rho_sin_phi`). These were converted to Cartesian Coordinates via a standalone code that implements the following: 

```
            x = EARTH_RADIUS * rho_cos_phi * math.cos(lon_rad)
            y = EARTH_RADIUS * rho_cos_phi * math.sin(lon_rad)
            z = EARTH_RADIUS_* rho_sin_phi
```

Since no velocities are directly available for those observatories, I created a corresponding velocities files containing all zeros in its columns. 

Both positions and velocities are then saved to the `mpc.sit` and `mpc.vel` files, which needs to be added to the `resources` folder, where also `glo.sit` is found (same applies for `mpc.vel` and `mpc_codes.dat`). 

Converting back the results from the `mpc.sit` file (`cartesian -> geodetic`) returns Project Pluto's latitudes and longitudes up to `~1e-5 deg` for most stations, while altitude differences wrt to Project Pluto are typically `~10 meters`, although some might be bigger because of ambiguous/erroneous ellipsoidal/geodetic definition (see this [discussion] (https://www.projectpluto.com/mpc_stat.htm)). These differences are acceptable as they should not matter much in routine SSA analysis.

### DIFF
--- a/include/tudat/simulation/environment_setup/defaultBodies.h
+++ b/include/tudat/simulation/environment_setup/defaultBodies.h
@@ -295,6 +295,8 @@ std::vector< std::shared_ptr< GroundStationSettings > > getDsnStationSettings( )
 
 std::vector< std::shared_ptr< GroundStationSettings > > getEvnStationSettings( );
 
+std::vector< std::shared_ptr< GroundStationSettings > > getMPCStationSettings( );
+
 std::vector< std::shared_ptr< GroundStationSettings > > getRadioTelescopeStationSettings( );
 
 }  // namespace simulation_setup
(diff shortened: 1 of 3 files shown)


luigigisolfi:
@DominicDirkx Leaving this as a draft for now. If we like the changes and methodology, we can merge it

DominicDirkx:
Sounds good! Do you already have the extra files for the resources somewhere?

luigigisolfi:
@DominicDirkx I created a PR on tudat-resources: https://github.com/tudat-team/tudat-resources/pull/8

Issue #586, opened 2025-12-18
[F&E]: MCD integration restructuring and PRP model


nikemais:
### Description:

PR #561 introduced the structure to allow the integration of MCD for an atmosphere model. However, extracting values from MCD can be an expensive process (a vector of keys is passed but some variables are always computed), especially if multiple models require the same variables.

### Suggestions for implementation:

I suggest the following changes to be as efficient as possible:
- MCD exposure lies within its own class (e.g., MarsClimateDatabaseExposure) and has an update function
- During setup, all the models requiring variables from MCD create a single instance of MCDExposure 
- MCDExposure::update is then called by EnvironmentUpdater

On top of that, a rework of the CI workflow is implemented:
- The dependencies (gfortran and NETCDF) are installed using conda
- A minimal set of MCD data for the tests and the source code is pulled from a private repo
- Build with MCD flag on and test

Lastly, a new planetary radiation pressure model is implemented, using the above mentioned MCD structure.

### Additional context:

For now the private repo is external, if everything works as intended we could move it to tudat-team.

PR #622, opened 2026-01-26
MCD integration restucturing

nikemais:
Issue #586  introduced the restructuring the MCD integration (see #382 and PR #561).

Outline of changes:
- new generalized ClimateModel class, MCD interface becomes MarsClimateDatabaseClimateModel. The ClimateModel is a property of a planet body and is updated via EnvironmentUpdater.
- restructured McdAtsmosphereModel accesses the required variables from MarsClimateDatabaseClimateModel.
- unit testing: limited to K1 and propagation using MCD due to test data size limitation for CI.
- CI rework to support MCD (repo token added to secrets) **[currently under testing using this PR as draft]**.

Regarding the CI workflow, a private repo contains  the minimum necessary data to run the tests (2 climate averages files and the MOLA32 dataset) that amounts to roughly 500MB of data. Using the release is the only option due to the large size. The CI uses caching so that subsequent runs shouldn't download the dataset again, speeding up the process.


nikemais:
After a quick check, I have found out forks by default cannot see the secrets of the main repo. This means we either allow forks to see the secrets (possible privacy breach for MCD data, which must be avoided) or keep the settings as they are and wait for a maintainer to approve the full CI, which exposes it to the secrets. One last option would be to fully test only from branches within the tudatpy repo or after the PR is merged.

DominicDirkx:
> wait for a maintainer to approve the full CI, which exposes it to the secrets

For a PR made from any repo except https://github.com/tudat-team/tudatpy, this is already the case, so that sounds fine to me.

DominicDirkx:
Question: What does it mean exactly for 'forks to read secrets from main repo'. Anyone can fork our repo, so for them to be able to see our secrets does not seem like a good idea (?)

nikemais:
I used wrong the wrong term, my bad. Forks cannot directly "read" the secrets or print them in the actions logs, but they can use them in the CI (https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target). What could happen is someone actively submitting a PR with a custom workflow that can echo them or run a script to save them somewhere. Since we must approve the PRs first, a quick check of the workflow file (if modified) should prevent secrets exposure. 

If this doesn't sound solid enough (which is also my concern), I'll come up with a different testing strategy (maybe dual CI, a public one and a private one, triggered only by maintainers).

DominicDirkx:
Thanks,  that's clear! I see the built in safety, but we'd always have to add a human check to our workflow, which will get forgotten. Maybe easier (to be discussed) adding a check that a modification to the workflow is only permitted in a PR from our own rep

nikemais:
I did some reading on the matter. Secrets could be read by the tests with a simple line (`std::getenv("SECRET_TOKEN")`), given they have been exposed by the CI first and thus saved in memory. 

This can be prevented by exposing the secret in a different job within the same workflow, basically creating a first job "download-data" and then the usual "build-and-test". By doing so, no code can read the secrets that have been used.

On top of that, we can create a CODEOWNER file and add the workflow.yml to it, which upon modification, would require intervention of the listed owners (similar to what you suggested).

From my quick research, 100% leak proof system is impossible when dealing with private repos (even within the same org) and their tokens. However, this setup (separate jobs + workflow owners) reduces the risk significantly.

DominicDirkx:
Sounds good! If we add a check that PRs with changes to workflows always have to be approved, this would be even more safe, but this could be moved to a different issue

DominicDirkx:
@nikemais How have you been doing? Do you think you'll want to finalize this issue still? If not that's all fine, but then I'll ask someone else (@larshinueber?) to get it to the finish line :)

nikemais:
There are many conflicts that arise from the new structure vs the present one. The core functionality remains exactly the same, while allowing for more flexibility in the future. 

The implementation for the climate model is basically complete, I was testing the automatic MCD testing using github actions, which first require resolving these issues. I think it would be easier for me to solve these instead of trying to explain the idea I was going after, but feel free to step in and ask for clarifications.

The expansion to planetary radiation pressure is not present in this PR.

DominicDirkx:
Continuing this PR in #823. Since this PR is from a different repo, there was an issue using the secret variable to download MCD.

PR #625, opened 2026-01-28
Harden mission data downloader + quieter, more reliable Rosetta flows

Markus-Reichel:
- **Downloader core hardening**
  - Introduced a shared `_REQUEST_TIMEOUT`, an atomic `_download_file` helper, and `_get_latest_versioned_files` so every kernel fetch uses bounded HTTP requests and produces clean skip summaries (instead of per-file chatter).
  - Touchpoints: `src/tudatpy/data/mission_data_downloader/mission_data_downloader.py` (around L26, L515, L537, L577)

- **Single-time + time-interval download flow rework**
  - Reworked both flows to use set-based `check_existing_files`, enforce highest-version selection with proper error handling, add a `quiet` mode, and batch “already present” reporting via `_format_existing_files`.
  - Callers still receive paths for files that were already on disk.
  - Touchpoints: `mission_data_downloader.py` (around L880, L932, L1092, L1119, L1126)

- **Meta-kernel behavior improvements**
  - Meta-kernel downloads now rewrite `PATH_VALUES` so freshly pulled `.tm` files work out of the box.
  - Added `skip_kernel_downloads` to prevent Rosetta workflows from re-downloading CK/SPK/FK/SCLK kernels after a meta-kernel pass.
  - Touchpoints: `mission_data_downloader.py` (around L1833, L1847, L2180, L4416, L4561)

- **Rosetta radio-science pipeline reliability**
  - Cached archive URLs, trimmed unsupported closed-loop products, summarized DP2 downloads, deduplicated RSI volume IDs, guarded mission-phase ↔ target mappings, and wrapped mapping fetches / HEAD checks in retryable, timeout-bound calls for much more predictable behavior on flaky networks.
  - Touchpoints: `mission_data_downloader.py` (around L4416, L4520, L4695, L4788, L4823, L4952)

### DIFF
--- a/src/tudatpy/data/mission_data_downloader/mission_data_downloader.py
+++ b/src/tudatpy/data/mission_data_downloader/mission_data_downloader.py
@@ -836,21 +836,22 @@ class LoadPDS:
                 data_type.lower()
             ]
 
-        except:
+        except KeyError:
             raise ValueError("Pattern not found among supported patterns.")
 
         existing_files = self.check_existing_files(
             data_type, local_subfolder, start_date, end_date
         )
+        if not existing_files:
+            existing_files = set()
 
-        if existing_files and self.flag_check_existing_files == True:
+        if existing_files:
             print(
                 f"--------------------------------------- EXISTING FILES CHECK ----------------------------------------------"
             )
             print(
                 f"The following files already exist in the folder:\n\n {existing_files}\n\n and will not be downloaded."
             )
-            self.relevant_files.extend(existing_files)
 
         all_dates = [
             start_date + timedelta(days=x)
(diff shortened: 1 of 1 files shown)


Markus-Reichel:
@luigigisolfi could you please have a look at this and check if your downloader still works as intended? It should work just fine but let me know if not!

luigigisolfi:
@Markus-Reichel I tried it on my codes and it works fine. I spotted a bug in Cassini's downloader, so I fixed it in this PR to your branch: https://github.com/Markus-Reichel/tudatpy/pull/1

Markus-Reichel:
@DominicDirkx It seems like everything is in order and if @luigigisolfi doesnt have any additions, this PR would be ready to merge

PR #628, opened 2026-01-28
Replace lagrange interpolator with barycentric lagrange interpolator

### LINKED ISSUE #612: [BUG]: Doppler residuals (IFMS files) depend on ephemeris setup (time-limited bodies + interpolated SPICE)
### Describe the issue:

When running a Doppler residual analysis with Doppler measurements from **IFMS** files, the resulting residuals are influenced by the choice of ephemerides used for the bodies (e.g. `environment_setup.direct_spice()`, `environment_setup.interpolated_spice()`, or `environment_setup.get_default_body_settings_time_limited()`).

In my case (Rosetta spacecraft), the Doppler residuals show an unexpected **"layered" structure** when using **time-limited body settings** and **interpolated SPICE**. The expected/correct residual behavior is achieved when using **default bodies (without time-limit)** and **direct SPICE**.

Image
Image

Comparison script:
https://github.com/Markus-Reichel/rosetta/blob/8f0566875165cf4de76cc43f89b5edb84fca6fde/doppler_data_analysis/rosetta_residuals/ephemerides_setup_comparison.py

### Case 1 — unexpected residuals
- Default bodies initialized with `environment_setup.get_default_body_settings_time_limited()`  
  https://github.com/Markus-Reichel/rosetta/blob/8f0566875165cf4de76cc43f89b5edb84fca6fde/doppler_data_analysis/rosetta_residuals/ephemerides_setup_comparison.py#L189C5-L196C28
- Ephemerides set to `environment_setup.interpolated_spice()` for:
  - Rosetta:  
    https://github.com/Markus-Reichel/rosetta/blob/8f0566875165cf4de76cc43f89b5edb84fca6fde/doppler_data_analysis/rosetta_residuals/ephemerides_setup_comparison.py#L262C5-L269C46
  - 67P:  
    https://github.com/Markus-Reichel/rosetta/blob/8f0566875165cf4de76cc43f89b5edb84fca6fde/doppler_data_analysis/rosetta_residuals/ephemerides_setup_comparison.py#L226C5-L233C40

**Result:** residuals show the “layered” structure (unexpected).

### Case 2 — expected/correct residuals
- Default bodies initialized with `environment_setup.get_default_body_settings()`  
  https://github.com/Markus-Reichel/rosetta/blob/8f0566875165cf4de76cc43f89b5edb84fca6fde/doppler_data_analysis/rosetta_residuals/ephemerides_setup_comparison.py#L199C8-L203C10
- Ephemerides set to `environment_setup.direct_spice()` for:
  - Rosetta:  
    https://github.com/Markus-Reichel/rosetta/blob/8f0566875165cf4de76cc43f89b5edb84fca6fde/doppler_data_analysis/rosetta_residuals/ephemerides_setup_comparison.py#L272C9-L276C10
  - 67P:  
    https://github.com/Markus-Reichel/rosetta/blob/8f0566875165cf4de76cc43f89b5edb84fca6fde/doppler_data_analysis/rosetta_residuals/ephemerides_setup_comparison.py#L251C5-L255C6

**Result:** residuals behave as expected.

### OS and Python version:

OS: macOS 26.1
Python: 3.11.13

### Expected behavior:

Doppler residuals should be consistent (within expected numerical differences) across ephemerides setups, and should not develop qualitative artifacts (“layering”) solely due to using time-limited bodies and interpolated SPICE.

### Current behavior:

Using `get_default_body_settings_time_limited()` + `interpolated_spice()` produces qualitatively different residuals (layered structure), while `get_default_body_settings()` + `direct_spice()` produces the expected residuals.

### Additional context:

_No response_

Markus-Reichel:
**The Issue**

-  On Mac Silicon (ARM), `long double` is only 64-bit (same as `double`), whereas on x86 Linux it's 80-bit. This caused precision loss in the Lagrange interpolator when computing Doppler  residuals with interpolated ephemerides.

The original formula was numerically unstable:
  ```
// Compute product of ALL differences first
  repeatedNumerator = Π(x - x_k)  // accumulates rounding errors

  // Then divide by each difference
  L(x) = Σ[ y_j × repeatedNumerator / ((x - x_j) × denominator_j) ]

```
 **The problem**: 

- multiplying many small numbers together (repeatedNumerator *= diff) accumulates significant rounding error, especially with reduced long double precision on Mac Silicon.

**The Fix**
Replaced with the barycentric Lagrange formula, which is mathematically equivalent but numerically stable:

`  L(x) = Σ[w_j × y_j / (x - x_j)] / Σ[w_j / (x - x_j)]`

**Improvements:**

- No accumulated product: each term is computed independently with just one division
- O(n) complexity: same as before, but with better numerical prperties
- Robust exact-match handling: checks all stencil points (original only checked 3)

 **Files Modified**

- `include/tudat/math/interpolators/lagrangeInterpolator.h`
- Replaced interpolation loop (lines 272-308) with barycentric formula
- Added required headers (`<algorithm>, <cmath>, <limits>`)
- Removed unused `independentVariableDifferenceCache` member variable

**Fixed:**

- Fixed #612

### DIFF
--- a/include/tudat/math/interpolators/lagrangeInterpolator.h
+++ b/include/tudat/math/interpolators/lagrangeInterpolator.h
@@ -280,17 +280,30 @@ public:
                 int j = i + lowerEntry - offsetEntries_;
                 ScalarType diff = static_cast< ScalarType >( targetIndependentVariableValue - independentValues_[ j ] );
 
-                // Check for exact match to avoid division by zero.
-                // Use a tolerance that accounts for floating-point representation.
-                ScalarType absIndepValue = std::abs( static_cast< ScalarType >( independentValues_[ j ] ) );
-                ScalarType tolerance = std::max(
-                    absIndepValue * std::numeric_limits< ScalarType >::epsilon( ) *
-                        mathematical_constants::getFloatingInteger< ScalarType >( 1000 ),
-                    std::numeric_limits< ScalarType >::epsilon( ) *
-                        mathematical_constants::getFloatingInteger< ScalarType >( 1000 ) );
+                // Check for exact match to avoid division by zero in barycentric formula.
+                // Use 10 ULP tolerance to catch:
+                // - Bitwise identical values
+                // - Values differing due to float/double/long double conversions
+                // - Values with small differences due to cross-platform rounding (Mac Silicon vs x86)
+                ScalarType gridValue = static_cast< ScalarType >( independentValues_[ j ] );
+                ScalarType targetValue = static_cast< ScalarType >( targetIndependentVariableValue );
 
-                if( std::abs( diff ) < tolerance )
+                // Compute relative tolerance based on the larger magnitude (symmetric comparison)
+                ScalarType largest = std::max( std::abs( gridValue ), std::abs( targetValue ) );
+                ScalarType relativeTolerance = largest * std::numeric_limits< ScalarType >::epsilon( ) *
+                    mathematical_constants::getFloatingInteger< ScalarType >( 5 );
+
+                // Provide minimum absolute tolerance for near-zero grid points
+                // This handles cases where both gridValue and targetValue are close to zero
+                ScalarType absoluteTolerance = std::numeric_limits< ScalarType >::epsilon( ) *
+                    mathematical_constants::getFloatingInteger< ScalarType >( 10 );
+
+                ScalarType tolerance = std::max( relativeTolerance, absoluteTolerance );
+
+                if( std::abs( diff ) <= tolerance )
                 {
+                    // Target coincides with grid point within tolerance: return exact tabulated value
+                    // This avoids division by zero and potential numerical issues in barycentric formula
                     return dependentValues_[ j ];
                 }
 
@@ -301,7 +314,6 @@ public:
                 numeratorSum = numeratorSum + dependentValues_[ j ] * term;
                 denominatorSum += term;
             }
-
             interpolatedValue = numeratorSum / denominatorSum;
         }
 
--- a/include/tudat/math/interpolators/lagrangeInterpolator.h
+++ b/include/tudat/math/interpolators/lagrangeInterpolator.h
@@ -281,7 +281,7 @@ public:
                 ScalarType diff = static_cast< ScalarType >( targetIndependentVariableValue - independentValues_[ j ] );
 
                 // Check for exact match to avoid division by zero in barycentric formula.
-                // Use 10 ULP tolerance to catch:
+                // Using 10 ULP tolerance to catch:
                 // - Bitwise identical values
                 // - Values differing due to float/double/long double conversions
                 // - Values with small differences due to cross-platform rounding (Mac Silicon vs x86)
@@ -291,7 +291,7 @@ public:
                 // Compute relative tolerance based on the larger magnitude (symmetric comparison)
                 ScalarType largest = std::max( std::abs( gridValue ), std::abs( targetValue ) );
                 ScalarType relativeTolerance = largest * std::numeric_limits< ScalarType >::epsilon( ) *
-                    mathematical_constants::getFloatingInteger< ScalarType >( 5 );
+                    mathematical_constants::getFloatingInteger< ScalarType >( 10 );
 
                 // Provide minimum absolute tolerance for near-zero grid points
                 // This handles cases where both gridValue and targetValue are close to zero
--- a/include/tudat/math/interpolators/lagrangeInterpolator.h
+++ b/include/tudat/math/interpolators/lagrangeInterpolator.h
@@ -15,7 +15,10 @@
 #ifndef TUDAT_LAGRANGEINTERPOLATOR_H
 #define TUDAT_LAGRANGEINTERPOLATOR_H
 
+#include <algorithm>
+#include <cmath>
 #include <iostream>
+#include <limits>
 
 #include "tudat/math/basic/mathematicalConstants.h"
 
@@ -137,9 +140,6 @@ public:
         // interpolation call.
         initializeDenominators( );
         initializeBoundaryInterpolators( selectedLookupScheme );
-
-        // Pre-allocate cache vector for computational efficiency.
-        independentVariableDifferenceCache.resize( 2 * offsetEntries_ + 2 );
     }
 
     //! Constructor from map of independent/dependent data.
@@ -215,8 +215,6 @@ public:
         // interpolation call.
         initializeDenominators( );
         initializeBoundaryInterpolators( selectedLookupScheme );
-
-        independentVariableDifferenceCache.resize( 2 * offsetEntries_ + 2 );
     }
 
     //! Destructor.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: include/tudat/math/interpolators/lagrangeInterpolator.h [hunk 1]
```diff
@@ -269,45 +267,42 @@ class LagrangeInterpolator : public OneDimensionalInterpolator< IndependentVaria
         }
         else
         {
-            // Initialize repeated numerator to 1
-            ScalarType repeatedNumerator = mathematical_constants::getFloatingInteger< ScalarType >( 1 );
-
-            // Check if requested independent variable is equal to data point
-            if( independentValues_[ lowerEntry ] == targetIndependentVariableValue )
```
DominicDirkx, line 276: We can keep these conditions right (checking if the target value is equal to one of the tabulated value). If they are true no interpolation is needed

include/tudat/math/interpolators/lagrangeInterpolator.h
```diff
@@ -269,45 +267,42 @@ class LagrangeInterpolator : public OneDimensionalInterpolator< IndependentVaria
...
-                int j = 0;
-                for( int i = 0; i <= 2 * offsetEntries_ + 1; i++ )
+                int j = i + lowerEntry - offsetEntries_;
+                ScalarType diff = static_cast< ScalarType >( targetIndependentVariableValue - independentValues_[ j ] );
+
+                // Check for exact match to avoid division by zero.
+                // Use a tolerance that accounts for floating-point representation.
+                ScalarType absIndepValue = std::abs( static_cast< ScalarType >( independentValues_[ j ] ) );
+                ScalarType tolerance = std::max(
+                    absIndepValue * std::numeric_limits< ScalarType >::epsilon( ) *
```
DominicDirkx: Why is this condition the way it is (with the value of 1000 in there? At which points will this condition become active? Are you certain the condition is not too restrictive, and could lead to numerical artifacts?

include/tudat/math/interpolators/lagrangeInterpolator.h
Markus-Reichel: I moved those checks inside the loop to check **all** tabulated values (Lines 283 - 295).

**Example:**
For  `numberOfStages = 6`, `offsetEntries_ = numberOfStages_ / 2 - 1 = 2`:
- `j` ranges from **lowerEntry - 2** to **lowerEntry + 3**
- Thats 6 points: `[lowerEntry-2, lowerEntry-1, lowerEntry, lowerEntry+1, lowerEntry+2, lowerEntry+3]`

**Before**:
The exact-match check was outside the loop and only checked 3 hardcoded points (`lowerEntry-1`, `lowerEntry` and `lowerEntry+1`) regardless of `numberOfStages_`.
If none of those 3 matched, it would continue to compute the interpolation normally. However, if the query exactly matches one of the **other** tabulated data points:
- `repeatedNumerator` would contain a zero factor (from `targetIndependentVariableValue - independentValues_[lowerEntry+2]=0`)
- The final division would compute `0 / 0` for that term

**New:**
The exact-match (with a tolerance) check is inside the loop at lines 292-295, so it checks all numerOfStages_ tabulated value points:
```
for( int i = 0; i < numberOfStages_; i++ ). // line 278
 {
      int j = i + lowerEntry - offsetEntries_;   // line 280
      // ...
      if( std::abs( diff ) < tolerance )    // line 292
      {
       return dependentValues_[ j ]; // line 294
       }
}

```

include/tudat/math/interpolators/lagrangeInterpolator.h
Markus-Reichel: The original code used exact equality (`==`):
`if(independentValues_[lowerEntry] == targetIndependentVariableValue)`

My thought was to change it to a tolerance check because  the issue with the new Macs Silicon chips is the  digits of precision. I think you mentioned that it only has 16 digits of precision instead of 19?  So exact floating-point equality can fail when the "same" time value is computed differently through type conversions. So a `Time` object converted to `long double` might not have the exact same pattern as the stored value.
Please correct me if I got this wrong!

So how the tolerance is calculated:
```

  ScalarType tolerance = std::max(
      absIndepValue * std::numeric_limits<ScalarType>::epsilon() * 1000,
      std::numeric_limits<ScalarType>::epsilon() * 1000 );
```
  - `std::numeric_limits<ScalarType>::epsilon()` ≈ 2e-16 for double
  - First term: relative tolerance, scales with the magnitude of the independent variable
  - Second term: absolute floor, ensures a minimum tolerance for values near zero
  - `std::max` takes the larger of the two

But you are correct. 1000 might be too large. I chose it as a safe "buffer" but when doing the math it is too large:
Lets say time = 5e8 seconds:
- tolerance ≈ 5e8 * 2e-16 * 1000 ≈ 1e-4 seconds ≈ 0.11 ms

So the question is do we even need a tolerance check or would an exact match be fine, and if we choose a tolerance, would `epsion()` be enough?

Markus-Reichel:
@DominicDirkx I looked a bit more into tolerances in this [guide](https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/) and I made some changes to the tolerances. But first off:


Previously exact equality (`==`) was used to detect when the target value coincided with a grid point to skip interpolation but also to avoid division by zero during the Lagrange interpolation. However, using the Lagrange interpolator lead to cross-platform inconsistencies between Mac Silicon and x86 architecture due to floating point representation differences.

I switeched to the Barycentric Lagrange Interpolator, which by itself already improved the numeric stability, because it only requires a single divsion instead of repeated products. I still opted to use tolerance-based comparison instead of exact equality becuase of:

1. Floating-point representation issues: Type conversions between `float`, `double` and `long double` will introduce small differences
2. Cross-platform binary representation: Mac Silicon (`long double` = 64-bit) vs x86 (`long double` = 80-bit) can produce slightly different rounding (as seen before)
3. Division by very small numbers: even with the new formula, dividing by very small values could cause precision loss.

How I implemented the comparison:

**First, Relative Tolerance:**
The tolerance is computed based on the larger magnitude of either the grid value or target value:
```

ScalarType largest = std::max(std::abs(gridValue), std::abs(targetValue));   // line 292
ScalarType relativeTolerance = largest * epsilon * 10;  line 293-294
```

This ensures "symmertric", consistent comparison regardless of which value is larger (It follows best practices from [Bruce Dawsons guide](https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/)). This ensures the tolerance adapts corretly to the scale of the numbers being compared.

Why 10 Units in the Last Place (ULPs)?
[Google Tests](https://google.github.io/googletest/reference/assertions.html) uses 4 ULPs for its floating-point assertions, which works well for unit testing. For our interpolation, I use 10 ULPs to have a buffer or safety margin for cross-platform consistency. Lets have an example:

For `time =  5e8` and `std::numeric_limits<ScalarType>::epsilon() ≈ 2e-16` for double:
- 4 ULP tolerance ≈ 5e8 * 2e16 * 4 ≈ 4e7 seconds ≈ 0.4 microseconds
- 10 ULP tolerance ≈ 5e8 * 2e16 * 10 ≈ 1e6 seconds ≈ 1 microsecond

10 ULP is still very tight and detects matches withing 1 microsecond which should be enough for our use cases  #(?).
Also in my previous PR (#616) I fixed this bug (#617) where a too small tolerance was leading to unexpected behavior.


**Absolute Tolerance**
```

ScalarType absoluteTolerance = epsilon * 10;  // ≈ 2.22×10⁻¹⁵ for double // line 298
ScalarType tolerance = std::max(relativeTolerance, absoluteTolerance);    // line 301

```

The absolute tolerance is just a safety net for near-zero comparisons.  I think in our case where we usually have scales of 5e8 seconds the relative tolerance (~10e-8) will always be larger. The absolute tolerance would only kick in if both values are extremely close to zero. So if this edge case would come up and the relative tolerance becomes too small, the comparison doesnt break down.



Please let me know what you think of this!

include/tudat/math/interpolators/lagrangeInterpolator.h
DominicDirkx: Thanks, that's very clear. So this is an issue in the current version of the code, your modification corrects for this. Nice!

DominicDirkx:
Thanks, that's clear! Using the 10 epsilon tolerance seems like a good solution. Just to confirm: there's not much difference in 
(i) the numerical stability close to the divide-by-zero points between your implementation and the new one (except of course that you now deal properly with them, but I mean for points slightly more than 10*epsilon away from them)
(ii) the computational efficiency is similar between the old and new implementation (same order is clear, but also the number of operations for a given n is similar right?)

Markus-Reichel:
i) the numerical stability close to zero but slightly more than 10*epsilon away from them should be fine as well. Lets say if `diff -> 0` but `diff > 10*epsilon` for point j:
- `term_j -> inf` // line 312-313
- `numeratorSum ≈ y_j × term_j` // line 314
- `denominatorSum ≈ term_j`
- `interpolatedValue = numeratorSum / denominatorSum ` should converge to `y_j` because `term_j` cancels out

ii) the computational efficiency is very similar:
**Old formula:**
Loop 1: n subtractions, n multiplications
Loop 2: 2n multiplications, n diviisions, n additions
Total: n subtractions, 3n multiplications, n divisions, n additions

**New formula:**
Single loop: n subtractions, 2n multiplications, n divisions, 2n additions
Final division: 1 division
Total: n subtractions, 2n multiplications, (n+1) divisions, 2n additions

So the new formula trades n multiplications for n additions + 1 division

DominicDirkx:
I posted on Slack to give people the chance to have a look at this code and express concerns if they have them. This PR will impact most Tudat analyses, so I'd like to give everyone the (short) oppurtunity to bring up any concerns.

PR #631, opened 2026-01-30
Add links to user guide in API sidebar

### LINKED ISSUE #518: [DOC]: Prototype Replication of User Guide structure in API reference
### Describe the issue:

Based on user feedback, the user guide is not visible enough when using the API reference. One idea to increase the visibility would be to integrate the user guide with the API reference. Since that involves a number of changes beyond the structure of the tudatpy repo, a first order prototype would be to mirror the same banner structure of the user guide in the API reference:

Image

with links to the getting started, examples, user guide pages.

### Additional context:

_No response_

larshinueber:
- Add links to user guide in API sidebar (fixes #518)
- Remove duplicate estimation heading


DominicDirkx:
Thanks :)

PR #634, opened 2026-01-31
Temporary fix to docs rendering

DominicDirkx:
This PR will (with a terrible vibe-coded solution, don't judge me....) remove the horrible input types in the API docs that popped up recently (I think with a package version update).

For instance

reference_area: typing.SupportsFloat, constant_force_coefficient: typing.Annotated[numpy.typing.ArrayLike, numpy.float64, "[3, 1]"], force_coefficients_frame: tudatpy.kernel.dynamics.environment_setup.aerodynamic_coefficients.AerodynamicCoefficientFrames = <AerodynamicCoefficientFrames.negative_aerodynamic_frame_coefficients: 2>) → [tudatpy.kernel.dynamics.environment_setup.aerodynamic_coefficients.AerodynamicCoefficientSettings](https://py.api.tudat.space/en/latest/dynamics/environment_setup/aerodynamic_coefficients.html#tudatpy.dynamics.environment_setup.aerodynamic_coefficients.AerodynamicCoefficientSettings)

 typing.Annotated[numpy.typing.ArrayLike, numpy.float64, "[3, 1]"] instead of numpy.ndarray
 
  typing.SupportsFloat instead of float
  
  This layout makes things very confusing. With numerical astro starting, I wanted to resolve this, and the best I could do on short notice is to make a modification that removes the types from the list altogether. The types wills till be listed in the parameter input list (and the stubs are unaffected). 

This fix is temporary but important to have quickly due to courses starting


src/tudatpy/data/mpc/mpc.py
```diff
@@ -1003,6 +1003,7 @@ def get_observations(
             # Assign the identifier to the 'number' column for the entire DataFrame.
             obs.loc[:, "number"] = identifier
             self._table = pd.concat([self._table, obs])
+            self._table = pd.concat([self._table, obs])
```
larshinueber: Was this intentional? This duplicates the observations and changes the behavior of the method.

larshinueber:
@DominicDirkx I took the liberty to have a look at this :)

While going through the implementation, I had an idea of preserving the signature typehints and format them nicely. I'll open a PR against your branch in a minute!

Edit: open in #635

PR #635, opened 2026-02-01
Simplify signature types through (naive) find and replace

larshinueber:
This PR simplifies the function signature docstrings by replacing the complex type annotations with more intuitive annotations through a rather naive find and replace in the docstring. Due to the limited amount of complex annotations, they are hard-coded in https://github.com/larshinueber/tudatpy/blob/cba2e6fa8e86b0658ce024d8534ad470ea849f39/docs/tudatpy/source/conf.py#L173

As an example, this produces the following function signatures:
Numpy arrays and floats:
image
Lists:
image

### DIFF
--- a/docs/tudatpy/source/conf.py
+++ b/docs/tudatpy/source/conf.py
@@ -75,7 +75,6 @@ autosummary_generate = True  # Turn on sphinx.ext.autosummary
 add_module_names = False
 autodoc_member_order = "groupwise"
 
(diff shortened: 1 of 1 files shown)


DominicDirkx:
Awesome, this is so much better than my horrible fix :) The only ones that this PR does not address (that I can see) are in overloaded functions, but there are very few of those and there should be none, so that's fine. Thanks!!

PR #636, opened 2026-02-01
Refactor deprecation functionality, add stubgen compatibility

larshinueber:
This PR:
- refactors the functionality to deprecate functions/properties in a backwards compatible manner to a separate module (`tudatpy._deprecation`)
- uses decorators to print deprecation warnings on invocation of the function/property instead of on import from the module
- adds compatibility to the stub generator for deprecation declarations in the `__init__.py` files

If an expression or assignment is detected during the stub generation in an `__init__.py` file, it is checked if the expression/assignment contains any of the deprecation function signatures. If that's the case, the expression/assignment is ignored and stub generation continues, thereby ignoring the deprecated functionality.

### DIFF
--- a/build.py
+++ b/build.py
@@ -212,6 +212,10 @@ class StubGenerator:
         # Define the path to the stub directory
         self.stubs_dir = build_dir / "tudatpy-stubs"
 
+        # Function names used for deprecation assignments
+        self.deprecation_function_names = [stat.name for stat in self.__parse_script(self.python_source_dir / "_deprecation.py").body if isinstance(stat, ast.FunctionDef)]
+
+
         return None
 
     def __parse_script(self, stub_path: Path) -> ast.Module:
@@ -869,6 +873,10 @@ class StubGenerator:
                         stub_body.append(statement)
                     continue
 
+                # No type hints for deprecation module
+                if "tudatpy._deprecation" in statement.module:
+                    continue
+
                 # External import statement
                 stub_body.append(statement)
 
@@ -892,12 +900,22 @@ class StubGenerator:
                     stub_body.append(statement)
                     continue
 
+                # Deprecation function assignments
+                if any(func_name in ast.unparse(statement) for func_name in self.deprecation_function_names):
+                    continue
+
                 # Any other assign statement is unexpected
                 raise NotImplementedError(
                     f"Failed to generate {stub_path}: "
                     f"Unexpected assign statement: "
                     f"{ast.unparse(statement)}"
                 )
+
+            # Ignore expression statements, e.g. for deprecation warnings
+            if isinstance(statement, ast.Expr):
+                # Deprecation function expressions
+                if any(func_name in ast.unparse(statement) for func_name in self.deprecation_function_names):
+                    continue
 
             # Other statements (We add them without modification)
             stub_body.append(statement)
--- a/build.py
+++ b/build.py
@@ -177,7 +177,7 @@ class StubGenerator:
     indentation: str = " " * 4
 
     # Ignored modules and methods
-    ignored_modules: list[str] = ["temp", "io", "numerical_simulation"]
+    ignored_modules: list[str] = ["temp", "io", "numerical_simulation", "_deprecation.py"]
     ignored_methods: list[str] = ["_pybind11_conduit_v1_"]
 
     def __init__(self, build_dir: Path, mock_env: "Environment") -> None:
--- a/src/tudatpy/_deprecation.py
+++ b/src/tudatpy/_deprecation.py
@@ -1,77 +1,194 @@
 import functools
 import warnings
 from types import ModuleType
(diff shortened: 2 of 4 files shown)


build.py
```diff
@@ -869,6 +873,10 @@ def __generate_single_init_stub(self, module_path: Path) -> None:
                         stub_body.append(statement)
                     continue
 
+                # No type hints for deprecation module
+                if "tudatpy._deprecation" in statement.module:
```
alfonsoSR, line 877: Checking from the phone, but I understand that this is removing the lines of the init.py including imports from "tudapy._deprecation" from the associated stub. Is it right?

build.py
alfonsoSR, line 1: Did you not add the "_deprecation" module to the list of ignored submodules because it does not have its own directory? That would make sense, but just to make sure I understand :)

alfonsoSR (commented):
Wasn't asked, but it looks great to me. I checked the code from the phone, so I haven't run it, but it looks much cleaner and less patchy than what I had done :)

build.py
larshinueber: Exactly, that was the intention 👍

build.py
larshinueber: That was my thought, yes 😅 I didn't realize it would also create stubs for top-level files (it makes sense :) ), so I'll add it to the list of ignored modules 👍

larshinueber:
> Wasn't asked

I'll refer to Dominic's comment on Slack ;)

> I checked the code from the phone, so I haven't run it, but it looks much cleaner and less patchy than what I had done :)

Thanks for having a look regardless!

DominicDirkx:
Thanks for the better fix! When running one of the examples with this, I get:

/home/dominic/miniconda3/envs/tudatpy-dev/bin/python /home/dominic/Tudat/tudatpy-examples/propagation/keplerian_satellite_orbit.py 
Traceback (most recent call last):
  File "/home/dominic/miniconda3/envs/tudatpy-dev/lib/python3.11/site-packages/tudatpy/estimation/observations/__init__.py", line 1, in <module>
    from tudatpy._deprecation import property_deprecation
ModuleNotFoundError: No module named 'tudatpy._deprecation'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/dominic/Tudat/tudatpy-examples/propagation/keplerian_satellite_orbit.py", line 29, in <module>
    from tudatpy.interface import spice
  File "/home/dominic/miniconda3/envs/tudatpy-dev/lib/python3.11/site-packages/tudatpy/interface/__init__.py", line 1, in <module>
    from . import spice
  File "/home/dominic/miniconda3/envs/tudatpy-dev/lib/python3.11/site-packages/tudatpy/interface/spice/__init__.py", line 1, in <module>
    from tudatpy.kernel.interface.spice import *
ImportError: initialization failed

Any idea how to fix it?

larshinueber:
Ahh, I wanted to add this to the PR  comment but forgot 😅
Since this adds a new module, you will have to
```
python uninstall.py
```
with your tudatpy conda env activated and then
```
python install.py -e
```
to update the module list. This should (hopefully) do it :)

DominicDirkx:
magic-henning

DominicDirkx:
Works now :) Thanks!

jo11he:
I have encountered the same issue as Dominic, and the uninstall / install fix does not work for me.
I am building with build.py, but not installing my local tudatpy builds into an environment. I add the path to the build/src on top of my scripts.
It seems like the uninstall / install fix is not compatible with this way of working, since it assumes there is a manifest in the build directory (which is a consequence of using the install.py script, right?)? 

Is there a way to keep supporting the build through build.py without the use of the install.py step? Or will I have to change my way of working?

DominicDirkx:
There's probably an easier/faster way, but deleting the build folder and rebuilding should from scratch should do the trick

build.py
alfonsoSR: Oh, I am actually not sure if it does, could also be the case that I missed that option

alfonsoSR:
@jo11he sorry about not having answered yesterday. 

When you run `build.py`, the script makes a copy of all the python files in `src/tudatpy` in the `<build-dir>` you chose. It makes sense to me that if you pull updates including new directories or `.py` files you would have to re-run `build.py` to avoid getting those errors, even if you don't use the `uninstall/install.py` tools.

> There's probably an easier/faster way, but deleting the build folder and rebuilding should from scratch should do the trick

I think that re-running `build.py` without deleting the previous build should also work :)

jo11he:
Thanks @alfonsoSR. I was building from scratch. I think the real problem was that I was working with an outdated conda environment... 
I updated the environment, build, ran my script and am now running into a different problem during imports. I cannot tell if it is before of after the previous error, but I also do not need this new build right now. 
I will pick this up next week and if I cannot fix it by myself I will open an issue :)

PR #655, opened 2026-02-11
Add space-time metric, relativistic time propagation, and INPOP tests

kimonito98:
**WIP - FEATURE**

unit tests passing (see document attached) except:
testCombinedProperTimeAndStateDynamics2: compares direct metric-based proper-time propagation with concatenated barycentric-plus-topocentric propagation --> divergence of 1e-4 after 1 year
[spacetime.pdf](https://github.com/user-attachments/files/25222944/spacetime.pdf)


DominicDirkx:
Thanks for adding this! Will be great to have this code in Tudat, finally updated to modern setup :) I see in the changed files of the PR that the data.zip is removed (and as a result all CI is failing). Could you put this file back (and add any test files that you may need, which there are probably a lot of!)

PR #661, opened 2026-02-13
Feature/eliminate duplicate obs in set construction to implement #659

### LINKED ISSUE #659: [F&E]: SingleObservationSet check for duplicate observations
### Description:

Some  ODF files (in the context of the cassini mission) result in duplicate observations (same epoch, same value) within the same SingleObservationSet. There could be a check (with user control) to eliminate these duplicates.

Define a member function that identifies duplicates and using the existing eraseDuplicates by index member functions of SingleObservationSet removes them. This function should be called in the constructor, toggled by a boolean in the constructor signature. The boolean should default to false (no elimination of duplicates) to avoid inconsistencies between transfer of set data from old to new sets like in e.g. observation set splits. 

Instead, I propose to use a boolean on the createOdfObservedObservationCollectionFromFile signature that connects to the first call of the SingleObservationSet constructor, which eliminates duplicates (if user wishes) right at the first creation of the ObservationCollection.

I am working on this already.

### Additional context:

_No response_

jo11he:
Addresses feature proposed in https://github.com/tudat-team/tudatpy/issues/659

### DIFF
--- a/include/tudat/simulation/estimation_setup/processOdfFile.h
+++ b/include/tudat/simulation/estimation_setup/processOdfFile.h
@@ -1341,8 +1341,8 @@ std::shared_ptr< observation_models::ObservationCollection< ObservationScalarTyp
                                 std::vector< Eigen::VectorXd >( ),
                                 nullptr,
                                 std::make_shared< observation_models::ObservationAncillarySimulationSettings >(
-                                        ancillarySettings.at( i ) ) ),
-                                !allowDuplicateObservationsWithinSingleObservationSet);
+                                        ancillarySettings.at( i ) ),
+                                !allowDuplicateObservationsWithinSingleObservationSet) );
 
             }
         }
@@ -1583,7 +1583,10 @@ createOdfObservedObservationCollectionFromFile( simulation_setup::SystemOfBodies
     observation_models::setOdfInformationInBodies( processedOdfFileContents, bodies );
 
     // Create observed observation collection
-    return observation_models::createOdfObservedObservationCollection< ObservationScalarType, TimeType >( processedOdfFileContents );
+    return observation_models::createOdfObservedObservationCollection< ObservationScalarType, TimeType >(
+        processedOdfFileContents, std::vector< observation_models::ObservableType >(),
+        std::make_pair< TimeType, TimeType >( TUDAT_NAN, TUDAT_NAN ),
+        allowDuplicateObservationsWithinSingleObservationSet);
 }
 
 }  // namespace observation_models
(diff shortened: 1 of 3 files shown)


jo11he:
I tested this on my ODF files, it seems to be working well. I did not include this into any unit tests.

include/tudat/simulation/estimation_setup/processOdfFile.h
```diff
@@ -1251,7 +1251,8 @@ template< typename ObservationScalarType = double, typename TimeType = Time >
 std::shared_ptr< observation_models::ObservationCollection< ObservationScalarType, TimeType > > createOdfObservedObservationCollection(
         std::shared_ptr< ProcessedOdfFileContents< TimeType > > processedOdfFileContents,
         std::vector< observation_models::ObservableType > observableTypesToProcess = std::vector< observation_models::ObservableType >( ),
-        std::pair< TimeType, TimeType > startAndEndTimesToProcess = std::make_pair< TimeType, TimeType >( TUDAT_NAN, TUDAT_NAN ) )
+        std::pair< TimeType, TimeType > startAndEndTimesToProcess = std::make_pair< TimeType, TimeType >( TUDAT_NAN, TUDAT_NAN ),
+        const bool allowDuplicateObservationsWithinSingleObservationSet = false)
```
DominicDirkx: Could you set the default to true? Otherwise we make the code non-backwards compatible

include/tudat/simulation/estimation_setup/singleObservationSet.h
```diff
@@ -696,6 +705,38 @@ class SingleObservationSet
...
+        std::vector<unsigned int> indicesToRemove;
+
+        // Single pass through sorted observations
+        for(unsigned int i = 1; i < numberOfObservations_; i++)
+        {
+            // Check if current observation time equals previous observation time
+            if(observationTimes_[i] == observationTimes_[i-1])
+            {
+                // Check if observation values are also identical
+                if(observations_[i].isApprox(observations_[i-1]))
```
DominicDirkx: why add this logic? and what is 'isApprox'?

include/tudat/simulation/estimation_setup/processOdfFile.h
jo11he: done.

include/tudat/simulation/estimation_setup/singleObservationSet.h
jo11he: I added this to insure that it is not two (numerically) different observations referencing the same receiver time, but also that the observation values are the same - so what I would consider a true duplication. 

The isApprox is a function from the Eigen library which does vector and matrix comparison using a relative tolerance. I researched this a bit more just now and found it might as well be replaced by std::abs(a - b) < tol, because this is just a scaler comparison.

jo11he: If we want to keep this, I could implement 

(std::abs(a - b) <= 1E-09) --> absolute comparison with relatively high tolerance


(std::abs(a - b) <= 1E-12 * std::max(std::abs(a), std::abs(b))) --> relative comparison with lower tolerance


Which one do we prefer?

jo11he: I implemented the second version now to set off CI again, but I will adjust if that is not desired.

DominicDirkx: I think it will depend on the use case, but for you the filter should only remove observations if they are really duplicate in time and value right? Then, I think we can keep it as is for now. If we need a 'full' duplicate time filter, we can always add it :)

DominicDirkx (approved):

PR #663, opened 2026-02-16
Feature/compilation time improvement

DominicDirkx:
## Rationale
Kernel and test builds showed high compile cost and memory usage, concentrated in a relatively small number of files (especially estimation/propagation wrappers and simulator/variational code paths and large tests). This caused long build times, unstable CI behavior for macOS due to timeouts, and frequent long recompile times when touching core estimation/propagation headers.

The goal of this PR is to reduce compile and memory usage

## Some aspects of the implementation
- Split large umbrella headers/classes into targeted components (keeping umbrella includes for compatibility)
  - dynamics simulator split into base/single/multi/hybrid arc headers
  - variational equations solver split into base/single/multi/hybrid arc headers
  - observation model setup split into settings/factory-focused headers
  - bodies/environment setup split (`createBodiesSettings` / `createBodiesFactory`)
- Reduced includes across Tudat and TudatPy:
  - migrated many internal includes from umbrella headers to targeted headers
  - introduced/used forward-declaration headers for estimation/observation interfaces
  - moved default ground-station helper/settings logic out of `defaultBodies` into dedicated files
- Added explicit-instantiation infrastructure for costly template paths:
  - dedicated extern-template headers/sources in kernel build path
  - one major template instantiation per `.cpp` to avoid monolithic instantiation units
- Added use of unity build for tudat libraries and precompiled headers for Python exposure and main classes of unit tests (these are both useful for a fresh build from scratch, in some cases will increase recompilation time after small changes, these settings can be toggled on/off)


PascalChristiaanse:
Hi @DominicDirkx , i did some experimentation using precompile headers and unity building and got about a 30% decrease in build times. Should i open a seperate branch for these changes?
Heres some results i got

Command
	CCache -C (full clean)
	time python build.py -j4 --clean

without unity/pch
	real    16m2.111s
	user    54m39.884s
	sys     4m33.278s	

with unity/pch
	real    10m44.825s
	user    36m35.193s
	sys     2m44.268s

	
Changes:
	Enabling PCH on std and pybind headers
	Enabling Unity builds reducing translation units from ~400 	
	~200
	Enabling CCache as default, not just on github actions
        Set regular build optimization to -O0 for tests, no need for fast code and drastically reduces memory usage

Additionally, debug builds should probably be on -O1, while release should be on either -O3 or -O4. This takes a lot less memory allowing higher corecounts to be used when compiling.

DominicDirkx:
@PascalChristiaanse If you could open a separate branch/PR for that, it would be great!

larshinueber:
@DominicDirkx Thanks for all the updates! Once ready to be merged, maybe you could squash & merge the commits, since the exact history of all the header modifications is not really relevant? This way we end up with a nice and clean git history :)

cmake/tudat/compiler.cmake
```diff
@@ -367,7 +367,7 @@ if (MSVC)
   message(STATUS "Setting /bigobj")
   add_compile_options(/bigobj)
 else()
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O3")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O0")
```
larshinueber: @DominicDirkx was this intentional? Since local builds are used for both development and running simulations (as opposed to Github CI builds), maybe leaving it to `-O3` is more sensible if it doesn't affect compile time too much?

changes.patch
larshinueber, line 1: I believe this is not needed 😄

CMakeLists.txt
```diff
@@ -57,7 +57,7 @@ if (MSVC)
     # This is equivalent to the big-obj for MinGW, this should stop error C1128.
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /bigobj")
 else()
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O3")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O0")
```
larshinueber: See comment in `cmake/tudat/compiler.cmake`

residual_test.m
larshinueber, line 1: Also slipped through :)

cmake/tudat/compiler.cmake
DominicDirkx: Absolutely! Thanks for reminding me, this was meant to be a temporary fix, the -O0 should be used for the CI (where build time/RAM usage is the main bottlneck) and possibly debugging. There should be a CMake argument to control this, but I was too lazy to add this so far :)

changes.patch
DominicDirkx: Oops...

larshinueber:
@DominicDirkx I took the liberty to ask Claude to review the PR since it turned out to be massive and beyond my C++ knowledge 😅 The review is overall positive, with a few recommendations to be addressed.

- I've fixed recommendation 3 (renaming of the `TUDAT_BUILD_ALL_TESTS` flag).
- For recommendation 1 and 5 I would argue that `createObservationManager.cpp` already serves as file for the explicit template instantiation, but renaming it and removing the commented out code could make it more explicit?
- I'm not sure about recommendation 2, since the CI works fine and the flag is not used for DLL copying elsewhere.
- Adding a note for Tudat C++ users about the refactored headers sounds useful, maybe in the form of a slack message?

<details><summary>Claude Code Review</summary>
<p>

# Code Review: PR #663 -- Feature/compilation time improvement

**Repo:** tudat-team/tudatpy | **Base:** develop | **Author:** DominicDirkx
**Stats:** 613 files changed, +25,052 / -25,899 (net -847 lines)
**State:** Open

---

## Overview

This PR is a comprehensive compilation-time reduction effort targeting four complementary strategies:

1. **Header splitting** -- Monolithic "god headers" broken into targeted components
2. **Include reduction** -- Forward declarations, targeted includes replacing umbrella headers
3. **Explicit template instantiation (ETI)** -- `extern template` + dedicated `.cpp` files for `<double, double>`
4. **Build system** -- Unity builds, precompiled headers (PCH), centralized optimization policy

The changes are purely structural/build-level. No functional or API-facing changes are introduced.

---

## Architecture Analysis

### Header Splitting (Well executed)

Five major monolithic headers were decomposed:

| Original Header | Lines Removed | Split Into |
|---|---|---|
| `body.h` | ~1120 | Forward declarations + out-of-line methods + `baseStateInterface.h`, `rigidBodyProperties.h` |
| `dynamicsSimulator.h` | ~2608 | `dynamicsSimulatorBase.h`, `singleArc...`, `multiArc...`, `hybridArc...` + forward decl header |
| `variationalEquationsSolver.h` | ~2212 | Same base/single/multi/hybrid pattern + forward decl header |
| `createEstimatableParameters.h` | ~3440 | `...Interface.h` (lightweight) + `...Factory.h` (heavy template code) |
| `createObservationModel.h` | ~3608 | `...Settings.h` + `...Factory.h` |
| `orbitDeterminationManager.h` | ~1186 | 5 implementation headers + forward decl header |

All original headers are preserved as umbrella includes for backward compatibility.

The `body.h` change is the most impactful -- it replaces 15+ subsystem `#include`s with forward declarations of aerodynamics, gravitation, electromagnetism, propulsion, ground stations, and system model types. Since `body.h` is transitively included by nearly every simulation/estimation file, this single change dominates the compilation savings.

### Forward Declaration Headers (Clean design)

Three new forward-declaration headers were introduced:
- `estimationInterfacesForwardDeclarations.h` -- CovarianceAnalysisInput/Output, EstimationInput/Output, OrbitDeterminationManager
- `observationInterfacesForwardDeclarations.h` -- ObservationModelSettings, ObservationCollection, ObservationSimulatorBase, etc.
- `dynamicsSimulatorForwardDeclarations.h` / `variationalEquationsSolverForwardDeclarations.h`

No circular dependency risks detected. The dependency graph remains a clean DAG.

### Explicit Template Instantiation (Sound, one bug found)

9 new `.cpp` files provide `<double, double>` instantiations for the dynamics simulator and variational equations solver class hierarchies, plus `createObservationManagersBase`. Each file contains exactly one explicit instantiation -- clean and consistent.

**Bug:** `createObservationManager.h` declares `extern template` for both `createObservationManagerBase<double,double>` and `createObservationManagersBase<double,double>`, but only the latter has an explicit instantiation `.cpp` file. The missing instantiation for `createObservationManagerBase` (singular) will cause a **linker error** when `TUDAT_BUILD_ALL_TESTS=1` and the function is called.

### Binding File Splits (Transparent to Python)

Four large pybind11 binding files were split into 13 focused compilation units using a delegate/orchestrator pattern. No Python API changes -- module names, function names, argument names, and docstrings are all preserved. The `.h` files were reduced to minimal `#include <pybind11/pybind11.h>` only, moving all heavy includes to `.cpp` files.

### Spice Exception Decomposition (Clever)

The monolithic `spiceException.h` (~4522 lines defining 898 exception classes) was decomposed using an X-macro pattern:
- `spiceError.h` -- base class only (45 lines)
- `spiceExceptionList.def` -- data-driven macro list
- `spiceExceptionTypes.h` -- macro-generates all exception classes
- `spiceInterface.h` now includes only `spiceError.h`

---

## Build System Analysis

### CMake Changes (Well-structured)

| New Option | Default | Purpose |
|---|---|---|
| `CMAKE_UNITY_BUILD` | OFF | Unity build for tudat libraries |
| `CMAKE_UNITY_BUILD_BATCH_SIZE` | 8 | TUs merged per batch |
| `TUDAT_BUILD_WITH_PCH` | ON | Precompiled headers for kernel target |
| `TUDAT_BUILD_WITH_TEST_PCH` | OFF | PCH for test support libraries |
| `TUDAT_BUILD_FOR_REDUCED_COMPILE_TIME` | OFF | Forces `-O1`/`/O1` across all build types |
| `TUDAT_BUILD_ALL_TESTS` | ON | Gates ETI `.cpp` files and `extern template` |
| `TUDAT_FORCE_CMAKE_CACHE_RESET` | OFF | Replaces unconditional cache deletion |

Key improvements:
- **Centralized optimization policy** (`tudat_apply_optimization_policy` macro in `compiler.cmake`) replaces scattered `-O3` flags with a single, configurable source of truth
- **Unconditional cache deletion removed** -- was destroying incremental builds
- **ccache auto-detection** added
- **Test PCH** via `REUSE_FROM` pattern -- one PCH per test domain reused across 10-30 test targets
- **Unity build disabled** for pybind11 `kernel` target and all test executables (correctly)

### CI Changes

- Linux parallelism: 2 -> 3 jobs; Windows: 3 -> 2 (adjusts for MSVC memory pressure)
- New resolve steps for per-OS build type and ccache mode
- All new build options enabled on CI (`PCH=ON`, `UNITY_BUILD=ON`, `REDUCED_COMPILE_TIME=ON`)

---

## Issues Found

### Critical

1. **Missing explicit instantiation for `createObservationManagerBase<double,double>`** -- The header declares `extern template` but no `.cpp` provides the instantiation. This will cause an undefined symbol linker error when `TUDAT_BUILD_ALL_TESTS=1`.

### Medium

2. **`TUDAT_BUILD_GITHUB_ACTIONS` repurposed on Windows** -- The new `resolve_ci_ccache_mode` step sets this to `OFF` on Windows to avoid a ccache+MSVC crash. But other CMake logic may depend on this flag for CI-specific behavior (e.g., MSVC runtime DLL copying). This could silently break Windows CI packaging.

3. **`TUDAT_BUILD_ALL_TESTS` naming is misleading** -- This option does not control test compilation (that's `TUDAT_BUILD_TESTS`). It controls explicit template instantiations. A name like `TUDAT_BUILD_EXPLICIT_INSTANTIATIONS` or `TUDAT_BUILD_ETI` would be clearer.

4. **Downstream transitive include breakage risk** -- External projects that relied on transitive includes through `body.h` (e.g., using `AerodynamicCoefficientInterface` member functions without explicitly including the aerodynamics headers) will get incomplete-type errors. The internal codebase was updated, but external users may need migration guidance.

### Low

5. **ccache detection after `project()`** -- Works for Ninja/Make generators but not guaranteed for all CMake generators. Acceptable for this project's use case.

6. **Debug optimization level `-O1` instead of `-Og`** -- Degrades debugger experience (variables optimized out). Deliberate trade-off but worth documenting.

7. **Inconsistent test support library gating** -- `tudat_estimation_test_case_support` is always built; aerodynamics/observation support libraries are conditional on `TUDAT_BUILD_WITH_TEST_PCH`. Should be unified.

8. **Pre-existing: `PRIVATE_LINKS` applied as `PUBLIC`** in `YOLOProjectAddTestCase.cmake` -- Not introduced by this PR but worth noting.

### Cosmetic

9. Mixed indentation in `tudat_apply_optimization_policy` macro
10. Column alignment inconsistency in `show_build_configuration.cmake`

---

## Strengths

- **Zero API breakage** -- All Python/C++ interfaces preserved
- **Clean backward compatibility** -- Umbrella headers preserved as thin wrappers
- **Principled decomposition** -- Consistent base/single/multi/hybrid arc pattern across both dynamics simulator and variational equations solver
- **All features toggleable** -- Every new mechanism has an `option()` with sensible defaults
- **Cross-platform awareness** -- MSVC vs GCC/Clang flag handling, GNU `ld` `--start-group`/`--end-group` workaround, Windows ccache avoidance
- **Well-chosen PCH headers** for pybind11 bindings (pybind11.h, stl.h, eigen.h, Eigen/Core, Eigen/Geometry, STL containers)

## Recommendations

1. **Fix the missing `createObservationManagerBase` instantiation** before merging
2. **Verify `TUDAT_BUILD_GITHUB_ACTIONS=OFF` on Windows** doesn't break MSVC runtime DLL logic
3. **Rename `TUDAT_BUILD_ALL_TESTS`** to something that reflects its actual purpose (ETI gating)
4. **Add migration notes** for downstream users about `body.h` transitive include changes
5. **Consider adding a second ETI file** or expanding the existing one for `createObservationManagerBase<double,double>`

---

## Verdict

This is a high-quality, well-engineered compilation-time optimization PR. The architectural changes are principled and consistent. The one critical bug (missing ETI for `createObservationManagerBase`) and the naming concern (`TUDAT_BUILD_ALL_TESTS`) should be addressed before merge. The risk of downstream transitive include breakage is real but manageable with documentation.
</p>
</details>

DominicDirkx:
Thanks! I also processed point 6 (-Og for debug unless compilation time is to be optimized).

Coming to your points:

> 
> 1. I've fixed recommendation 3 (renaming of the TUDAT_BUILD_ALL_TESTS flag).

Thanks!

> 2. For recommendation 1 and 5 I would argue that createObservationManager.cpp already serves as file for the explicit template instantiation, but renaming it and removing the commented out code could make it more explicit?

For point 1, the instantiation does happen for both both functions, but in different files (by design, to reduce per translation unit compilation memory, preventing RAM bottlenecks and permitting higher compilation parallelization)

> 3. I'm not sure about recommendation 2, since the CI works fine and the flag is not used for DLL copying elsewhere.

Right now, this is not relevant, since ccache is disable on CI. Once we turn it back on, we could look into this

> 4. Adding a note for Tudat C++ users about the refactored headers sounds useful, maybe in the form of a slack message?

Absolutely!

PR #671, opened 2026-02-18
Feature/build remote unit tests

luigigisolfi:
@DominicDirkx, @alfonsoSR , I have been working with spacetrack unit tests, which require user authentication to run.

if users run those from `tests`, then they can in principle add the decorator `@pytest.mark.parametrize.remote_data` to online tests and then run

`pytest --remote-data`

to allow for those to run as well (they'd be skipped otherwise).

However, in order to  use
`pytest --remote-data`
they need to have credentials exported to their env path. 

I added a check in `build.py` that checks whether the credentials needed for those remote_data tests have been set by the user, and if they have not, those remote tests are skipped. 

In other words, this:
`python build.py --tests`
will trigger the credentials check a print a warning if no credentials for those functionalities are found. 

This PR shows a very raw first iteration of it, only checking for the spacetrack credentials. If you like the idea I can make it better and expand it to other `tudatpy.data` tests requiring` --remote-data`

EDIT: Just realised commit history is a bit messy, to see the relevant changes in `build.py`, just check the last commit


DominicDirkx:
I really like the idea of this! Also to make the runs of the tests on Azure more robust (which is in part what motivated this right?)

I see quite some changes in the code in this PR, but I think only a part of them is to support this change in testing right?

luigigisolfi:
@DominicDirkx I did not think about fancier applications than just allowing users to run both offline and online tests with:

`python build.py --tests`, 
but I guess it is a first step towards making them more robust yes. 

As i mentioned in the edit to the PR, the only relevant commit for this functionality is the last one, namely [this one](https://github.com/tudat-team/tudatpy/pull/671/commits/670e7c086d2741d814b08a66195255e945912040)

DominicDirkx:
@luigigisolfi I have modified the setup so that the default is that it *does* use remote data, and provides a warning if a test requires credentials that are ot available (skipping the test)

DominicDirkx:
Waiting for #834  to be merged before merging this

PR #672, opened 2026-02-18
Feature/build time improvements

PascalChristiaanse:
Enable ccache on all builds
set precompiled headers
set unity / blob building to reduce translation units
reduce optimization level on tests


DominicDirkx:
@PascalChristiaanse I had to drop what I was doing on this: https://github.com/tudat-team/tudatpy/pull/663 (mostly) to focus on some other things. But, I think the changes you've made here complement the changes there very well. When I get back to it (probably next week), I'll merge this PR into mine (and make modifications if needed)

DominicDirkx:
Now included (partially) into #663 The PCH in particular has been quite effective!

Issue #676, opened 2026-02-25
[F&E]: Add alternative astropy -> ObservationCollection support function in mpc.py


luigigisolfi:
### Description:

The `to_tudat()` method of `mpc.py` allows to convert a batch MPC object into an observation collection. To do this, it needs the a Tudat `bodies` object. This is valuable, as it allows to automatically set the link ends within the `to_tudat()` method itself, but it might be an overkill if users are just interested in converting their own observations from an `astropy` table into an `ObservationCollection` object. 

On a similar note, `to_tudat()` has the following flags: 

`apply_weights_VFCC17: bool = False`
`apply_star_catalog_debias: bool = False`

set to True by default. This is again valuable, as it allows to debias observations and set weights, but it might be an overkill for those users who just want to play around by manipulating their raw observations. Even more so, setting those values to `True` by default triggers a validation step for the table, that checks for some required column names, and fails if they are not found. Although the absence of such columns from an MPC table is unusual, users might just have minimal astropy tables with just RA, Dec, Epoch and Observatory name columns, where all other mandatory columns are actually missing. 

For this reasons, a complementary function will be made available, called `create_observations_from_astropy_table` that is similar in scope to `to_tudat()` but does not require the `bodies` object and, although it allows for catalog debiasing and weights settings, it sets those flags to `False` by default.

PR #677, opened 2026-02-25
Feature/create obs collection from astropy

luigigisolfi:
This PR addresses #676 by adding `create_observations_from_astropy_table()` method to `mpc.py`

### DIFF
--- a/src/tudatpy/data/mpc/mpc.py
+++ b/src/tudatpy/data/mpc/mpc.py
@@ -838,7 +838,7 @@ class BatchMPC:
 
     # data retrieval options: from_file: allows external observations to be added
     def from_file(
-        self, filename: str, in_degrees: bool = False, frame: str = "J2000"
+        self, filename: str, frame: str = "J2000"
     ) -> None:
         """
         Loads observations from a local MPC 80-column text file.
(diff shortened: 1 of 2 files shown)


DominicDirkx:
@luigigisolfi Could you add a unit test for this functionality?

luigigisolfi:
> @luigigisolfi Could you add a unit test for this functionality?
Done

DominicDirkx:
Two comments:

* Can the new pyarrow dependency be removed?
* There is a mixing of defaults of the in_degrees inputs (some default to False, some default to True). This seems like a risky situation, and one that will easily lead to errors/misunderstandings

DominicDirkx:
Not the default I would've gone for, but I'll take it ;) (and this is the one that is backwards compatible right?)

luigigisolfi:
> There is a mixing of defaults of the in_degrees inputs (some default to False, some default to True). This seems like a risky situation, and one that will easily lead to errors/misunderstandings

although this is easy to change by setting it to "True (or False, but it has always been True so we have to stick to it to make it backwards compatible :P)" for `create_observations_from_astropy_table`, the same cannot really be done for the `from_file()` method. This is because the `from_file` method parses an 80col file (containing Ra and Dec in HH:mm:ss.ddd and DD:mm:ss.ddd respectively (so Ra is neither degrees nor radians, but Hours...), which returns the Ra Dec in radians (this is a design choice I had made). 

We could use an `in_degrees` flag in the parsing function as well, to tell the parser how to **OUTPUT** the data but, anywhere else in the code, `in_degrees` refers to the format of **INPUT** coords, so this will still be misleading. 

My suggestion is therefore to remove the `in_degrees` flag from the `from_file()` method and hardcode it into the `from_file()`'s `from_astropy` call.

DominicDirkx:
> We could use an in_degrees flag in the parsing function as well, to tell the parser how to OUTPUT the data but, anywhere else in the code, in_degrees refers to the format of INPUT coords, so this will still be misleading.

I'd say that the *output* should always be in radians (since this output will be directly put into a Tudat observation collection). So: input in whichever units it comes (with suitable user-defined settings), but output always in radians. What do you think? Is

luigigisolfi:
> So: input in whichever units it comes (with suitable user-defined settings), but output always in radians.

that's how it is :)

PR #699, opened 2026-03-23
Addressed issue #675

### LINKED ISSUE #675: [F&E]: Improve setter flexibility of observation model parameters
### Description:

When estimating observation model parameters such as e.g. the ConstantObservationBiasParameter, the getter/setter interface to the associated models is not established until the creation of the Estimator (OrbitDeterminationManager) Object. Any attempt to manipulate the parameter vector (from EstimatableParameterSet object) before the creation of the Estimator and the associated "closure" of parameters and observation models results in an exception (https://github.com/tudat-team/tudatpy/blob/dcab5f828936d99e5f2fe1ad6bedac7f7799eeb3/include/tudat/astro/orbit_determination/estimatable_parameters/observationBiasParameter.h#L395). 

This requires tricky workarounds (create Estimator w/o integration or use of dummy iteration) that have also led to some issues regarding the correct setup of the Estimator object. A more convenient solution would be (as suggested by @DominicDirkx) to defer the setting of parameter values inside the parameter class (implementation suggestion below, untested).

diff --git a/include/tudat/astro/orbit_determination/estimatable_parameters/observationBiasParameter.h b/include/tudat/astro/orbit_determination/estimatable_parameters/observationBiasParameter.h
index 64a2fc339..ed4844804 100644
--- a/include/tudat/astro/orbit_determination/estimatable_parameters/observationBiasParameter.h
+++ b/include/tudat/astro/orbit_determination/estimatable_parameters/observationBiasParameter.h
@@ -262,7 +262,8 @@ public:
         }
         else
         {
-            throwExceptionIfNotFullyDefined( );
+            deferredValueToSet_ = parameterValue;
+            // throwExceptionIfNotFullyDefined( );
         }
     }

@@ -294,6 +295,11 @@ public:

         getBiasList_ = getBiasList;
         resetBiasList_ = resetBiasList;
+
+        if( deferredValueToSet_.rows(  ) > 0 )
+        {
+            setParameterValue( deferredValueToSet_ );
+        }
     }

     //! Function to retrieve the observation link ends for which the bias is active.
@@ -403,6 +409,8 @@ private:

     //! Object used to determine the current arc, based on the current time.
     std::shared_ptr< interpolators::LookUpScheme< double > > lookupScheme_;
+
+    Eigen::VectorXd deferredValueToSet_ = Eigen::VectorXd::Zero( 0 );
 };

### Additional context:

_No response_

DominicDirkx:
Closes #675


jo11he:
I tested the functionality introduced by this PR in my cassini estimation setup.
It is now possible to set observation bias values in the parameter vector before the "closure" of the observation model in the estimator creation. The set values are also reflected in the 0th entry of the parameter_history attribute of the EstimationOutput object. 

Thank you very much, this will allow me to make my scripts much more elegant!

One additional remark: originally (before setting any values) in the parameter vector, the values of the "dangling" parameters appear as nan. This means that if someone copies the original vector, applies a perturbation by summation with another (perturbing) vector, the resulting vector will still contain these nan entries and the user is inadvertently setting these values to nan, which will then crash the estimation. This is easy to spot and troubleshoot, but maybe there could also be a safe and convenient handling of this (although I cannot think of one).

jo11he:
And I now see there are issues with unit tests on two platforms, and interestingly it is not consistently the same tests failing, did you see this already @DominicDirkx ?

DominicDirkx:
I did, it's related to another issue that had to be solved due an update in OS image, cmake version or something similar. It's rounding error that falls slightly differently than before, unrelated to the code in this PR

Issue #716, opened 2026-04-14
[F&E]: Resolving Pandas SettingWithCopyWarning in test_data_weights.py


luigigisolfi:
### Description:

When running 

`pytest test_data_weights.py`, 

the following warning shows up:

```
tests/test_tudatpy/test_data_weights.py::test_MPC_weights_to_ObsCol[use_dummy_weights0-observatories_to_filter0-True]
tests/test_tudatpy/test_data_weights.py::test_MPC_weights_to_ObsCol[use_dummy_weights1-observatories_to_filter0-True]
  /Users/ama/Downloads/tudatpy/.pixi/envs/default/lib/python3.13/site-packages/tudatpy/data/mpc/mpc.py:826: SettingWithCopyWarning: 
  A value is trying to be set on a copy of a slice from a DataFrame.
  Try using .loc[row_indexer,col_indexer] = value instead
  
  See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
    table["epoch_seconds_UTC"] = [dt_obj.epoch() for dt_obj in dt_objects]

tests/test_tudatpy/test_data_weights.py::test_MPC_weights_to_ObsCol[use_dummy_weights0-observatories_to_filter0-True]
tests/test_tudatpy/test_data_weights.py::test_MPC_weights_to_ObsCol[use_dummy_weights1-observatories_to_filter0-True]
  /Users/ama/Downloads/tudatpy/.pixi/envs/default/lib/python3.13/site-packages/tudatpy/data/mpc/mpc.py:829: SettingWithCopyWarning: 
  A value is trying to be set on a copy of a slice from a DataFrame.
  Try using .loc[row_indexer,col_indexer] = value instead
  
  See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
    table["epoch_seconds_TDB"] = [

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

### Suggestions for implementation:

Create a copy of the pandas table before augmenting it with the UTC and TDB columns, to let pandas know we are safe as we are dealing with two different objects.

DominicDirkx:
Now resolved right

PR #717, opened 2026-04-14
Resolving Issue #716 (Pandas warning in test_data_weights.py)

luigigisolfi:
This PR solves #716 by creating a copy of the pandas table before augmenting it with the needed time columns.


minervino96 (approved):
No more warnings :)

PR #745, opened 2026-04-22
Correct leap seconds in DateTime

DominicDirkx:


include/tudat/astro/basic_astro/dateTime.h
```diff
@@ -95,6 +97,102 @@ TimeType timeFromIsoString( const std::string &isoTime )
     return timeFromDecomposedDateTime< TimeType >( year, month, days, hours, minutes, seconds );
 }
 
+inline constexpr std::array< std::array< int, 3 >, 27 > leapSecondIntroductionDays = {{
```
larshinueber, line 100: Instead of hardcoding, the leap second introduction dates could also be retrieved from the `DELTET/DELTA_AT` variable of an lsk kernel using the `gdpool` function: https://naif.jpl.nasa.gov/pub/naif/misc/toolkit_docs_N0067/C/cspice/gdpool_c.html
This way we dont have to maintain the list ourselves 😅
https://naif.jpl.nasa.gov/pub/naif/generic_kernels/lsk/naif0012.tls

DominicDirkx:
I considered that, and conceptually agree with your suggestion, but I'd like (i) for this functionality to not have to link to spice or sofa and (ii) some compile-time checks to be done, which is not possible with spice kernels.

Also, leap seconds will be phased out (yay, finally!) So I'd propose to keep it as is for now.

larshinueber:
Thank you for the context, that makes sense :)

PR #750, opened 2026-04-24
Fix frame mismatch in 1st-order TCB->proper-time direct correction

kimonito98:
The station position function feeding the -v_E.r/c^2 direct correction in TimeEphemerisWithFirstOrderDirectConversion was bound to GroundStationState::getNominalCartesianPosition, which returns the body-fixed position. That vector was dot-producted with the central body's inertial barycentric velocity, so the frames did not match and Earth's diurnal rotation was silently dropped.

The IBP identity underlying the correction,
    integral( v_E . v_rel dt ) = [v_E . r] - integral( a_E . r dt ),
requires v_rel = dr/dt in the inertial frame, which means r must be inertial too. The resulting bug produced a ~2 us/day periodic residual against direct-from-metric: 4.91 us detrended amplitude over one year against a 6 ns physical truncation floor.

Fix (setNumericallyIntegratedStates.h:1254-1272): rotate the body-fixed nominal position to the inertial frame at each query time using the central body's rotational ephemeris before storing the function in groundStationPositionFunctions_. The 1-year strict test dropped to 3.89 ns detrended amplitude (slope 1.4e-16 s/s), passes 6 ns tol.

Tests:
 * Added testDirectFromMetricVsChainedPnEquatorOneYearEarthMoonSun (Earth eq. station, ggm02c deg 20, Moon deg 2, Sun, 100 s step, 1 year, first-order PN both sides, topocentric useAccelerationTerm OFF as paired with the -v_E.r/c^2 direct correction under IBP). Passes at 3.89 ns / 6 ns tolerance.
 * Instrumented testCombinedProperTimeAndStateDynamics2 to print actual residual magnitudes (slope / offset / max|raw| / det_amp / endpoint). After the fix, the loose-tolerance 2nd-order test passes at det_amp 116 us - an independent, larger bug remains in the 2nd-order pipeline (see below).

Open investigation - 2nd-order chain vs direct (testCombinedProperTimeAndStateDynamics2):
 * Sign audit of calculateSecondOrderTcbToTcgIntegrand against the IAU BCRS derivation: all four principal coefficients (-v^4/8, -3/2 v^2 U, +4 v.W, +1/2 U^2) are correct. Physical magnitude of the 2nd-order correction at Earth's orbit is ~0.1 ns over 115 days, while the observed residual is 116 us detrended - six orders above physics.
 * Third-body isolation (stripping Moon/Mars/Venus to Sun+Earth only) leaves the residual bit-identical, so third-body handling is not the source.
 * Residual rate = 1.2e-11 ~ 0.22 % of the 1st-order integrand v^2/(2c^2) rate. The fingerprint is a small A_1 slippage in the 2nd-order pipeline, not an A_2 formula error.
 * SolarSystemMetric::updateMetric uses Sum_i U_i^2 (sum of individual squares) where the IAU convention calls for (Sum U_i)^2 in g_00; and secondOrderBodyList_ only influences the spatial metric, not g_00. Neither alone explains 116 us but both are worth reviewing.
 * TimeEphemerisWithSecondOrderDirectConversion is a throw-stub, never instantiated - resetIntegratedPostNewtonianTimeEphemeris always creates the 1st-order ephemeris even for 2nd-order propagator settings. Implementing it properly requires plumbing the external scalar/vector potential functions through the constructor and a wiring branch that switches on propagator-settings type.

Incidental note: calculateFirstOrderTcbToTcgDirectCorrection and calculateSecondOrderTcbToTcgDirectCorrection in
relativisticTimeConversion.cpp are unreferenced (dead code) - the live direct correction lives in
TimeEphemerisWith*OrderDirectConversion::calculateDirectTimeDifferenceTerm.

### DIFF
--- a/include/tudat/simulation/propagation_setup/setNumericallyIntegratedStates.h
+++ b/include/tudat/simulation/propagation_setup/setNumericallyIntegratedStates.h
@@ -1254,9 +1254,27 @@ void resetIntegratedPostNewtonianTimeEphemeris(
             std::shared_ptr< ground_stations::GroundStationState > groundStationState =
                     std::dynamic_pointer_cast< simulation_setup::Body >( bodies.getBody( referencePointIdentifier.first ) )->getGroundStation(
                         referencePointIdentifier.second )->getNominalStationState( );
+            // The direct correction in TimeEphemerisWith*OrderDirectConversion dots the central body's
+            // INERTIAL barycentric velocity with this position vector. The IBP identity that underlies
+            // the -v_E.r/c^2 correction requires r in the inertial frame (so that v_rel = dr/dt = omega x r).
+            // Returning the body-fixed nominal position here would silently drop Earth's diurnal rotation
+            // and leave a ~2 us/day periodic residual against direct-from-metric.
+            const auto rotationalEphemeris =
+                    bodies.getBody( referencePointIdentifier.first )->getRotationalEphemeris( );
+            if( rotationalEphemeris == nullptr )
+            {
+                throw std::runtime_error(
+                            "Error when wiring direct-correction station position function: body " +
+                            referencePointIdentifier.first + " has no rotational ephemeris." );
+            }
+            auto stationPositionInInertialFrame = [ groundStationState, rotationalEphemeris ]( const TimeType t ) -> Eigen::Vector3d
+            {
+                return rotationalEphemeris->getRotationToBaseFrame( static_cast< double >( t ) ).toRotationMatrix( )
+                       * groundStationState->getNominalCartesianPosition( );
+            };
             timeEphemeris->resetBodycentricToTopocentricInterpolators(
                         floatingPointValueNumericalSolution, referencePointIdentifier.second,
-                        std::bind( &ground_stations::GroundStationState::getNominalCartesianPosition, groundStationState ) );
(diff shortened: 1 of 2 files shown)


DominicDirkx:
@kimonito98 Thanks for putting the time into this and great to see you found the bug! If I understand correctly, the 1st order model is now fully validated, while the 2nd order one is not. Is this correct?

kimonito98:
@DominicDirkx Yes — exactly that. 1st-order chain: fully validated against direct-from-metric. Strict 1-year test detrended residual 3.89 ns (slope 1.4e-16 s/s, double-precision floor) vs the 6 ns tolerance. No open questions.
2nd-order chain: not validated. With the instrumentation this PR adds, testCombinedProperTimeAndStateDynamics2 now prints detrended amplitude 116 μs / endpoint −622 μs over 115 days — six orders above the physical 2nd-order magnitude (~0.1 ns).

DominicDirkx:
Thanks! Then I'll merge this into the other PR, have the CI run and verify that the 2nd order is the only failure. Then, I suggest me make a modification that an exception gets thrown if the 2nd order propagation is called, effectively disabling it without deleting any of our code for it. We then open an issue to correct this and merge the time rate functionality (with everything except the 2nd order effect working!)!

PR #753, opened 2026-04-26
Feature/light time correction components

### LINKED ISSUE #2: Unable to build boost-python-examples with tudatBundle (branch:develop)
The following information is returned for versioning of g++, make, gcc.

```
g++ --version
g++ (Ubuntu 7.4.0-1ubuntu1~18.04) 7.4.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

make --version
GNU Make 4.1
Built for x86_64-pc-linux-gnu
Copyright (C) 1988-2014 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

gcc --version
gcc (Ubuntu 7.4.0-1ubuntu1~18.04) 7.4.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

When building `/boost-python-examples` in the following way:

```
mkdir build
cd build
cmake -DCMAKE_BUILD_TYPE=DEBUG ..
make
```

The following debugging information is returned for a successful compilation, with all tests passed:

```
-- Boost version: 1.61.0
-- Found the following Boost libraries:
--   python3
-- Found PythonInterp: /usr/bin/python3.6 (found suitable version "3.6.7", minimum required is "3") 
-- PYTHON_LIBRARIES = /usr/lib/x86_64-linux-gnu/libpython3.6m.so
-- PYTHON_EXECUTABLE = /usr/bin/python3.6
-- PYTHON_INCLUDE_DIRS = /usr/include/python3.6m
-- Boost_LIBRARIES = /home/ggarrett/Software/boost_1_61_0/stage/lib/libboost_python3.so
-- Configuring done
-- Generating done
-- Build files have been written to: /home/ggarrett/Repositories/tudatBundle/boost-python-examples/build
+ make
[  6%] Built target hello
[ 13%] Built target classes
[ 20%] Built target ctor
[ 26%] Built target member
[ 33%] Built target inheritance
[ 40%] Built target virtual
[ 46%] Built target operators
[ 53%] Built target policies
[ 60%] Built target overload
[ 66%] Built target mymodule
[ 76%] Built target embedding
[ 83%] Built target iterators
[ 90%] Built target myexceptions
[100%] Built target auto_instance
+ make test
Running tests...
Test project /home/ggarrett/Repositories/tudatBundle/boost-python-examples/build
      Start  1: 01-HelloWorld
 1/13 Test  #1: 01-HelloWorld ....................   Passed    0.03 sec
      Start  2: 02-ExposingClasses
 2/13 Test  #2: 02-ExposingClasses ...............   Passed    0.03 sec
      Start  3: 03-Constructors
 3/13 Test  #3: 03-Constructors ..................   Passed    0.02 sec
      Start  4: 04-ClassMembers
 4/13 Test  #4: 04-ClassMembers ..................   Passed    0.03 sec
      Start  5: 05-Inheritance
 5/13 Test  #5: 05-Inheritance ...................   Passed    0.03 sec
      Start  6: 06-VirtualFunctionsInPython
 6/13 Test  #6: 06-VirtualFunctionsInPython ......   Passed    0.02 sec
      Start  7: 07-Operators
 7/13 Test  #7: 07-Operators .....................   Passed    0.03 sec
      Start  8: 08-CallPolicies
 8/13 Test  #8: 08-CallPolicies ..................   Passed    0.03 sec
      Start  9: 09-Overloading
 9/13 Test  #9: 09-Overloading ...................   Passed    0.03 sec
      Start 10: 10-Embedding
10/13 Test #10: 10-Embedding .....................   Passed    0.02 sec
      Start 11: 11-Iterators
11/13 Test #11: 11-Iterators .....................   Passed    0.03 sec
      Start 12: 12-Exceptions
12/13 Test #12: 12-Exceptions ....................   Passed    0.02 sec
      Start 13: 13-AutoInstantiation
13/13 Test #13: 13-AutoInstantiation .............   Passed    0.02 sec

100% tests passed, 0 tests failed out of 13

Total Test time (real) =   0.34 sec
```

Now trying to obtain the same results from compiling the tudatBundle: **When only adding** `add_subdirectory( "${PROJECTROOT}/boost-python-examples")` from the tudatBundle top level CMakelist.txt, the following is output in debugging:

```
-- Could NOT find Boost
-- Found PythonInterp: /usr/bin/python3.6 (found suitable version "3.6.7", minimum required is "3") 
-- PYTHON_LIBRARIES = /usr/lib/x86_64-linux-gnu/libpython3.6m.so
-- PYTHON_EXECUTABLE = /usr/bin/python3.6
-- PYTHON_INCLUDE_DIRS = /usr/include/python3.6m
-- Boost_LIBRARIES = 
-- << Tudat (DEBUG - part of TudatBundle) >>
```

In order to temporarily fix this, the following temporary fix was added to `/boost-python-examples/CMakelist.txt` (so that CMAKE is aware of my own installation of boost 1.61):

```
set(BOOST_LIBRARYDIR ~/Software/boost_1_61_0/stage/lib/)
set(BOOST_ROOT ~/Software/boost_1_61_0/)
```

This results in the following debugging information:

```
-- Boost version: 1.64.0
-- Found the following Boost libraries:
--   python3
-- Found PythonInterp: /usr/bin/python3.6 (found suitable version "3.6.7", minimum required is "3") 
-- PYTHON_LIBRARIES = /usr/lib/x86_64-linux-gnu/libpython3.6m.so
-- PYTHON_EXECUTABLE = /usr/bin/python3.6
-- PYTHON_INCLUDE_DIRS = /usr/include/python3.6m
-- Boost_LIBRARIES = /home/ggarrett/Software/boost_1_61_0/stage/lib/libboost_python3.a

```

The compilation fails for the tudatBundle build, presumably due to `Boost version: 1.64.0` being detected within the tudatBundle, and `/home/ggarrett/Software/boost_1_61_0` being used from my path variable. The problem is that if the boost is used from the tudatBundle, the compiled python.boost is not visible from a clean non-compiled state. I remember you mentioned something yesterday concerning PyGMO, but I fail to see how this would make the boost.python available and visible at compile time. How do I get around this, would you mind mentioning the setting you changed yesterday @DominicDirkx?

kimonito98:
## Background

In the context of the ACES mission analysis we need to inspect the individual contributions of light-time corrections to simulated observables — in particular to separate **code** and **carrier-only** terms, which can have *opposite signs*. The canonical example is the ionospheric correction: its sign flips between code range (group delay, positive) and carrier phase (phase advance, negative), so for ambiguity resolution (code-carrier) we need them individually. 

Currently `LightTimeCalculator::setTotalLightTimeCorrection` folds every registered correction into a single scalar `currentCorrection_`. Only that cumulative value is exposed, so the only way to figure out what any one correction contributes is to re-simulate the full observable set with each correction toggled on or off in isolation. For more than a handful of correction types this is both slow (N + 1 full simulation passes) and noisy (the converged transmission time shifts slightly with each re-simulation, leaving a second-order residual that never quite vanishes).

This PR makes the per-type breakdown first-class: it is computed and cached along with the cumulative value (zero additional work at simulate-time) and can be saved alongside observables as a dependent variable, filtered by correction type.

### 1. `Add per-correction light-time correction output` (a266ebcf9)

C++ feature.

- **`LightTimeCalculator`** populates a `currentCorrectionComponents_`
  vector (one entry per registered `LightTimeCorrection`) during every
  call to `setTotalLightTimeCorrection`. Cleared on the same iteration
  resets that zero `currentCorrection_`, so after simulation it always
  matches the last converged state. Sum of components ≡ `currentCorrection_`
  by construction.

- New non-templated base class **`LightTimeCalculatorBase`** exposes
  `getCurrentLightTimeCorrectionComponents()` and `getLightTimeCorrectionList()`
  for type-erased access (used by the dependent-variable plumbing and
  Python bindings).

- New **`light_time_correction_components` `ObservationDependentVariable`**
  with a `LightTimeCorrectionComponentsDependentVariableSettings` subclass
  and a `lightTimeCorrectionComponentsDependentVariable(...)` factory.
  Saves one entry per registered correction on the selected
  transmitter → receiver leg, optionally filtered by `LightTimeCorrectionType`
  (e.g. only ionospheric + relativistic for code/carrier sign-difference
  analysis).

- **`ObservationDependentVariableCalculator`** gains a per-leg
  `LightTimeCalculatorBase` map, wired up at simulate-time via a templated
  `extractLegLightTimeCalculators(observationModel)` helper that handles
  one-way range, n-way range, DSN n-way range, and one-way Doppler.
  When a user adds a `light_time_correction_components` setting through
  the usual `addDependentVariablesToObservationSimulationSettings` flow
  (which happens *before* the observation model exists), the
  `Bookkeeping` defers the registration; the simulator-time `Calculator`
  constructor drains those deferred entries once it has the leg map.

- **Drive-by fix #1**: `Calculator::addDependentVariable` was a no-op
  on `develop` (left `dependentVariableAddFunctions_` unpopulated); it
  now actually calls `addDependentVariableFunction`.

- **Drive-by fix #2**: `simulateObservations.h` had four references to
  `observation_models::ObservationDependentVariableCalculator`, but the
  class lives in `simulation_setup`. Misqualifications corrected.

### 2. `Expose per-correction light-time correction output` (40de5481e)

Python bindings.

- `LightTimeCorrectionType` enum exposed on
  `tudatpy.estimation.observable_models_setup.light_time_corrections`
  for use as a filter argument.
- `LightTimeCalculatorBase` exposed with the two getters.
- `light_time_correction_components_dependent_variable(...)` factory on
  `tudatpy.estimation.observations_setup.observations_dependent_variables`.

### 3. `Add unit test for light_time_correction_components dependent variable` (ee973eb15)

`tests/test_tudat/src/astro/observation_models/unitTestLightTimeCorrectionComponents.cpp`,
registered under `test_light_time_correction_components`.

- **`testPerCorrectionComponentsOneWayRange`**: simulates a one-way
  range from an Earth ground station to a Moon orbiter at 10 epochs,
  four times: (a) no corrections, (b) Sun relativistic only,
  (c) Jupiter relativistic only, (d) both corrections + the new
  dependent variable. Per epoch verifies that
  `c × component[i] ≈ range(only_i) − range(none)` and
  `c × Σ component ≈ range(both) − range(none)` to within 1 mm
  (tolerance dominated by the second-order residual from light-time
  iteration re-converging when the correction set changes — well above
  the actual residual at this geometry).
- **`testCorrectionTypeFilter`**: registers two first-order-relativistic
  corrections (Sun + Jupiter perturbers) and confirms that filtering by
  `LightTimeCorrectionType` yields a single entry equal to the first
  registered correction of that type in the unfiltered vector.

Both cases pass locally.

## Scope / limitations

- Angular-position and relative-angular-position observables
  (`ObservationSize == 2`) are intentionally not handled by
  `extractLegLightTimeCalculators` — `dynamic_pointer_cast` across
  incompatible `ObservationSize` template parameters is a compile-time
  error. A size-2-specialised overload can be added when needed.

- The `correction_type_filter` argument matches *by type*; if several
  corrections of the same `LightTimeCorrectionType` are registered on
  the same leg (e.g. two `first_order_relativistic` settings with
  different perturbers), the filter picks the first match. Pass an
  empty filter to get every registered correction.

- All other dependent-variable types are unchanged; only the metadata
  switches gain a case for the new enum value.

## Usage (Python)

```python
from tudatpy.estimation.observable_models_setup import links, light_time_corrections
from tudatpy.estimation.observations_setup import observations_dependent_variables

# Save every registered correction on the one-way leg
dep_var = observations_dependent_variables.light_time_correction_components_dependent_variable(
    transmitter_link_end_type=links.transmitter,
    receiver_link_end_type=links.receiver,
)

# Or filter to a specific subset (e.g. inspect iono + relativistic only,
# for code vs. carrier sign-difference analysis)
dep_var = observations_dependent_variables.light_time_correction_components_dependent_variable(
    transmitter_link_end_type=links.transmitter,
    receiver_link_end_type=links.receiver,
    correction_type_filter=[
        light_time_corrections.LightTimeCorrectionType.first_order_relativistic,
        light_time_corrections.LightTimeCorrectionType.saastamoinen_tropospheric,
        light_time_corrections.LightTimeCorrectionType.jakowski_vtec_ionospheric,
    ],
)

### CODE REVIEW

File: include/tudat/astro/observation_models/lightTimeSolution.h [hunk 1]
```diff
@@ -669,6 +692,29 @@ class LightTimeCalculator
...
 
+    //! Function to get the values of the individual light-time corrections that were summed into
+    //! `currentCorrection_` during the last call to `setTotalLightTimeCorrection`. Order matches
+    //! `correctionFunctions_` / `getLightTimeCorrection()`.
+    std::vector< ObservationScalarType > getCurrentLightTimeCorrectionComponentsTyped( ) const
+    {
+        return currentCorrectionComponents_;
+    }
+
+    std::vector< double > getCurrentLightTimeCorrectionComponents( ) const override
```
DominicDirkx: Is there not some fancy C++ STL function that does this?

include/tudat/astro/observation_models/lightTimeSolution.h
```diff
@@ -793,11 +843,14 @@ class LightTimeCalculator
             linkEndStatesDouble.at( i ) = linkEndStates.at( i ).template cast< double >( );
         }
 
+        currentCorrectionComponents_.resize( correctionFunctions_.size( ) );
         for( unsigned int i = 0; i < correctionFunctions_.size( ); i++ )
         {
-            totalLightTimeCorrections +=
-                    static_cast< ObservationScalarType >( correctionFunctions_[ i ]->calculateLightTimeCorrectionWithMultiLegLinkEndStates(
+            const ObservationScalarType singleCorrection = static_cast< ObservationScalarType >(
```
DominicDirkx: What is the point of upcasting them to ObservationScalarType as separate contributions? The calculations are only in double precision

include/tudat/simulation/estimation_setup/simulateObservations.h
```diff
@@ -33,6 +37,82 @@ namespace tudat
...
 
+//! Extract, from an observation model, the LightTimeCalculator associated with each leg of the
+//! underlying light-time computation, keyed by (transmitter link-end type, receiver link-end type).
+//!
+//! Used by `ObservationDependentVariableCalculator` to resolve the
+//! `light_time_correction_components` dependent variable. Currently supports: one-way range,
+//! n-way range, DSN n-way range, and one-way Doppler (ObservationSize = 1). For unsupported
+//! observation types the returned map is empty, and requesting `light_time_correction_components`
+//! for such a type yields a clear error at setup.
+template< int ObservationSize, typename ObservationScalarType, typename TimeType >
```
DominicDirkx: I think this can/should be generalized. We can make a function that checks whether an observation model has:
* single-leg light-time
* multi-leg light-time
* no light time
And then extracts the light-time calculator for all observable types

src/tudatpy/estimation/observable_models_setup/light_time_corrections/expose_light_time_corrections.cpp
```diff
@@ -42,6 +42,40 @@ namespace light_time_corrections
 
 void expose_light_time_corrections( py::module& m )
 {
+    py::enum_< tom::LightTimeCorrectionType >(
+            m,
+            "LightTimeCorrectionType",
+            R"doc(Enum identifying each type of light-time correction registered on a link.
+
+Used as a filter by :func:`~tudatpy.estimation.observations_setup.observations_dependent_variables.light_time_correction_components_dependent_variable`
+to select which correction contributions are saved individually.)doc" )
```
DominicDirkx, line 51: quite a list we have now :)

DominicDirkx:
@kimonito98 I'm playing around with addressing the comments in this PR (have a look if you disagree with any of them)

include/tudat/simulation/estimation_setup/observationOutput.h
```diff
@@ -132,26 +162,52 @@ class ObservationDependentVariableBookkeeping
...
                     dependentVariableBookkeeping_->getDependentVariableSettings( ).at( i ) );
             addDependentVariableFunction(
                     dependentVariableBookkeeping_->getDependentVariableSettings( ).at( i ), bodies, indices.first, indices.second );
         }
+
+        // Then, drain any deferred `light_time_correction_components` settings. They were added
+        // before the leg-LightTimeCalculator map was known; now we have it, so we can register
+        // them properly (this updates both the bookkeeping and the add-function list).
+        if( !legLightTimeCalculators_.empty( ) )
+        {
```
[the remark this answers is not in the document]
kimonito98: Fixed in 97f09d55e: the constructor now drains bookkeeping->takeDeferredSettings() unconditionally and lets registerLightTimeCorrectionComponents raise the existing "no LightTimeCalculator registered for leg ..." error per entry. Unsupported observables (and unsupported leg pairs on supported observables) therefore fail loudly when the simulator builds the calculator, instead of producing an empty dependent-variable vector. New test testUnsupportedObservableRaises exercises the path on angular_position.

include/tudat/simulation/estimation_setup/observationOutput.h
```diff
@@ -120,6 +128,28 @@ class ObservationDependentVariableBookkeeping
         totalDependentVariableSize_ = 0;
     }
 
+    //! Store a setting whose size depends on state not yet available (e.g.
+    //! `light_time_correction_components` with an empty type-filter — its size equals the
+    //! number of registered light-time corrections, which is only known once the observation
+    //! model has been built). `ObservationDependentVariableCalculator` flushes this list and
+    //! registers the settings properly once `setLegLightTimeCalculators` is called.
+    void addDeferredSetting( const std::shared_ptr< ObservationDependentVariableSettings > setting )
+    {
```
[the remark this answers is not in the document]
kimonito98: Fixed in 97f09d55e. Bookkeeping::addDeferredSetting now skips an incoming setting if any entry in settingsList_ or deferredSettings_ reports areSettingsCompatible(...) true, mirroring the dedup that Bookkeeping::addDependentVariables already does for non-deferred entries. So repeated addDependentVariablesToObservationSimulationSettings calls before simulate-time can no longer produce duplicate slots.

src/tudatpy/estimation/observations_setup/observations_dependent_variables/expose_observations_dependent_variables.cpp
```diff
@@ -499,6 +500,44 @@ void expose_observations_dependent_variables( py::module& m )
...
+        Function to create a dependent variable that saves each light-time correction contribution
+        individually for a single leg (transmitter → receiver) of the observable.
+
+        The returned vector contains one entry per registered :class:`~tudatpy.estimation.observable_models_setup.light_time_corrections.LightTimeCorrection`
+        on the selected leg, in registration order. If ``correction_type_filter`` is supplied,
+        only contributions whose type appears in the filter are returned, in the order of the filter.
+
+        Supported for observables whose light-time computation is extracted by the simulator
+        (one-way range, n-way range, DSN n-way range, one-way Doppler). Unsupported observables
+        produce a clear error at setup time.
```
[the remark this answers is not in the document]
kimonito98: Resolved together with first comment (97f09d55e). With the unconditional drain the implementation now matches what the docstring claimed; I also tightened the wording to be precise about when the error fires:

Requesting this dependent variable for an unsupported observable, or for a transmitter/receiver pair that does not correspond to an actual leg of the chosen observable, raises a RuntimeError when the observation is simulated (i.e. on the first call to simulate_observations for that ObservationSimulationSettings), not silently dropped.

include/tudat/astro/observation_models/lightTimeSolution.h
kimonito98: setTotalLightTimeCorrection now uses std::transform to fill currentCorrectionComponents_ and std::accumulate to widen each cached double to ObservationScalarType while summing into currentCorrection_

include/tudat/astro/observation_models/lightTimeSolution.h
kimonito98: You're right that it was gratuitous. LightTimeCorrection::calculateLightTimeCorrectionWithMultiLegLinkEndStates already returns double, so widening every component on the way in cost a copy and bought nothing in precision. Fixed in 97f09d55e: currentCorrectionComponents_ is now std::vector<double>, the public getCurrentLightTimeCorrectionComponents() returns it directly (no per-element cast), and the only widening to ObservationScalarType happens once, in the accumulator that builds currentCorrection_. The redundant getCurrentLightTimeCorrectionComponentsTyped is dropped.

include/tudat/simulation/estimation_setup/simulateObservations.h
kimonito98: extractLegLightTimeCalculators now categorises observation models into two shapes — single-leg (one LightTimeCalculator from transmitter to receiver) and multi-leg (a chain of N − 1 calculators behind a MultiLegLightTimeCalculator) — handled by two helpers tryExtractSingleLegLightTimeCalculator and tryExtractMultiLegLightTimeCalculators that take a model type plus a small accessor lambda

DominicDirkx:
@kimonito98 Two comments on the functionality:
* I'd propose that, if a light_time_correction_components dependent variables is requested, the saved dependent variable is a list of *all* light time corrections that match (I think in resolveLightTimeCorrectionSourceIndices)
* I'd propose removing the ObservationDependentVariableCalculator::setLegLightTimeCalculators function. The ObservationDependentVariableCalculator is created at a point where the observation models and light time calculators are always available, and it would be cleaner and safer to enforce that the LegLightTimeCalculators are passed to the object through the constructor

I also have an idea for how to make the leg light time calculator extraction from the observation model more generic, to allow it to be used for *all* obervation models without much extra code.

If you agree with all of this, a thumbs up is enough, and I'll implement it (I have a good chunk of it already). If not, let's discuss :)

PR #755, opened 2026-04-27
Add proper_time_rate_{kinematic,potential}_term dependent variables

kimonito98:
Expose the kinematic (special-relativistic, second-order Doppler) contribution -v^2/(2 c^2) and the potential (gravitational redshift) contribution -U_ext/c^2 of the proper-time-rate integrand as first-class dependent variables that can be saved at every integrator step. Both terms are computed from quantities the relativistic-time state derivative already caches; no extra integration work is added on the hot path.

Two pipelines are wired up:
 * Post-Newtonian chain (FirstOrderBarycentricToBodyCentricTimeStateDerivative and its second-order subclass): values come from the cached currentVelocity_ and currentExternalPotential_ on the state derivative, which are refreshed in updateStateDerivativeModel each step. Two new public inline getters expose them.
 * Direct-from-metric (DirectProperTimeRateStateDerivative): the kinematic term is reconstructed from the bound reference-point state function (BCRS velocity), and the potential term is read from the SolarSystemMetric that the state derivative drives via a new getReferencePointStateFunction() accessor and the existing SolarSystemMetric::getCurrentScalarPotential. Other Metric subclasses (e.g. Schwarzschild) currently throw an explicit error when used with the potential variant; a separate dependent variable would be required to expose them cleanly.

Library wiring:
 * propagationOutputSettings.h: new enum entries proper_time_rate_kinematic_term=79 and proper_time_rate_potential_term=80, plus inline builders properTimeRateKinematicTermDependentVariable(body_name) and properTimeRatePotentialTermDependentVariable(body_name).
 * propagationOutput.h: dispatch case in getDoubleDependentVariableFunction that locates the relativistic-time state derivative for the requested body and produces the corresponding closure.
 * propagationOutput.cpp: getDependentVariableSize returns 1 for both.
 * propagationOutputSettings.cpp: getDependentVariableName provides human-readable strings.
 * createEnvironmentUpdater.cpp: explicit no-op cases (values come from the state derivative, which is already updated by the propagator).

Python:
 * expose_dependent_variable.cpp: m.def("proper_time_rate_kinematic_term", ...) and m.def("proper_time_rate_potential_term", ...) wrap the C++ builders with Numpy-style docstrings.

Test:
 * tests/.../unitTestRelativisticTimePropagation.cpp::testProperTimeRateGrSrSplitDependentVariables propagates a first-order TCB->TCG state for 5 days with both variables saved, then reconstructs them independently from the body ephemerides at every saved epoch. Maximum recovery error 8.27e-25 s/s (rate units) - numerical noise floor.

Notes for review / follow-ups:
 * Direct-from-metric potential term requires a SolarSystemMetric; adding a virtual getCurrentScalarPotential to the Metric base would let other metrics participate without dynamic_cast.
 * The kinematic + potential split is well-defined only at first order in 1/c^2. The 1/c^4 corrections that the second-order TCB->TCG integrand and the direct-from-metric series add are not split; users wanting a clean total-rate value should also use the existing state history.

### DIFF
--- a/include/tudat/astro/propagators/relativisticTimeStateDerivative.h
+++ b/include/tudat/astro/propagators/relativisticTimeStateDerivative.h
@@ -273,6 +273,17 @@ public:
         return currentProperTimeDerivative_;
     }
 
+    //! Function returning the reference-point Cartesian state used at each evaluation.
+    /*!
+     *  Wraps the function bound at construction (either the body's own state or, for a ground
+     *  station reference, the topocentric state). Used by GR/SR-split dependent variables to
+     *  read out the BCRS velocity that drives the kinematic time-dilation term.
+     */
+    std::function< Eigen::Vector6d( ) > getReferencePointStateFunction( ) const
+    {
+        return referencePointStateFunction_;
+    }
+
 
 protected:
     std::shared_ptr< relativity::Metric > spaceTimeMetric_;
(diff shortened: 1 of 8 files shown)


DominicDirkx:
Reopened in #757

PR #757, opened 2026-04-28
Add proper_time_rate_{kinematic,potential}_term dependent variables

DominicDirkx:
Expose the kinematic (special-relativistic, second-order Doppler) contribution -v^2/(2 c^2) and the potential (gravitational redshift) contribution -U_ext/c^2 of the proper-time-rate integrand as first-class dependent variables that can be saved at every integrator step. Both terms are computed from quantities the relativistic-time state derivative already caches; no extra integration work is added on the hot path.

Two pipelines are wired up:
 * Post-Newtonian chain (FirstOrderBarycentricToBodyCentricTimeStateDerivative and its second-order subclass): values come from the cached currentVelocity_ and currentExternalPotential_ on the state derivative, which are refreshed in updateStateDerivativeModel each step. Two new public inline getters expose them.
 * Direct-from-metric (DirectProperTimeRateStateDerivative): the kinematic term is reconstructed from the bound reference-point state function (BCRS velocity), and the potential term is read from the SolarSystemMetric that the state derivative drives via a new getReferencePointStateFunction() accessor and the existing SolarSystemMetric::getCurrentScalarPotential. Other Metric subclasses (e.g. Schwarzschild) currently throw an explicit error when used with the potential variant; a separate dependent variable would be required to expose them cleanly.

Library wiring:
 * propagationOutputSettings.h: new enum entries proper_time_rate_kinematic_term=79 and proper_time_rate_potential_term=80, plus inline builders properTimeRateKinematicTermDependentVariable(body_name) and properTimeRatePotentialTermDependentVariable(body_name).
 * propagationOutput.h: dispatch case in getDoubleDependentVariableFunction that locates the relativistic-time state derivative for the requested body and produces the corresponding closure.
 * propagationOutput.cpp: getDependentVariableSize returns 1 for both.
 * propagationOutputSettings.cpp: getDependentVariableName provides human-readable strings.
 * createEnvironmentUpdater.cpp: explicit no-op cases (values come from the state derivative, which is already updated by the propagator).

Python:
 * expose_dependent_variable.cpp: m.def("proper_time_rate_kinematic_term", ...) and m.def("proper_time_rate_potential_term", ...) wrap the C++ builders with Numpy-style docstrings.

Test:
 * tests/.../unitTestRelativisticTimePropagation.cpp::testProperTimeRateGrSrSplitDependentVariables propagates a first-order TCB->TCG state for 5 days with both variables saved, then reconstructs them independently from the body ephemerides at every saved epoch. Maximum recovery error 8.27e-25 s/s (rate units) - numerical noise floor.

Notes for review / follow-ups:
 * Direct-from-metric potential term requires a SolarSystemMetric; adding a virtual getCurrentScalarPotential to the Metric base would let other metrics participate without dynamic_cast.
 * The kinematic + potential split is well-defined only at first order in 1/c^2. The 1/c^4 corrections that the second-order TCB->TCG integrand and the direct-from-metric series add are not split; users wanting a clean total-rate value should also use the existing state history.

### DIFF
--- a/include/tudat/simulation/propagation_setup/propagationOutputSettings.h
+++ b/include/tudat/simulation/propagation_setup/propagationOutputSettings.h
@@ -1396,27 +1396,64 @@ inline std::shared_ptr< SingleDependentVariableSaveSettings > actualCrossSection
             actual_cross_section, bodyName, centralBodyName, accelerationType );
 }
 
-//! Build a SingleDependentVariableSaveSettings for the kinematic (special-relativistic, second-order
-//! Doppler) term \f$-v^{2} / (2c^{2})\f$ of the proper-time-rate integrand at the central body of a
-//! relativistic time propagation. Supported pipelines: post-Newtonian (first- and second-order
-//! TCB->TCG) and direct-from-metric.
+//! Build a SingleDependentVariableSaveSettings for the kinematic (special-relativistic,
+//! second-order Doppler) contribution to the proper-time-rate integrand of an observer.
+//!
+//! The observer is the reference point of one of the relativistic-time propagator settings:
+//!   - \ref FirstOrderBodycentricRelativisticTimePropagatorSettings (and its second-order
+//!     subclass) for a body centre, where the integrand follows Soffel et al. 2003 Eq. (58):
+//!     \f$ d\Delta_{BC}/dt_B = -(v_C^2/2 + w_{0,\mathrm{ext}})/c^2 \f$. This dependent variable
+//!     returns the \f$-v_C^2/(2c^2)\f$ contribution.
+//!   - \ref DirectRelativisticTimePropagatorSettings, where the integrand is the series
+//!     expansion \f$ d\tau/dt - 1 = -\varepsilon/2 - \varepsilon^2/8 \f$ with
+//!     \f$\varepsilon = (u^\mu h_{\mu\nu} u^\nu + v^2)/c^2\f$. This dependent variable returns
+//!     the leading kinematic part \f$-v^2/(2c^2)\f$ where \f$v\f$ is the reference-point BCRS
+//!     speed.
+//!   - \ref BodycenteredToTopocentricTimePropagatorSettings for a ground station, where the
+//!     integrand follows Turyshev et al. 2013 Eq. (22). This dependent variable returns the
+//!     \f$-v_0^2/(2c^2)\f$ contribution where \f$v_0\f$ is the reference-point velocity in
+//!     the body-centred frame.
+//!
+//! \param bodyName Body whose relativistic-time state is being propagated.
+//! \param referencePoint Optional ground-station / topocentric reference point on the body.
+//!        Leave empty (default) for body-centre proper time (TCG-like conversions).
 //! @get_docstring(properTimeRateKinematicTermDependentVariable)
 inline std::shared_ptr< SingleDependentVariableSaveSettings > properTimeRateKinematicTermDependentVariable(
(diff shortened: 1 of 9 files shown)


include/tudat/simulation/propagation_setup/propagationOutputSettings.h
```diff
@@ -1394,6 +1396,29 @@ inline std::shared_ptr< SingleDependentVariableSaveSettings > actualCrossSection
...
 
+//! Build a SingleDependentVariableSaveSettings for the kinematic (special-relativistic, second-order
+//! Doppler) term \f$-v^{2} / (2c^{2})\f$ of the proper-time-rate integrand at the central body of a
+//! relativistic time propagation. Supported pipelines: post-Newtonian (first- and second-order
+//! TCB->TCG) and direct-from-metric.
+//! @get_docstring(properTimeRateKinematicTermDependentVariable)
+inline std::shared_ptr< SingleDependentVariableSaveSettings > properTimeRateKinematicTermDependentVariable(
+        const std::string& bodyName )
+{
+    return std::make_shared< SingleDependentVariableSaveSettings >( proper_time_rate_kinematic_term, bodyName );
```
DominicDirkx: Should this not take the body name and reference point as input? The dependent variable will be specific to the proper time rate of an observer (which can be a body center for a TCG-TCB like conversion, in which case this will work fine) which should include the ground station definition

src/tudat/simulation/propagation_setup/createEnvironmentUpdater.cpp
```diff
@@ -1313,6 +1313,12 @@ std::map< propagators::EnvironmentModelsToUpdate, std::vector< std::string > > c
             break;
         case aerodynamic_coefficients:
             break;
+        case proper_time_rate_kinematic_term:
+        case proper_time_rate_potential_term:
+            // No additional environment update needed: the values come straight from the
```
DominicDirkx: No need to add the comment, it's not there in other updates

src/tudatpy/dynamics/propagation_setup/dependent_variable/expose_dependent_variable.cpp
```diff
@@ -2592,6 +2592,45 @@ The type of the acceleration that is to be saved.
            py::arg( "central_body_name" ),
            py::arg( "acceleration_type" ) = "radiation_pressure",
            R"doc(No documentation found.)doc" );
+
+    m.def( "proper_time_rate_kinematic_term",
+           &tp::properTimeRateKinematicTermDependentVariable,
+           py::arg( "body_name" ),
+           R"doc(Function to retrieve the kinematic (special-relativistic) term -v^2/(2 c^2) of
+the proper-time-rate integrand for a body that is being propagated as a relativistic
+time state. Works for post-Newtonian (first- or second-order) and direct-from-metric
```
DominicDirkx: This docs (and the next one) should have an explicit reference to the state derivative model from which it extracts the specific term, and be explicit how it relates to that model's governing equation

tests/test_tudat/src/astro/propagators/unitTestRelativisticTimePropagation.cpp
```diff
@@ -231,6 +231,118 @@ BOOST_AUTO_TEST_CASE( testRelativisticTimePropagationPlaceholder )
     BOOST_CHECK( true );
 }
 
+// Smoke test for the new GR/SR split dependent variables introduced in this
```
DominicDirkx: What is a smoke test?

PR #758, opened 2026-04-29
(draft): Add esa-neocc api + proposed universal observation collection adapter concept 

luigigisolfi:
Added esaneocc API support (for observational ephemeris only) + created new proposed observations adapter acting on any observation collection. 

@DominicDirkx @larshinueber if we like this, I can work on it. Try out the related example [here](https://github.com/tudat-team/tudatpy-examples/pull/141).


DominicDirkx:
I will close this PR for now, since much of it has been superseded (and in fact, motivated and inspired!) the major refactor on the data module we did. I think having an interface to teh NEOCC directly would be great! This would live under (in the new setup) data_input.environment_data.neocc (and we should then also add the neocc package as dependency)

Issue #669, opened 2026-02-17
[F&E]: Serialization Main Thread


PascalChristiaanse:
### Description:

Hi all, Recently i opened #652 and have since brought up the topic of serialization in the monthly development meeting. To gauge where efforts should be directed, im interested in use cases and priorities. The current implementation might be somewhat limited as i have little time available to work on this; its meant as a stepping stone for future development / something to figure out effective workflows on the user side.

Current use-cases we've discussed:
- Multiprocessing or Interprocess communication in general
- Archiving results
- Being able to save and archive settings objects for reproducibility / archiving
- Saving intermediate results for complicated analyses
- Gateway to support other fileformats (such as HDF5 for true portability)

Some complications have shown which i will make a seperate comment for once ive collected my thoughts on this; 
- Function pointers cannot be serialized
- custom functions cannot be serialized
This leads some interesting cases where some objects _can_ and some _cannot_ be fully reconstructed just based on the serialized data. 

Please leave your opinions and expected workflows below. For example; would you care if a certain object would be *non-functional* and *only* contain _data_? What objects do you think this wouldn't be an issue for? etc. 

### Suggestions for implementation:

Serialization using cereal
 allows use-cases like above, and possibly many more

Boost::serialize appears to be less feature rich than cereal, thanks for the suggestion Lars Hinuber!

### Additional context:

Ill take this matter up at some point with dominic and possibly eScience Center.

PascalChristiaanse:
@valeriof7

jo11he:
@PascalChristiaanse I learned that @valeriof7 already collected some information on this to facilitate a task kick-off with our colleagues at the E-Science Centre, maybe that material can go here too?

PascalChristiaanse:
I see youve found the thread, perfect.

That sounds great, I think its a good idea to keep everything collected in one spot, so lets put it here!

HannoSpreeuw:
> * Multiprocessing or Interprocess communication in general

Please clarify the use case, are you launching multiple instances of Tudatpy, i.e. multiple instances of the Python interpreter, or are you creating multiple processes in C++ and try to communicate between them?
In the first case, the "file storage solution", see below, is preferred.

> * Archiving results
> 
> * Being able to save and archive settings objects for reproducibility / archiving
> 
> * Saving intermediate results for complicated analyses
> 
> * Gateway to support other fileformats (such as HDF5 for true portability)

[...]

PR #652, opened 2026-02-09
Feature/serialization using boost

PascalChristiaanse:
Hi all, Ive taken some time to look into serialization and ive found a workflow that i think is interesting.

Boost offers serialization and while it is invasive in all classes we would want to serialize, the amount is VERY minimal in my opinion. For basic classes it is straight up just boiler plate as shown below 

```
class MyBase {
public:
    virtual ~MyBase() = default;
    // ... public interface ...

protected:
    MyBase() : someField_(defaultValue) {}  // default ctor for deserialization

private:
    friend class boost::serialization::access;

    template< class Archive >
    void serialize( Archive& ar, const unsigned int version )
    {
        (void)version;
        ar & someField_;
    }
};
```

Polymorphic classes are slightly more involved, but also not that challenging. 

So far ive only implemented serialization for the ObservationCollection class and its constituent parts, making up around 10 classes or so. 

Serializable so far are 
* Time
* LinkendId
* LinkDefinition
* ObservationAncilliarySimulationSettings
* ObservationDependentVariableSettings
* StationAngleObservationDependentVariableSettings
* InterlinkObservationDependentVariableSettings
* AncillaryObservationDependentVariableSettings
* ObservationDependentVariableCalculator
* SingleObservationSet
* ObservationCollection

and pickleable so far are:
* Time
* LinkEndId
* LinkDefinition
* ObservationAncilliarySimulationSettings
* ObservationDependentVariableSettings
* SingleObservationSet
* ObservationCollection

I want to eventually do SystemOfBodies, but ill do the results objects first as i think these are super nice to be able to save.

Let me know what you think!

PR #763, opened 2026-05-05
Feature/serialization

PascalChristiaanse:
addresses #669 

contains a lovely claude script to check progress; though all tests still need to be audited properly. 
(the script double counts because of the basic regex based search, but shouldnt miss anything; although thats also not tested so should stay somewhat vigilant)

==================================================
  6 tested / 43 untested / 49 total
  12 with pickle / 37 without pickle
==================================================

❌ NO SERIALIZE TEST FOUND:
   AccelerationSettings                     ✗ no pickle    
   AncillaryObservationDependentVariableSettings ✗ no pickle    
   BodyMassStateDerivative                  ✗ no pickle    
   CovarianceAnalysisOutput                 ✓ pickle       
   CustomAccelerationSettings               ✗ no pickle    
   CustomTorqueSettings                     ✗ no pickle    
   DirectTidalDissipationAccelerationSettings ✗ no pickle    
   EmpiricalAccelerationSettings            ✗ no pickle    
   EstimationOutput                         ✓ pickle       
   FixedTimeHodographicShapingOptimisationProblem ✗ no pickle    
   HodographicShapingOptimisationProblem    ✗ no pickle    
   InterlinkObservationDependentVariableSettings ✗ no pickle    
   LinkDefinition                           ✓ pickle       
   LinkEndId                                ✓ pickle       
   MomentumWheelDesaturationAccelerationSettings ✗ no pickle    
   MutualSphericalHarmonicAccelerationSettings ✗ no pickle    
   NonSequentialPropagationTerminationSettings ✗ no pickle    
   ObservationAncillarySimulationSettings   ✓ pickle       
   ObservationDependentVariableBookkeeping  ✓ pickle       
   ObservationDependentVariableSettings     ✓ pickle       
   PropagationCPUTimeTerminationSettings    ✗ no pickle    
   PropagationCustomTerminationSettings     ✗ no pickle    
   PropagationDependentVariableTerminationSettings ✗ no pickle    
   PropagationHybridTerminationSettings     ✗ no pickle    
   PropagationTerminationDetails            ✗ no pickle    
   PropagationTerminationDetailsFromHybridCondition ✗ no pickle    
   PropagationTerminationSettings           ✗ no pickle    
   PropagationTimeTerminationSettings       ✗ no pickle    
   PropagatorType                           ✗ no pickle    
   RTGAccelerationSettings                  ✗ no pickle    
   RadiationPressureAccelerationSettings    ✗ no pickle    
   RelativisticAccelerationCorrectionSettings ✗ no pickle    
   RootFinderSettings                       ✗ no pickle    
   SimulationResults                        ✓ pickle       
   SingleDependentVariableSaveSettings      ✗ no pickle    
   SphericalHarmonicAccelerationSettings    ✗ no pickle    
   SphericalHarmonicTorqueSettings          ✗ no pickle    
   StationAngleObservationDependentVariableSettings ✗ no pickle    
   ThrustAccelerationSettings               ✗ no pickle    
   Time                                     ✓ pickle       
   TorqueSettings                           ✗ no pickle    
   YarkovskyAccelerationSettings            ✗ no pickle    
   type_caster                              ✓ pickle       

none have been audited yet, but its a start
✅ SERIALIZE TEST EXISTS: 
   HybridArcSimulationResults               ✗ no pickle    
   MultiArcSimulationResults                ✗ no pickle    
   ObservationCollection                    ✓ pickle       
   SingleArcSimulationResults               ✗ no pickle    
   SingleArcVariationalSimulationResults    ✗ no pickle    
   SingleObservationSet                     ✓ pickle       
   SingleObservationSet                     ✓ pickle


PascalChristiaanse:
Probably going to change the testing strategy to consolidate all serialization into a single unit test, as then i can use harnesses and stuff.

Each serializable class should also implement an equality operator with itself (templating raises some interesting questions); such that the testing logic can use 
'''assert_equal(original, deserialized)'''
Tests should not dictate what equality means, classes should.

single testfile will also reduce the number of binaries that would be included if we did each test in a separate file (already at 45 classes serialized, this would increase the number of tests by 20 ish percent alone)

PascalChristiaanse:
Warnings should be emitted if object contains non-serializable members such as function pointers if they cannot be reconstructed upon deserialization
affected classes are:
Custom<thing>Settings
possibly others

PascalChristiaanse:
I think this is a decent setup

each serializable class must implement
operator==()
serialize()

if its a polymorphic derivation it must implement
serialize()
equals()
where the base must implement 
operator== using equals from the derived classes

to pickle, two helpers are provided:
make_pickle for non polymorphic
make_pickle_polymorphic for polymorphic cases

cereal requires polymorphic relations to be registered, which we can use to pickle more easily. 

status right now:

==================================================
  17 tested / 31 untested / 48 total
  21 with pickle / 27 without pickle
==================================================

❌ NO SERIALIZE TEST FOUND:
   AncillaryObservationDependentVariableSettings ✗ no pickle     (defined in: include/tudat/simulation/estimation_setup/observationOutputSettings.h)
   BodyMassStateDerivative                  ✗ no pickle     (defined in: include/tudat/astro/propagators/bodyMassStateDerivative.h)
   CovarianceAnalysisOutput                 ✓ pickle        (defined in: include/tudat/astro/orbit_determination/podInputOutputTypes.h)
   CustomTorqueSettings                     ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/torqueSettings.h)
   EstimationOutput                         ✓ pickle        (defined in: include/tudat/astro/orbit_determination/podInputOutputTypes.h)
   FixedTimeHodographicShapingOptimisationProblem ✗ no pickle     (defined in: include/tudat/astro/low_thrust/shape_based/hodographicShapingOptimisationSetup.h)
   HodographicShapingOptimisationProblem    ✗ no pickle     (defined in: include/tudat/astro/low_thrust/shape_based/hodographicShapingOptimisationSetup.h)
   InterlinkObservationDependentVariableSettings ✗ no pickle     (defined in: include/tudat/simulation/estimation_setup/observationOutputSettings.h)
   LinkDefinition                           ✓ pickle        (defined in: include/tudat/astro/observation_models/linkTypeDefs.h)
   LinkEndId                                ✓ pickle        (defined in: include/tudat/astro/observation_models/linkTypeDefs.h)
   NonSequentialPropagationTerminationSettings ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/propagationTerminationSettings.h)
   ObservationAncillarySimulationSettings   ✓ pickle        (defined in: include/tudat/astro/observation_models/observationAncillarySettings.h)
   ObservationDependentVariableBookkeeping  ✓ pickle        (defined in: include/tudat/simulation/estimation_setup/observationOutput.h)
   ObservationDependentVariableSettings     ✓ pickle        (defined in: include/tudat/simulation/estimation_setup/observationOutputSettings.h)
   PropagationCPUTimeTerminationSettings    ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/propagationTerminationSettings.h)
   PropagationCustomTerminationSettings     ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/propagationTerminationSettings.h)
   PropagationDependentVariableTerminationSettings ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/propagationTerminationSettings.h)
   PropagationHybridTerminationSettings     ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/propagationTerminationSettings.h)
   PropagationTerminationDetails            ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/propagationTermination.h)
   PropagationTerminationDetailsFromHybridCondition ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/propagationTermination.h)
   PropagationTerminationSettings           ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/propagationTerminationSettings.h)
   PropagationTimeTerminationSettings       ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/propagationTerminationSettings.h)
   PropagatorType                           ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/propagationSettings.h)
   RootFinderSettings                       ✗ no pickle     (defined in: include/tudat/math/root_finders/createRootFinder.h)
   SimulationResults                        ✓ pickle        (defined in: include/tudat/simulation/propagation_setup/propagationResults.h)
   SingleDependentVariableSaveSettings      ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/propagationOutputSettings.h)
   SphericalHarmonicTorqueSettings          ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/torqueSettings.h)
   StationAngleObservationDependentVariableSettings ✗ no pickle     (defined in: include/tudat/simulation/estimation_setup/observationOutputSettings.h)
   Time                                     ✓ pickle        (defined in: include/tudat/basics/timeType.h)
   TorqueSettings                           ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/torqueSettings.h)
   type_caster                              ✓ pickle        (defined in: src/tudatpy/astro/time_representation/expose_time_representation.cpp)

✅ SERIALIZE TEST EXISTS:
   CustomAccelerationSettings               ✓ pickle        → tests/test_tudat/src/io/unitTestSettingsSerialization.cpp
   DirectTidalDissipationAccelerationSettings ✓ pickle        → tests/test_tudat/src/io/unitTestSettingsSerialization.cpp
   EmpiricalAccelerationSettings            ✓ pickle        → tests/test_tudat/src/io/unitTestSettingsSerialization.cpp
   HybridArcSimulationResults               ✗ no pickle     → tests/test_tudat/src/io/unitTestPropagationResultsSerialization.cpp
   MomentumWheelDesaturationAccelerationSettings ✓ pickle        → tests/test_tudat/src/io/unitTestSettingsSerialization.cpp
   MultiArcSimulationResults                ✗ no pickle     → tests/test_tudat/src/io/unitTestPropagationResultsSerialization.cpp
   MutualSphericalHarmonicAccelerationSettings ✓ pickle        → tests/test_tudat/src/io/unitTestSettingsSerialization.cpp
   ObservationCollection                    ✓ pickle        → tests/test_tudat/src/io/unitTestObservationCollectionSerialization.cpp
   RTGAccelerationSettings                  ✓ pickle        → tests/test_tudat/src/io/unitTestSettingsSerialization.cpp
   RadiationPressureAccelerationSettings    ✗ no pickle     → tests/test_tudat/src/io/unitTestSettingsSerialization.cpp
   RelativisticAccelerationCorrectionSettings ✓ pickle        → tests/test_tudat/src/io/unitTestSettingsSerialization.cpp
   SingleArcSimulationResults               ✗ no pickle     → tests/test_tudat/src/io/unitTestPropagationResultsSerialization.cpp
   SingleArcVariationalSimulationResults    ✗ no pickle     → tests/test_tudat/src/io/unitTestPropagationResultsSerialization.cpp
   SingleObservationSet                     ✓ pickle        → tests/test_tudat/src/io/unitTestObservationCollectionSerialization.cpp
   SphericalHarmonicAccelerationSettings    ✓ pickle        → tests/test_tudat/src/io/unitTestSettingsSerialization.cpp
   ThrustAccelerationSettings               ✓ pickle        → tests/test_tudat/src/io/unitTestSettingsSerialization.cpp
   YarkovskyAccelerationSettings            ✗ no pickle     → tests/test_tudat/src/io/unitTestSettingsSerialization.cpp

PascalChristiaanse:
==================================================
  14 tested / 30 untested / 44 total
  25 with pickle / 19 without pickle
==================================================

❌ NO SERIALIZE TEST FOUND:
   AncillaryObservationDependentVariableSettings ✗ no pickle     (defined in: include/tudat/simulation/estimation_setup/observationOutputSettings.h)
   BodyMassStateDerivative                  ✗ no pickle     (defined in: include/tudat/astro/propagators/bodyMassStateDerivative.h)
   CovarianceAnalysisOutput                 ✓ pickle        (defined in: include/tudat/astro/orbit_determination/podInputOutputTypes.h)
   CustomTorqueSettings                     ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/torqueSettings.h)
   EstimationOutput                         ✓ pickle        (defined in: include/tudat/astro/orbit_determination/podInputOutputTypes.h)
   FixedTimeHodographicShapingOptimisationProblem ✗ no pickle     (defined in: include/tudat/astro/low_thrust/shape_based/hodographicShapingOptimisationSetup.h)
   HodographicShapingOptimisationProblem    ✗ no pickle     (defined in: include/tudat/astro/low_thrust/shape_based/hodographicShapingOptimisationSetup.h)
   InterlinkObservationDependentVariableSettings ✗ no pickle     (defined in: include/tudat/simulation/estimation_setup/observationOutputSettings.h)
   LinkDefinition                           ✓ pickle        (defined in: include/tudat/astro/observation_models/linkTypeDefs.h)
   LinkEndId                                ✓ pickle        (defined in: include/tudat/astro/observation_models/linkTypeDefs.h)
   NonSequentialPropagationTerminationSettings ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/propagationTerminationSettings.h)
   ObservationAncillarySimulationSettings   ✓ pickle        (defined in: include/tudat/astro/observation_models/observationAncillarySettings.h)
   ObservationDependentVariableBookkeeping  ✓ pickle        (defined in: include/tudat/simulation/estimation_setup/observationOutput.h)
   ObservationDependentVariableSettings     ✓ pickle        (defined in: include/tudat/simulation/estimation_setup/observationOutputSettings.h)
   PropagationCPUTimeTerminationSettings    ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/propagationTerminationSettings.h)
   PropagationCustomTerminationSettings     ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/propagationTerminationSettings.h)
   PropagationDependentVariableTerminationSettings ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/propagationTerminationSettings.h)
   PropagationHybridTerminationSettings     ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/propagationTerminationSettings.h)
   PropagationTerminationDetails            ✓ pickle        (defined in: include/tudat/simulation/propagation_setup/propagationTermination.h)
   PropagationTerminationDetailsFromHybridCondition ✓ pickle        (defined in: include/tudat/simulation/propagation_setup/propagationTermination.h)
   PropagationTerminationSettings           ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/propagationTerminationSettings.h)
   PropagationTimeTerminationSettings       ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/propagationTerminationSettings.h)
   PropagatorType                           ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/propagationSettings.h)
   RootFinderSettings                       ✗ no pickle     (defined in: include/tudat/math/root_finders/createRootFinder.h)
   SingleDependentVariableSaveSettings      ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/propagationOutputSettings.h)
   SphericalHarmonicTorqueSettings          ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/torqueSettings.h)
   StationAngleObservationDependentVariableSettings ✗ no pickle     (defined in: include/tudat/simulation/estimation_setup/observationOutputSettings.h)
   Time                                     ✓ pickle        (defined in: include/tudat/basics/timeType.h)
   TorqueSettings                           ✗ no pickle     (defined in: include/tudat/simulation/propagation_setup/torqueSettings.h)
   type_caster                              ✓ pickle        (defined in: src/tudatpy/astro/time_representation/expose_time_representation.cpp)

✅ SERIALIZE TEST EXISTS:
   CustomAccelerationSettings               ✓ pickle        → tests/test_tudat/src/io/unitTestSettingsSerialization.cpp
   DirectTidalDissipationAccelerationSettings ✓ pickle        → tests/test_tudat/src/io/unitTestSettingsSerialization.cpp
   EmpiricalAccelerationSettings            ✓ pickle        → tests/test_tudat/src/io/unitTestSettingsSerialization.cpp
   MomentumWheelDesaturationAccelerationSettings ✓ pickle        → tests/test_tudat/src/io/unitTestSettingsSerialization.cpp
   MutualSphericalHarmonicAccelerationSettings ✓ pickle        → tests/test_tudat/src/io/unitTestSettingsSerialization.cpp
   ObservationCollection                    ✓ pickle        → tests/test_tudat/src/io/unitTestObservationCollectionSerialization.cpp
   RTGAccelerationSettings                  ✓ pickle        → tests/test_tudat/src/io/unitTestSettingsSerialization.cpp
   RadiationPressureAccelerationSettings    ✓ pickle        → tests/test_tudat/src/io/unitTestSettingsSerialization.cpp
   RelativisticAccelerationCorrectionSettings ✓ pickle        → tests/test_tudat/src/io/unitTestSettingsSerialization.cpp
   SimulationResults                        ✓ pickle        → tests/test_tudat/src/io/unitTestPropagationResultsSerialization.cpp
   SingleObservationSet                     ✓ pickle        → tests/test_tudat/src/io/unitTestObservationCollectionSerialization.cpp
   SphericalHarmonicAccelerationSettings    ✓ pickle        → tests/test_tudat/src/io/unitTestSettingsSerialization.cpp
   ThrustAccelerationSettings               ✓ pickle        → tests/test_tudat/src/io/unitTestSettingsSerialization.cpp
   YarkovskyAccelerationSettings            ✓ pickle        → tests/test_tudat/src/io/unitTestSettingsSerialization.cpp

DominicDirkx:
Does this supersede #652 ?

include/tudat/basics/timeType.h
```diff
@@ -17,6 +17,12 @@
 #include <cmath>
 #include <algorithm>
 
+#include <boost/algorithm/string/trim.hpp>
+#include <boost/lexical_cast.hpp>
+#include <boost/algorithm/string/split.hpp>
+#include <boost/date_time/gregorian/gregorian.hpp>
```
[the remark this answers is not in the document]
DominicDirkx: @PascalChristiaanse Plase check the additions of include statements (here and in other files) and remove ones that are unneeded

include/tudat/simulation/propagation_setup/torqueSettings.h
```diff
@@ -94,6 +129,21 @@ class CustomTorqueSettings : public TorqueSettings
...
+
+private:
+    friend class cereal::access;
+
+    template< class Archive >
+    void serialize( Archive& ar )
+    {
+        ar( cereal::base_class< TorqueSettings >( this ) );
+        // Warning: std::function member cannot be serialized
+        std::cerr << "Warning: serializing/deserializing CustomTorqueSettings, std::function member 'torqueFunction_' will not be preserved." << std::endl;
```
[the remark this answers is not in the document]
DominicDirkx: @PascalChristiaanse Have you checked what the behaviour is of this warning? I do think there should be a warning to the user (but not an exception). Also check the same for the other cerr outputs you introduced

include/tudat/simulation/propagation_setup/accelerationSettings.h
```diff
@@ -636,6 +796,25 @@ class RTGAccelerationSettings : public AccelerationSettings
...
+    friend class cereal::access;
+
+    template< class Archive >
+    void serialize( Archive& ar )
+    {
+        ar( cereal::base_class< AccelerationSettings >( this ) );
+        ar( const_cast< Eigen::Vector3d& >( bodyFixedForceVectorAtReferenceEpoch_ ) );
+        ar( const_cast< double& >( decayScaleFactor_ ) );
+        ar( const_cast< double& >( referenceEpoch_ ) );
+    }
```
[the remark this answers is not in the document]
DominicDirkx: @PascalChristiaanse I'm not sure it's important for the member variables of this object to be const. Feel free to remove the const

include/tudat/simulation/estimation_setup/singleObservationSet.h
```diff
@@ -1259,6 +1266,38 @@ class SingleObservationSet
...
+        ar( const_cast< LinkEndType& >( referenceLinkEnd_ ) );
+        ar( observationsDependentVariables_ );
+        ar( dependentVariableBookkeeping_ );
+        ar( const_cast< std::shared_ptr< observation_models::ObservationAncillarySimulationSettings >& >( ancillarySettings_ ) );
+        ar( numberOfObservations_ );
+        ar( singleObservationSize_ );
+        ar( weights_ );
+        ar( residuals_ );
+        ar( filteredObservationSet_ );
+    }
```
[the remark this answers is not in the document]
DominicDirkx: @PascalChristiaanse For the SingleObservationSet, the use of const is important (I think) to prevent data structures from being modified in an inconsistent manner after creation

include/tudat/simulation/propagation_setup/propagationResults.h
```diff
@@ -624,6 +689,22 @@ class SingleArcVariationalSimulationResults : public SimulationResults< StateSca
...
+    void serialize( Archive& ar )
+    {
+        ar( cereal::base_class< SimulationResults< StateScalarType, TimeType > >( this ) );
+        // const_cast needed because members are declared const for runtime safety,
+        // but deserialization must populate them
+        ar( const_cast< std::shared_ptr< SingleArcSimulationResults< StateScalarType, TimeType > >& >( singleArcDynamicsResults_ ) );
+        ar( const_cast< int& >( stateTransitionMatrixSize_ ) );
+        ar( const_cast< int& >( sensitivityMatrixSize_ ) );
+        ar( stateTransitionSolution_ );
+        ar( sensitivitySolution_ );
```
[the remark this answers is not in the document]
DominicDirkx: @PascalChristiaanse For the SimulationResults, the use of const is important (I think) to prevent data structures from being modified in an inconsistent manner after creation

tests/test_tudat/src/io/unitTestPropagationResultsSerialization.cpp
```diff
@@ -0,0 +1,287 @@
...
+void compareMaps( const MapType& left, const MapType& right )
+{
+    BOOST_REQUIRE_EQUAL( left.size( ), right.size( ) );
+    auto leftIterator = left.begin( );
+    auto rightIterator = right.begin( );
+    for( ; leftIterator != left.end( ); ++leftIterator, ++rightIterator )
+    {
+        BOOST_CHECK_EQUAL( leftIterator->first, rightIterator->first );
+    }
+}
```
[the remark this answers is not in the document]
DominicDirkx: @PascalChristiaanse Please correct

DominicDirkx:
@PascalChristiaanse I made some small changes to make this PR compatible with the latest merges onto develop. I also had a quick look through the copilot code review, and most of its points seem valid. I have not yet had a chance to review the PR myself

PascalChristiaanse:
Hi Dominic, Agreed on the copilot review points, ive so far ignored them to just get the bulk of the work done. Should i mark this PR as a draft, since its verrrrry much a work in progress?


notes to self (and if people want to leave opinions):
At the very minimum i still need to implement a versioning scheme for the files generated. I also want to change appropriate settings to XML so theyre human readable and editable (so not tabulatedSettings, but ie integratorSettings, as discussed).
Testing scheme is also not definitive yet; i kind of like implementing an equals function and operator overload everywhere, but im not sure what the consequences are for the user. Im exposing them to python to easily test the round trip integrity, but obviously if people notice an equals operator, theyll use it. It is not possible however to fully evaluate equality as we cant check function pointers.

PascalChristiaanse:
@DominicDirkx I think this will be the final version for serialization that i will implement; ive written up a design document that i think covers every design aspect. 

I intend to implement this on all classes and settings in the list ive posted a hundred times. This document mainly serves then to hold a consistent design, and hopefully allow others to implement the exact same design for classes they themselves require. This way I wont spend 100 hours implementing this for each settings class (there are soooooo many....).

We have discussed being able to create settings objects from existing objects, i have not included that in this design. I also do not think this would be required at this stage; especially considering we have already decided to not serialize runtime objects but only settings and results-type objects. 

Ill drop a link to this document in the slack too so people who are interested can drop their final thoughts. 

I think a final nice thing to mention is that with this setup, we can support "any" file format; so for example HDF5 would totally be possible, in fact it should be quite straight forward. 
# Serialization Design

Status: draft for team alignment

## Purpose

This document defines the serialization approach for Tudat and tudatpy so that C++ serialization, Python pickle support, file formats, versioning, and tests all follow the same rules.

The goal is not to describe every class in detail. The goal is to establish one consistent contract that can be applied across the codebase while the remaining classes are finished.

## Current State

The repository already has a partial cereal-based serialization stack:

- `include/tudat/io/serialization/base.h` provides cereal archive helpers, Eigen support, and binary string helpers.
- `include/tudat/io/serialization.h` is the public umbrella header for the serialization layer.
- `include/tudat/io/observationSerialization.h` is a deprecated compatibility include that forwards to the new serialization base.
- C++ tests already exercise roundtrip serialization for settings, observation containers, and propagation results using cereal binary archives.
- Python bindings already expose pickle support for several classes by reusing cereal through pybind11.
- Python equality is already exposed for a number of serializable classes, which is the right foundation for roundtrip tests.

At the same time, a few pieces are still missing or inconsistent:

- I did not find a centralized class version manifest in `src/` or `include/` yet.
- I did not find `CEREAL_CLASS_VERSION` declarations in the current source tree search.
- I did not find a unified `save` / `load` convention in the current source tree search.
- Polymorphic registration appears to be handled in binding code for some Python-exposed types, but not yet as a clearly centralized per-submodule policy.

That means the project already has working pieces, but not yet a single serialization contract that every developer can follow.

## Design Principles

1. Use cereal as the single serialization library.
2. Prefer explicit `save` and `load` methods on each class over a single generic `serialize` method.
3. Use XML for non-tabulated settings objects.
4. Use regular binary serialization for tabulated settings and other data-heavy objects.
5. Give every serializable class its own version number.
6. Keep a central manifest for current serialization versions.
7. Register polymorphic cereal types in one place per submodule.
8. Treat roundtrip survival plus equality as the primary correctness check.
9. Expose the same roundtrip behavior to Python through pybind11.

## File Format Policy

The file format choice should follow the nature of the data:

- Non-tabulated settings: XML.
- Tabulated settings: binary.

The reason for this split is practical:

- XML is readable, diffable, and suitable for small structured settings.
- Binary is better for large numeric tables, avoids huge text files, and preserves floating-point data without text conversion.

The file format is a persistence concern. It should not be confused with Python pickle. Python pickle is an in-memory transport mechanism and does not need to use the on-disk XML format.

### Adding A New Archive Type

If we want a new file format such as HDF5, the cereal path is to add a new archive type, register it, and then reuse the same class `save` / `load` methods.

Minimal sketch:

```cpp
// archive definition (very small sketch, not a full HDF5 implementation)
class Hdf5OutputArchive : public cereal::OutputArchive< Hdf5OutputArchive >
{
public:
	explicit Hdf5OutputArchive( H5::H5File& file ) : file_( file ) {}

	template< class T >
	Hdf5OutputArchive& operator()( const T& value )
	{
		// write value to an HDF5 dataset or attribute
		return *this;
	}

private:
	H5::H5File& file_;
};

// make cereal aware that this is an archive type
CEREAL_REGISTER_ARCHIVE( Hdf5OutputArchive )
```

```cpp
// usage stays the same in the serializable class
template < class Archive >
void save( Archive& archive ) const
{
	archive( cereal::make_nvp( "value", value_ ) );
}
```

```cpp
// file writing code chooses the archive type
H5::H5File file( "settings.h5", H5F_ACC_TRUNC );
Hdf5OutputArchive archive( file );
archive( cereal::make_nvp( "MySettings", settings ) );
```

The key idea is that the class serialization does not change when the archive changes. Only the archive wrapper and its registration are new.

## Class Serialization Contract

Each serializable class should follow the same pattern:

- Implement `save` and `load` methods.
- Avoid relying on a single generic `serialize` method when object construction or version branching is easier with explicit save/load logic.
- Include version-aware branching inside the class methods when the format changes.
- Keep transient or runtime-only members out of the serialized state unless they can be reconstructed safely.

This is especially important for classes whose deserialization requires some members to be constructed before the rest of the state can be restored.

### Versioning

Each class must have a version identifier.

Recommended policy:

- The class owns the version number for its serialized representation.
- The current version for all classes is published in one central manifest.
- The version number is written into the file metadata and used by class `load` methods to branch when reading older files.

This lets the team evolve file formats without guessing which classes changed or which archive format was used.

### Metadata

Every serialized file should carry metadata that identifies:

- The object or class version.
- The Tudat / tudatpy version used to generate the file.
- The serialization format family, if needed for future debugging.

This metadata should be stored in the serialized file itself, not in operating-system file properties. Cross-platform file contents are the only reliable place for it.

## Polymorphism

Polymorphic types should be registered in one canonical place per submodule.

That registration layer should contain the cereal type registration and the base-to-derived relationships required for pointer serialization.

This matters because cereal needs the registration information at compile time, and the project should not spread those declarations across unrelated files.

Recommended layout:

- Prefer one dedicated registration header per submodule.
- Put all derived cereal registrations for that submodule in that header.
- Have binding code and any other serialization users include that header.
- Only use a source-file registration TU when there is a strong reason to hide implementation details or control archive-specific linking behavior.

For this codebase, the header-based option is the default because it is easier to audit, easier to include from pybind modules, and less likely to create platform-specific linkage surprises.

Minimal example:

```cpp
// include/tudat/io/serialization/registrations/my_settings_registration.h
#pragma once

#include <cereal/types/polymorphic.hpp>
#include <cereal/archives/xml.hpp>
#include <cereal/archives/binary.hpp>

#include "my_settings.h"

CEREAL_REGISTER_TYPE( MyDerivedSettings )
CEREAL_REGISTER_POLYMORPHIC_RELATION( MyBaseSettings, MyDerivedSettings )
```

```cpp
// src/tudatpy/some_module/expose_my_settings.cpp
#include "tudat/io/serialization/registrations/my_settings_registration.h"

// the bindings can now use the registered polymorphic types safely
```

This keeps the registration in one obvious place while allowing the pybind or C++ users of the type to include the same header wherever they need the polymorphic metadata.

## Python Exposure

Python exposure should remain aligned with the C++ contract.

### Persistence Policy

Pickle is a convenience API, not the primary persistence mechanism.

Recommended policy:

- Use pickle for roundtrips, multiprocessing, caching, and tests.
- Use explicit file-oriented save/load methods for durable storage.
- Expose XML-saving methods for non-tabulated settings.
- Expose binary-saving methods for tabulated or data-heavy objects.

This keeps the Python API honest about what format is being used and avoids presenting pickle as the canonical file format.

### Pickle

Pickle support should be exposed through pybind11.

For value types, pickle can serialize the object state directly.

For polymorphic types, pickle should preserve the dynamic type through a base pointer, not flatten the object to the base class.

The current helper approach is the right shape:

- non-polymorphic types use a value-based pickle helper,
- polymorphic types use a shared-pointer-based pickle helper.

### Equality

The equality operator must be exposed to Python for every class that participates in roundtrip tests.

Recommended policy:

- The public Python API exposes `__eq__`.
- The C++ class exposes `operator==`.
- The class implementation should use a protected `equals` helper where inheritance makes that easier.
- For polymorphic hierarchies, the base class should expose the operators and the derived classes can specialize the equality implementation.

This matches the existing pattern in several parts of the codebase and keeps the test surface compact.

### XML Settings and Python Pickle

Open question: how should XML-saved settings objects appear in Python pickle?

Recommended answer:

- Do not couple Python pickle to XML file encoding.
- Pickle should serialize the in-memory C++ object state directly, using cereal through pybind11.
- XML remains the on-disk interchange format for settings files.

This separation keeps pickle simple, avoids forcing XML parsing into Python roundtrips, and allows future changes to the file format without breaking Python pickle semantics.

If a class needs a Python-visible persistence API beyond pickle, add explicit file export/import functions instead of overloading pickle with file-format concerns.

### Python File API

The preferred Python-facing persistence API should be explicit and format-specific.

Recommended method names:

- `save_xml(path)` / `load_xml(path)` for XML-backed settings objects.
- `save_binary(path)` / `load_binary(path)` for binary-backed objects.
- `to_xml_string()` / `from_xml_string()` only if an in-memory XML representation is genuinely useful.

Pickle should remain available, but it should not be the method users are told to use when they want to save a file for later reuse.

### Minimal Example

The intended pattern is:

```cpp
// C++ class: only one save/load pair is needed for cereal.
class MySettings
{
public:
	template <class Archive>
	void save( Archive& archive ) const
	{
		archive( cereal::make_nvp( "value", value_ ) );
	}

	template <class Archive>
	void load( Archive& archive )
	{
		archive( cereal::make_nvp( "value", value_ ) );
	}

	int value_ = 0;
};
```

```cpp
// Python exposure: explicit file methods, implemented as thin wrappers.
py::class_< MySettings, std::shared_ptr< MySettings > >( m, "MySettings" )
	.def( py::init<>() )
	.def( "save_xml", []( const MySettings& obj, const std::string& path )
	{
		std::ofstream stream( path );
		cereal::XMLOutputArchive archive( stream );
		archive( cereal::make_nvp( "MySettings", obj ) );
	} )
	.def_static( "load_xml", []( const std::string& path )
	{
		std::ifstream stream( path );
		cereal::XMLInputArchive archive( stream );

		MySettings obj;
		archive( obj );
		return obj;
	} )
	.def( "save_binary", []( const MySettings& obj, const std::string& path )
	{
		std::ofstream stream( path, std::ios::binary );
		cereal::BinaryOutputArchive archive( stream );
		archive( cereal::make_nvp( "MySettings", obj ) );
	} )
	.def_static( "load_binary", []( const std::string& path )
	{
		std::ifstream stream( path, std::ios::binary );
		cereal::BinaryInputArchive archive( stream );

		MySettings obj;
		archive( obj );
		return obj;
	} )
	.def( py::pickle(
		[]( const MySettings& obj )
		{
			return py::bytes( tudat::serialization::serializeToBinaryString( obj ) );
		},
		[]( py::bytes data )
		{
			return tudat::serialization::deserializeFromBinaryString< MySettings >( data.cast< std::string >() );
		} ) );
```

The rule this example demonstrates is simple:

- cereal `save` / `load` live on the C++ class,
- Python file persistence is an explicit wrapper around the archive type,
- pickle stays a separate binary convenience path.

## File Extensions

Open question: what file extensions make sense?

Recommended policy:

- Use `.xml` for non-tabulated settings files.
- Use `.bin` for binary tabulated files.

If the team wants a more explicit binary extension, a domain-specific suffix such as `.tudatbin` is acceptable, but it is not required.

The important part is consistency across the repository, not the exact naming choice.

## Testing Policy

The primary test criterion should be roundtrip survival.

### C++ Tests

The C++ side should verify:

- The object can be serialized and deserialized.
- The restored object compares equal to the original object.
- The restored object keeps the correct dynamic type when polymorphism is involved.

Where possible, the tests should exercise real constructors and representative sample data instead of mocked internal state.

### Python Tests

The Python side should verify:

- Pickle roundtrip survival.
- Equality after unpickling.
- Concrete type preservation after unpickling for polymorphic classes.

At this stage, Python tests only need to confirm that the object survives the roundtrip. Later functional tests can assert more behavior once the serialization surface is stable.

### Equality Pattern

The recommended test pattern is:

1. Serialize the object.
2. Deserialize it.
3. Compare the result with `==`.
4. For polymorphic types, also verify `type(restored) is type(original)` on the Python side or an equivalent C++ dynamic-cast check.

This keeps the tests simple and makes version mismatches or missing registrations fail immediately.

## Current Implementation Limitations

The current codebase still needs a few follow-up decisions and implementation passes:

- A central version manifest does not yet exist.
- A standardized `save` / `load` pattern is not yet enforced everywhere.
- Type registration is not yet visibly centralized across all submodules.
- Some classes still need a final decision on whether they are serializable at all if they contain callbacks or other non-serializable runtime state.

One concrete example is callback-heavy or function-heavy settings objects. Those should either provide custom serialization logic or be explicitly excluded with a documented reason.

## Rollout Plan

1. Add or finalize the central version manifest.
2. Convert remaining serializable classes to explicit `save` / `load` methods.
3. Add or move polymorphic registration into one file per submodule.
4. Add file metadata support for version and Tudat/tudatpy version.
5. Expand C++ roundtrip tests for the remaining classes.
6. Expand Python pickle tests to cover the remaining exposed types.
7. Add migration tests for at least one older file format version per major object family.

## Acceptance Criteria

The serialization feature is ready when:

- Every intended serializable class has a versioned `save` / `load` implementation.
- Every intended polymorphic type is registered in the correct submodule registration point.
- Every public serializable class has a roundtrip equality test in C++.
- Every Python-exposed serializable class has a pickle roundtrip test.
- Old files can be loaded according to their declared version.
- File metadata records both the file version and the Tudat/tudatpy version used to create the file.

DominicDirkx:
Thanks!! Could you clarify what you mean by these limitations:

> 
> At the same time, a few pieces are still missing or inconsistent:
> 
> I did not find a centralized class version manifest in src/ or include/ yet.
> I did not find CEREAL_CLASS_VERSION declarations in the current source tree search.
> I did not find a unified save / load convention in the current source tree search.
> Polymorphic registration appears to be handled in binding code for some Python-exposed types, but not yet as a clearly centralized per-submodule policy.

And in 'Design principles' you have this:

> Give every serializable class its own version number.
> Keep a central manifest for current serialization versions.
> 

could you clarify these?

PascalChristiaanse:
For sure; I missed this when proof reading as i wrote this at like 1 am and had copilot help me write legible sentences. The "i" in those sentences is the chatbot; ill update the document to fix this.


>  I did not find a centralized class version manifest in src/ or include/ yet.

I think its a good idea to have a central list of what versions each class is on, so we can keep track of when this changes and so if we need to account for possibly breaking people's saved files. This would mean we'd need to either have them remake those files, or provide a conversion method.

> I did not find CEREAL_CLASS_VERSION declarations in the current source tree search.

Each class must implement a parameter ```CEREAL_CLASS_VERSION``` in which the version of the class is saved. I think we can start with a TUDATVERSION::CLASSVERSION.CLASSPATCH scheme; so for example all classes would be 1.0.0::1.0

> I did not find a unified save / load convention in the current source tree search.

Some classes implement ```serialize(archive ar)``` when they should all implement ```save(archive ar)``` and ```load(archive ar)```
This is because i did a lot of work already while still working on the design. The reason for the split between loading and saving is because when loading, we might be able to reconstruct some function pointers and this becomes annoying when using the serialize method. I want to use the same everywhere because a polymorphism would require one to be chosen, and it simplifies the interaction (no need to check which one was used). Its slightly more lines of code however (most of the implementation work can be chatgpt'd anyways so who cares right).

> Polymorphic registration appears to be handled in binding code for some Python-exposed types, but not yet as a clearly centralized per-submodule policy.

Cereal requires polymorphic types to be registered so it knows which ones to pick when deserializing. I want this in a centralized spot per submodule as i think you're only allowed to register a type ONCE. I first thought it was only allowed in a single translation unit, but luckily its not that strict.

> And in 'Design principles' you have this:
> Give every serializable class its own version number.

This is just a design requirement stating "each serializable class SHALL implement a version number"

> Keep a central manifest for current serialization versions.

again, a requirement: "the software SHALL host a centralized place to collect each version currently active for each serializable class"

Like i mentioned in the slack message; if we change versions i think it will always require a custom solution. Maybe sometimes we can just do a dumb conversion, while sometimes we might need to ask the user for additional data before we can update their files. Thats why i havent included anything on the topic in the design document

docs/serialization_design.md
```diff
@@ -0,0 +1,412 @@
...
+- Include version-aware branching inside the class methods when the format changes.
+- Keep transient or runtime-only members out of the serialized state unless they can be reconstructed safely.
+
+This is especially important for classes whose deserialization requires some members to be constructed before the rest of the state can be restored.
+
+### Versioning
+
+Every serializable class must declare a version identifier using the `CEREAL_CLASS_VERSION` macro.
+
+The version scheme is: `TUDATVERSION::CLASSVERSION.CLASSPATCH`
```
PascalChristiaanse: Actually, this scheme is not that nice, as each new tudat version will invalidate all files; even if the class did not change. this should probably just be an integer like
vvvppp with v for version digits and p for patch digits

docs/serialization_design.md
```diff
@@ -0,0 +1,412 @@
...
+
+- The object or class version.
+- The Tudat / tudatpy version used to generate the file.
+- The serialization format family, if needed for future debugging.
+
+This metadata should be stored in the serialized file itself, not in operating-system file properties. Cross-platform file contents are the only reliable place for it.
+
+## Polymorphism
+
+Polymorphic types must be registered in one canonical place per submodule.
```
PascalChristiaanse: In fact, This might just be better to perform in the file where the class itself is defined; like is currently done in propagationResults.h

HannoSpreeuw (changes requested):
@suvayu and me think that we should address [the use-cases you defined in the issue](https://github.com/tudat-team/tudatpy/issues/669#issue-3954439529) not through serialization, but in a different manner, one that is more sustainable and allows for proper debugging, see [our comments](https://github.com/tudat-team/tudatpy/issues/669#issuecomment-4622568711).

HannoSpreeuw (approved):
After [discussion](https://github.com/tudat-team/tudatpy/issues/669#issuecomment-4648772301) we conclude that this PR is fine for serving the fourth use case, from [the five use cases listed in the issue](https://github.com/tudat-team/tudatpy/issues/669#issue-3954439529):
```
Saving intermediate results for complicated analyses
```

include/tudat/astro/low_thrust/shape_based/hodographicShapingOptimisationSetup.h
```diff
@@ -75,9 +75,15 @@ struct FixedTimeHodographicShapingOptimisationProblem {
     }
 
     template< typename Archive >
-    void serialize( Archive& ar )
+    void save( Archive& ar ) const
     {
-        ar( problemBounds_ );
+        ar( CEREAL_NVP( problemBounds_ ) );
```
PascalChristiaanse: This is not all fields; incorrect?

include/tudat/basics/timeType.h
```diff
@@ -1053,7 +1059,48 @@ class Time
...
     //! Pre-declared variable used in often-called normalizeMembers function
     int daysToAdd;
-};
+
+    //! Save time to a binary file
+    void saveToBinary( const std::string& path ) const;
+
+    //! Load time from a binary file
+    static Time loadFromBinary( const std::string& path );
+
```
PascalChristiaanse: @DominicDirkx Currently, all serializable classes will get a "save_to_<format>" method. It seems a bit excessive, but then again we dont really know what people will use this for, so maybe itll prove usefull?

include/tudat/astro/orbit_determination/podInputOutputTypes.h
```diff
@@ -1224,6 +1324,82 @@ struct EstimationOutput : public CovarianceAnalysisOutput< ObservationScalarType
...
+    // Used for serialization testing
+    bool operator==( const EstimationOutput& rhs ) const
+    {
+        return equals( rhs );
+    }
+
+    bool operator!=( const EstimationOutput& rhs ) const
+    {
+        return !equals( rhs );
+    }
```
PascalChristiaanse: Remove, inherit s from CovarianceAnalysisOutput

include/tudat/basics/timeType.h
```diff
@@ -504,8 +511,7 @@ class Time
      */
     friend bool operator==( const Time& timeToCompare1, const Time& timeToCompare2 )
     {
-        return ( ( timeToCompare1.getFullPeriods( ) == timeToCompare2.getFullPeriods( ) ) &&
-                 ( timeToCompare1.getSecondsIntoFullPeriod( ) == timeToCompare2.getSecondsIntoFullPeriod( ) ) );
+        return timeToCompare1.equals( timeToCompare2 );
     }
```
PascalChristiaanse: Should not be friend function, but just member

include/tudat/simulation/estimation_setup/observationOutput.h
```diff
@@ -240,6 +296,19 @@ class ObservationDependentVariableCalculator
...
+        return ( dependentVariableBookkeeping_ == rhs.dependentVariableBookkeeping_ ||
+                 ( dependentVariableBookkeeping_ && rhs.dependentVariableBookkeeping_ &&
+                   *dependentVariableBookkeeping_ == *rhs.dependentVariableBookkeeping_ ) ) &&
+                legLightTimeCalculators_ == rhs.legLightTimeCalculators_;
+    }
+
+    bool operator!=( const ObservationDependentVariableCalculator& rhs ) const
+    {
+        return !( *this == rhs );
+    }
```
PascalChristiaanse, line 328: should use equals() patern

include/tudat/simulation/estimation_setup/observationOutputSettings.h
```diff
@@ -628,4 +825,21 @@ inline std::shared_ptr< ObservationDependentVariableSettings > lightTimeCorrecti
...
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::simulation_setup::ObservationDependentVariableSettings,
+                                      tudat::simulation_setup::StationAngleObservationDependentVariableSettings )
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::simulation_setup::ObservationDependentVariableSettings,
+                                      tudat::simulation_setup::InterlinkObservationDependentVariableSettings )
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::simulation_setup::ObservationDependentVariableSettings,
+                                      tudat::simulation_setup::AncillaryObservationDependentVariableSettings )
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::simulation_setup::ObservationDependentVariableSettings,
+                                      tudat::simulation_setup::LightTimeCorrectionComponentsDependentVariableSettings )
+
 #endif  // TUDAT_OBSERVATIONOUTPUTSETTINGS
```
PascalChristiaanse: all polymorphic registrations need to be centralized according to spec

include/tudat/simulation/estimation_setup/singleObservationSet.h
```diff
@@ -1271,6 +1278,80 @@ class SingleObservationSet
...
+        ar( observationTimes_ );
+        ar( const_cast< LinkEndType& >( referenceLinkEnd_ ) );
+        ar( observationsDependentVariables_ );
+        ar( dependentVariableBookkeeping_ );
+        ar( const_cast< std::shared_ptr< observation_models::ObservationAncillarySimulationSettings >& >( ancillarySettings_ ) );
+        ar( numberOfObservations_ );
+        ar( singleObservationSize_ );
+        ar( weights_ );
+        ar( residuals_ );
+        ar( filteredObservationSet_ );
```
PascalChristiaanse: Preferably use CEREAL_NVP helper macro from cereal/cereal.hpp

include/tudat/astro/observation_models/linkTypeDefs.h
```diff
@@ -68,14 +74,22 @@ struct LinkEndId {
...
     friend bool operator==( const LinkEndId& linkEnd1, const LinkEndId& linkEnd2 )
     {
-        return ( ( linkEnd1.bodyName_ == linkEnd2.bodyName_ ) && ( linkEnd1.stationName_ == linkEnd2.stationName_ ) );
+        return linkEnd1.equals( linkEnd2 );
     }
 
     friend bool operator!=( const LinkEndId& linkEnd1, const LinkEndId& linkEnd2 )
     {
         return !operator==( linkEnd1, linkEnd2 );
     }
```
PascalChristiaanse: should not be friend function, but just member

include/tudat/io/serialization.h
PascalChristiaanse, line 1: Is this file really necessary?

include/tudat/math/interpolators/createInterpolator.h
```diff
@@ -810,4 +904,9 @@ convertBetweenStaticDynamicEigenTypeInterpolators(
 
 }  // namespace tudat
 
+CEREAL_REGISTER_TYPE( tudat::interpolators::InterpolatorSettings )
+CEREAL_REGISTER_TYPE( tudat::interpolators::LagrangeInterpolatorSettings )
+
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::interpolators::InterpolatorSettings, tudat::interpolators::LagrangeInterpolatorSettings )
+
```
PascalChristiaanse: polymorphic relations should be registered in a central place according to the spec

include/tudat/simulation/environment_setup/createGravityFieldVariations.h
```diff
@@ -370,12 +515,45 @@ class ModeCoupledSolidBodyGravityFieldVariationSettings : public GravityFieldVar
...
+        ar( CEREAL_NVP( deformingBodies_ ) );
+        ar( CEREAL_NVP( loveNumbers_ ) );
+    }
+
+    template< class Archive >
+    void load( Archive& ar )
+    {
+        ar( CEREAL_NVP( deformingBodies_ ) );
+        ar( CEREAL_NVP( loveNumbers_ ) );
+    }
```
PascalChristiaanse, line 554: doesnt explicitely serialize base-class?

include/tudat/simulation/environment_setup/createGravityFieldVariations.h
```diff
@@ -326,6 +450,27 @@ class BasicSolidBodyGravityFieldVariationSettings : public GravityFieldVariation
...
+        ar( CEREAL_NVP( meanForcingSineTerms_ ) );
+    }
+
+    template< class Archive >
+    void load( Archive& ar )
+    {
+        ar( CEREAL_NVP( deformingBodies_ ) );
+        ar( CEREAL_NVP( loveNumbers_ ) );
+        ar( CEREAL_NVP( meanForcingCosineTerms_ ) );
+        ar( CEREAL_NVP( meanForcingSineTerms_ ) );
```
PascalChristiaanse, line 467: Doesnt explicitly serialize base class?

include/tudat/simulation/environment_setup/createGravityFieldVariations.h
```diff
@@ -457,6 +654,27 @@ class TabulatedGravityFieldVariationSettings : public GravityFieldVariationSetti
...
+    }
+
+    template< class Archive >
+    void load( Archive& ar )
+    {
+        ar( CEREAL_NVP( cosineCoefficientCorrections_ ) );
+        ar( CEREAL_NVP( sineCoefficientCorrections_ ) );
+        ar( CEREAL_NVP( minimumDegree_ ) );
+        ar( CEREAL_NVP( minimumOrder_ ) );
+    }
```
PascalChristiaanse, line 678: baseclass serialization?

include/tudat/simulation/environment_setup/createGravityFieldVariations.h
```diff
@@ -532,6 +772,35 @@ class PeriodicGravityFieldVariationsSettings : public GravityFieldVariationSetti
...
+    void load( Archive& ar )
+    {
+        ar( CEREAL_NVP( cosineShAmplitudesCosineTime_ ) );
+        ar( CEREAL_NVP( cosineShAmplitudesSineTime_ ) );
+        ar( CEREAL_NVP( sineShAmplitudesCosineTime_ ) );
+        ar( CEREAL_NVP( sineShAmplitudesSineTime_ ) );
+        ar( CEREAL_NVP( frequencies_ ) );
+        ar( CEREAL_NVP( referenceEpoch_ ) );
+        ar( CEREAL_NVP( minimumDegree_ ) );
+        ar( CEREAL_NVP( minimumOrder_ ) );
```
PascalChristiaanse, line 806: baseclass?

include/tudat/simulation/environment_setup/createGravityFieldVariations.h
```diff
@@ -581,6 +869,29 @@ class PolynomialGravityFieldVariationsSettings : public GravityFieldVariationSet
...
+
+    template< class Archive >
+    void load( Archive& ar )
+    {
+        ar( CEREAL_NVP( cosineAmplitudes_ ) );
+        ar( CEREAL_NVP( sineAmplitudes_ ) );
+        ar( CEREAL_NVP( referenceEpoch_ ) );
+        ar( CEREAL_NVP( minimumDegree_ ) );
+        ar( CEREAL_NVP( minimumOrder_ ) );
+    }
```
PascalChristiaanse, line 901: baseclass

include/tudat/simulation/environment_setup/createGravityFieldVariations.h
```diff
@@ -833,4 +1144,23 @@ std::shared_ptr< gravitation::GravityFieldVariations > createGravityFieldVariati
...
+                                      tudat::simulation_setup::BasicSolidBodyGravityFieldVariationSettings )
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::simulation_setup::GravityFieldVariationSettings,
+                                      tudat::simulation_setup::ModeCoupledSolidBodyGravityFieldVariationSettings )
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::simulation_setup::GravityFieldVariationSettings,
+                                      tudat::simulation_setup::TabulatedGravityFieldVariationSettings )
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::simulation_setup::GravityFieldVariationSettings,
+                                      tudat::simulation_setup::PeriodicGravityFieldVariationsSettings )
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::simulation_setup::GravityFieldVariationSettings,
+                                      tudat::simulation_setup::PolynomialGravityFieldVariationsSettings )
+
```
PascalChristiaanse: centralized PM registration

include/tudat/math/interpolators/createInterpolator.h
```diff
@@ -203,12 +262,47 @@ class LagrangeInterpolatorSettings : public InterpolatorSettings
...
+        ar( CEREAL_NVP( interpolatorOrder_ ) );
+        ar( CEREAL_NVP( lagrangeBoundaryHandling_ ) );
+    }
+
+    template< class Archive >
+    void load( Archive& ar )
+    {
+        ar( CEREAL_NVP( interpolatorOrder_ ) );
+        ar( CEREAL_NVP( lagrangeBoundaryHandling_ ) );
+    }
```
PascalChristiaanse, line 303: baseclass serialization?

include/tudat/simulation/propagation_setup/accelerationSettings.h
```diff
@@ -422,6 +632,48 @@ class EmpiricalAccelerationSettings : public AccelerationSettings
...
+        cosineAcceleration_( Eigen::Vector3d::Zero( ) )
+    {}
+    //! Zero-arg default constructor for cereal
+    // (constructor with default parameters above is already default-constructible)
+
+    // Used for serialization testing
+    bool equals( const AccelerationSettings& other ) const override
+    {
+        const auto* rhs = dynamic_cast< const EmpiricalAccelerationSettings* >( &other );
+        if( !rhs )
```
PascalChristiaanse, line 688: probably should check if all derived classes are checking a dynamic_cast to see if the types are indeed equal. not guaranteed bc polymorphic instancing

include/tudat/simulation/propagation_setup/accelerationSettings.h
```diff
@@ -726,6 +1082,50 @@ class RTGAccelerationSettings : public AccelerationSettings
...
+    }
+
+    template< class Archive >
+    void load( Archive& ar )
+    {
+        ar( cereal::base_class< AccelerationSettings >( this ) );
+        ar( cereal::make_nvp( "bodyFixedForceVectorAtReferenceEpoch_",
+                              const_cast< Eigen::Vector3d& >( bodyFixedForceVectorAtReferenceEpoch_ ) ) );
+        ar( cereal::make_nvp( "decayScaleFactor_", const_cast< double& >( decayScaleFactor_ ) ) );
+        ar( cereal::make_nvp( "referenceEpoch_", const_cast< double& >( referenceEpoch_ ) ) );
```
PascalChristiaanse: Copilot doesnt like const_cast, lets find out why?

include/tudat/simulation/propagation_setup/accelerationSettings.h
```diff
@@ -819,4 +1042,40 @@ typedef std::map< std::string, std::vector< std::pair< std::string, std::shared_
...
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::simulation_setup::AccelerationSettings,
+                                      tudat::simulation_setup::ThrustAccelerationSettings )
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::simulation_setup::AccelerationSettings,
+                                      tudat::simulation_setup::CustomAccelerationSettings )
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::simulation_setup::AccelerationSettings,
+                                      tudat::simulation_setup::RTGAccelerationSettings )
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::simulation_setup::AccelerationSettings,
+                                      tudat::simulation_setup::DirectTidalDissipationAccelerationSettings )
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::simulation_setup::AccelerationSettings,
+                                      tudat::simulation_setup::MomentumWheelDesaturationAccelerationSettings )
```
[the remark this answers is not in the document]
PascalChristiaanse: see spec; should be centralized

include/tudat/simulation/propagation_setup/propagationOutputSettings.h
```diff
@@ -1585,4 +2315,91 @@ inline std::shared_ptr< SingleDependentVariableSaveSettings > properTimeRatePote
 
 }  // namespace tudat
 
+// Register derived classes for polymorphic serialization
```
PascalChristiaanse: PM registration should be centralized

include/tudat/simulation/propagation_setup/propagationResults.h
```diff
@@ -624,6 +774,57 @@ class SingleArcVariationalSimulationResults : public SimulationResults< StateSca
...
+    {
+        static_cast< void >( version );
+        ar( cereal::base_class< SimulationResults< StateScalarType, TimeType > >( this ) );
+        // const_cast needed because members are declared const for runtime safety,
+        // but deserialization must populate them
+        ar( cereal::make_nvp(
+                "singleArcDynamicsResults_",
+                const_cast< std::shared_ptr< SingleArcSimulationResults< StateScalarType, TimeType > >& >( singleArcDynamicsResults_ ) ) );
+        ar( cereal::make_nvp( "stateTransitionMatrixSize_", const_cast< int& >( stateTransitionMatrixSize_ ) ) );
+        ar( cereal::make_nvp( "sensitivityMatrixSize_", const_cast< int& >( sensitivityMatrixSize_ ) ) );
```
PascalChristiaanse: copilot doesnt like const_cast?

include/tudat/simulation/propagation_setup/propagationResults.h
```diff
@@ -1056,15 +1350,123 @@ class HybridArcSimulationResults : public SimulationResults< StateScalarType, Ti
...
+using SingleArcVariationalResultsDT = SingleArcVariationalSimulationResults< double, Time >;
+using MultiArcDynamicsResultsDT = MultiArcSimulationResults< SingleArcSimulationResults, double, Time >;
+using MultiArcVariationalResultsDT = MultiArcSimulationResults< SingleArcVariationalSimulationResults, double, Time >;
+using HybridArcDynamicsResultsDT = HybridArcSimulationResults< SingleArcSimulationResults, double, Time >;
+using HybridArcVariationalResultsDT = HybridArcSimulationResults< SingleArcVariationalSimulationResults, double, Time >;
+
+// Base‐class aliases (needed for CEREAL macros, which cannot handle commas in template args)
+using SimulationResultsDD = SimulationResults< double, double >;
+using SimulationResultsDT = SimulationResults< double, Time >;
+
```
PascalChristiaanse, line 1449: Feels janky, is this the only way?

include/tudat/simulation/propagation_setup/propagationResults.h
```diff
@@ -1056,15 +1350,123 @@ class HybridArcSimulationResults : public SimulationResults< StateScalarType, Ti
...
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::propagators::SimulationResultsDD, tudat::propagators::HybridArcDynamicsResults )
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::propagators::SimulationResultsDD, tudat::propagators::HybridArcVariationalResults )
+
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::propagators::SimulationResultsDT, tudat::propagators::SingleArcDynamicsResultsDT )
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::propagators::SimulationResultsDT, tudat::propagators::SingleArcVariationalResultsDT )
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::propagators::SimulationResultsDT, tudat::propagators::MultiArcDynamicsResultsDT )
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::propagators::SimulationResultsDT, tudat::propagators::MultiArcVariationalResultsDT )
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::propagators::SimulationResultsDT, tudat::propagators::HybridArcDynamicsResultsDT )
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::propagators::SimulationResultsDT, tudat::propagators::HybridArcVariationalResultsDT )
+
```
PascalChristiaanse: should be centralized, see spec

include/tudat/simulation/propagation_setup/propagationSettings.h
```diff
@@ -56,6 +58,45 @@ struct PropagatorType {
...
+    //! Equality comparison for PropagatorType
+    friend bool operator==( const PropagatorType& a, const PropagatorType& b )
+    {
+        return a.equals( b );
+    }
+
+    friend bool operator!=( const PropagatorType& a, const PropagatorType& b )
+    {
+        return !( a == b );
+    }
```
PascalChristiaanse, line 92: should not be friend, but member. Should be public?

include/tudat/simulation/propagation_setup/propagationTermination.h
```diff
@@ -726,13 +771,53 @@ class PropagationTerminationDetailsFromHybridCondition : public PropagationTermi
...
 
 }  // namespace propagators
 
 }  // namespace tudat
 
+CEREAL_REGISTER_TYPE( tudat::propagators::PropagationTerminationDetails )
+CEREAL_REGISTER_TYPE( tudat::propagators::PropagationTerminationDetailsFromHybridCondition )
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::propagators::PropagationTerminationDetails,
+                                      tudat::propagators::PropagationTerminationDetailsFromHybridCondition )
+
```
PascalChristiaanse: centralized

include/tudat/simulation/propagation_setup/propagationTerminationSettings.h
```diff
@@ -329,4 +603,25 @@ inline std::shared_ptr< PropagationTerminationSettings > nonSequentialPropagatio
...
+                                      tudat::propagators::PropagationCPUTimeTerminationSettings )
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::propagators::PropagationTerminationSettings,
+                                      tudat::propagators::PropagationDependentVariableTerminationSettings )
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::propagators::PropagationTerminationSettings,
+                                      tudat::propagators::PropagationCustomTerminationSettings )
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::propagators::PropagationTerminationSettings,
+                                      tudat::propagators::PropagationHybridTerminationSettings )
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::propagators::PropagationTerminationSettings,
+                                      tudat::propagators::NonSequentialPropagationTerminationSettings )
+
```
PascalChristiaanse: centralized

include/tudat/simulation/propagation_setup/torqueSettings.h
```diff
@@ -11,6 +11,15 @@
 #ifndef TUDAT_TORQUESETTINGS_H
 #define TUDAT_TORQUESETTINGS_H
 
+#include <boost/tuple/tuple.hpp>
+#include <iostream>
```
PascalChristiaanse: remove
```suggestion
#include <iostream>
```

include/tudat/simulation/propagation_setup/torqueSettings.h
```diff
@@ -189,4 +301,11 @@ typedef std::map< std::string, std::map< std::string, std::vector< std::shared_p
 
 }  // namespace tudat
 
+CEREAL_REGISTER_TYPE( tudat::simulation_setup::TorqueSettings )
+CEREAL_REGISTER_TYPE( tudat::simulation_setup::SphericalHarmonicTorqueSettings )
+CEREAL_REGISTER_TYPE( tudat::simulation_setup::CustomTorqueSettings )
+
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::simulation_setup::TorqueSettings, tudat::simulation_setup::SphericalHarmonicTorqueSettings )
+CEREAL_REGISTER_POLYMORPHIC_RELATION( tudat::simulation_setup::TorqueSettings, tudat::simulation_setup::CustomTorqueSettings )
+
```
PascalChristiaanse: centraize

include/tudat/io/serialization/base.h
```diff
@@ -0,0 +1,415 @@
...
+    matrix.resize( rows, cols );
+
+    for( Eigen::Index i = 0; i < rows; ++i )
+    {
+        for( Eigen::Index j = 0; j < cols; ++j )
+        {
+            ar( make_nvp( "element", matrix( i, j ) ) );
+        }
+    }
+}
```
PascalChristiaanse: are these implicitely called? defo not explicitely called, so if not necessary should be removed to avoid confusion

include/tudat/io/serialization/base.h
```diff
@@ -0,0 +1,415 @@
...
+
+//! Helper function to deserialize into an existing object from a binary string
+template< typename T >
+void deserializeFromBinaryString( const std::string& data, T& object )
+{
+    std::istringstream iss( data, std::ios::binary );
+    cereal::BinaryInputArchive ia( iss );
+    ia( object );
+}
```
PascalChristiaanse: are these still used? might just be replaced by macros

include/tudat/io/serialization/base.h
```diff
@@ -0,0 +1,415 @@
...
+    std::string filePath = path + ".tudat";
+    std::ifstream inputStream( filePath, std::ios::binary );
+    if( !inputStream )
+    {
+        throw std::runtime_error( "Unable to open file for binary load: " + filePath );
+    }
+
+    cereal::BinaryInputArchive archive( inputStream );
+    archive( object );
+}
```
PascalChristiaanse: maybe we shouldnt force the extension? maybe user want to specify themselves (idk why theyd want that but who knows)

include/tudat/io/serialization/base.h
```diff
@@ -0,0 +1,419 @@
...
+
+//! Helper function to serialize an object to a JSON string
+template< typename T >
+std::string serializeToJsonString( const T& object )
+{
+    std::ostringstream oss;
+    {
+        cereal::JSONOutputArchive oa( oss );
+        oa( cereal::make_nvp( "root", object ) );
+    }
```
PascalChristiaanse: should use cereal/cereal.hpp's CEREAL_NVP

docs/serialization_design.md
PascalChristiaanse, line 1: Remove for PR

find_tested_serialized.py
PascalChristiaanse, line 1: Remove for PR

track_serialization.py
PascalChristiaanse, line 1: Remove for PR

PascalChristiaanse (commented):
Processing copilot comments at the moment, but the design is fully implemented and ready for review. Theres some minor issues left:

Python side dynamic typecasting is not working, you must select the right derived type at the moment, this will be fixed at some point.
eg.

```
SimulationResult.load_from_binary(file: SingleArcSimulationResult) -> Throws RuntimeError but will work at some point

SingleArcSimulationResult.load_from_binary(file: SingleArcSimulationResult) -> Works
```

PascalChristiaanse:
@DominicDirkx Should i implement custom exceptions for serialization?

DominicDirkx:
Just one more unit test failure :) Almost there!

DominicDirkx:
> 
@DominicDirkx Should i implement custom exceptions for serialization?

Not for now, no. If we think it's useful, we can always create a new issue/PR to do this :)

PascalChristiaanse:
@DominicDirkx 

Build/test passes. 
CustomSomethingSettings classes currently throw when trying to serialize/deserialize. This is the safest behavior in my opinion, but we could also emit a warning and return none?

There are some methods in #base.h that im not entirely sure are being used; i should get rid of those at some point. Do you want to wait with merging before I do that? otherwise i think its ready considering removing dead functions wouldnt break anything

DominicDirkx:
And merging!

Issue #509, opened 2025-10-17
Protect master branch


HannoSpreeuw:
### Description:

At the last developer meeting, we agreed on protecting the master branch.
But what should that protection look like?
There is a lot of granularity to settle on.

See [this link](https://github.com/tudat-team/tudatpy/settings/rules/new?target=branch&enforcement=disabled) for all the options.

Which branch rules should apply?

Do we want a bypass list?

DominicDirkx:
I think we can keep it fairly simple for now: no changes to the master or develop branch except through a Github PR would be my proposal. 

p.s. I like the original issue name :)

DominicDirkx:
@HannoSpreeuw From your experience with these, any advice on which settings would be most suitable?

HannoSpreeuw:
Okay, to protect the master branch, I would tick these boxes:

1. Restrict deletions
2. Require pull request before merging
3. Block force pushes
4. Automatically request Copilot code review

HannoSpreeuw:
@DominicDirkx Please let me know if you agree with this and I will have it effectuated.

DominicDirkx:
Apologies, I missed this. Yes, this sounds good! I'm  not convinced yet about the Copilor review, but let's see how it goes. These riles would be for both master and develop right?

HannoSpreeuw:
> These riles would be for both master and develop right?

Yes, we could do both.

HannoSpreeuw:
This has been implemented and will affect both `develop` and `master`.

DominicDirkx:
Implementing this has led to an issue with the code auto-formatting, which runs after each commit to develop. See for instance here:

https://github.com/tudat-team/tudatpy/actions/runs/22567237478

due to the branch protection, the auto-formatting can no longer make commits. We should either

1) Modify the branch protection to allow the github action to modify the code without PR
2) Remove the auto-formatting and make this the responsibility of the developer
3) ?

My preference would be to go with (1) @HannoSpreeuw What do you think?

larshinueber:
Chiming in on this, we also have this as a topic for the developer meetings, with some discussion: https://tudat.slack.com/lists/TU777Q9EF/F08NJU0446S?record_id=Rec09CE3FQ3U6

I personally prefer (2), with a Github action that checks the formatting on every PR and fails if it detects any code that doesn't conform with the standard. I think this keeps our commit history cleaner and, once setup, will also make it easier to review PRs since the PR content will be stripped of any unformatted code.
[...]

PR #793, opened 2026-05-18
Formatting using pre-commit hooks

### LINKED ISSUE #792: [F&E]: Setup repository pre-commit hooks
### Description:

See #509 for full discussion.

Setup pre-commit hooks for automated formatting (C++ and Python) on each commit.

- [ ] setup pre-commit infrastructure 
- [ ] decide on Python formatter, decide on/setup config file
- [ ] enable Python pre-commit formatting
- [ ] enable C++ pre-commit formatting
- [ ] run pre-commit on all files, commit changes, add commit to `.git-blame-ignore-revs`
- [ ] add pre-commit CI action to verify successful run
- [ ] remove clang-format action
- [ ] make announcement for developers to setup pre-commit

### Additional context:

FYI @HannoSpreeuw

larshinueber:
Early draft PR to keep everyone in the loop and get feedback.

This draft PR configures pre-commit hooks for automatic code formatting for C++ (using clang-format) and Python (using black). Fixes #792.

# Background/Motivation
For C++, we have the [clang format workflow](https://github.com/tudat-team/tudatpy/actions/workflows/clang-format.yml), which should automatically format the codebase after each commit on develop and push the changes (if any) to develop.
The issue is that a) this creates quite a lot of noise in the commit history, b) PRs can contain unformatted code, making the review more tedious and c) this conflicts with the new branch protection rules. c) could be solved by allowing the action to bypass the branch protection rules, however this still leaves a) and b).

After some discussion in #509 we came to the result of using pre-commit hooks to take care of the formatting, as a compromise between burden on the developer side and simplicity of setup.
Pre-commit hooks basically run commands on certain git events. In this case, we use the [pre-commit framework](https://pre-commit.com) to run auto-formatting before each commit. From a developer-perspective, this doesn't require any attention, after running `pre-commit install` once.

Additionally, I propose to add a pre-commit GH action that runs on every PR and commit to main/develop. This action does not apply any formatting, it only runs the pre-commit checks and reports if they pass or fail. If they fail, the developer opening the PR would have to run the pre-commit command on the changed files. I would expect this to happen very rarely after the initial adoption period.

# PR Contents
This PR implements the pre-commit infrastructure for C++ and Python formatting.

## C++ formatting
For C++, we simply use the format as specified in the .clang-format file

## Python formatting
We haven't had a standard Python formatting before. However, based on previous discussion (https://tudat.slack.com/lists/TU777Q9EF/F08NJU0446S?record_id=Rec09CE3FQ3U6) I would argue for black. The only thing really to configure with black is the line-length. I've set it to 100 since I find the 140 line length from C++ quite excessive, but I'm happy to discuss this.

## Notes
The pre-commit formatting is applied tentatively in #806.

# TODO
- [x] pre-commit configuration 
- [x] Python auto-format configuration (black)
- [x] Add pre-commit GH action
- [x] Remove previous clang-format GH action
- [x] Apply pre-commit formatting in new PR: #806


DominicDirkx:
@larshinueber Could you clarify: at which point will the code now be auto-formatter? Is this along with every local commit made by a developer (once the pre-commit install has been run)?

Also, I see that the clang format file has been deleted. Where do we now define our code formatting rules?

larshinueber:
> Is this along with every local commit made by a developer (once the pre-commit install has been run)?

@DominicDirkx Correct 👍 There will no longer be separate commits for applying formatting, the formatter (clang-format/black) will be run automatically before the commit is made, such that only formatted code is committed.

> Also, I see that the clang format file has been deleted. Where do we now define our code formatting rules?

The clang-format configuration file is still there, I only deleted the clang-format GH action which will be redundant :)

DominicDirkx:
Then I think this can be merged! Could you make an announcement on #development in Slack to let people know of the update?

Issue #484, opened 2024-12-16
DSN transmitting frequency ramp tables overlap


valeriof7:
In PR tudat-team/tudat#268, the following code block in `include/tudat/astro/ground_stations/transmittingFrequencies.h` and relative unit test has been commented out, but needs to be addressed:
"""
// If there are discontinuities
            if ( startTimes_.at( i ) != endTimes_.at( i - 1 ) )
            {
                // If the start and end times are inconsistent throw error
                if ( endTimes_.at( i - 1 ) > startTimes_.at( i ) )
                {
                    throw std::runtime_error(
                            "Error when creating piecewise linear frequency interpolator: inconsistency between ramp end "
                            "time (" + std::to_string( double( endTimes_.at( i - 1 ) ) ) + ") and start time of the following ramp (" +
                            std::to_string( double( startTimes_.at( i ) ) ) + "); the end is smaller than the start time." );
                }
                // If there are gaps in the data save that information
                else
                {
                    invalidTimeBlocksStartTimes_.push_back( endTimes_.at( i - 1 ) );
                    invalidTimeBlocksEndTimes_.push_back( startTimes_.at( i ) );
                }
"""

In particular, we need to properly implement the logic to deal with overlaps in the DSN transmit frequency ramp table. Possible solutions suggested by @DominicDirkx:
- check if the tables are only slightly different (same frequencies possibly referenced to a different date)
- select one of the tables and reset the time range of the other one
- give a warning
- give an exception

DominicDirkx:
This PR: https://github.com/tudat-team/tudatpy/pull/563 made some modifications to the loading of transmission frequencies from IFMS files that are tangentially related to this.

I think there are a couple of related issues here:

1 What to do if a frequency is requested outside of the bounds where data is available?
2 What to do if a frequency is requested at a point where multiple ramp tables are available?

I think it's important to not hard-code anything, but let all values we use be user-definable (but, with nominal values if the user does not provide anything)

For (1), I would propose we simply extend the frequency data (extrapolate it in time beyond the formally valid region) for a given number of seconds (default 10 s?). This will be useful in case the the frequency is queried *just* before/after the definition of the ramp tables.  The more complicated question is what to do outside these bounds

* Throwing an error: may make the code crash when in case of (for instance) outliers
* Giving a warning: can spam the terminal with many warnings
* Just extrapolating further in time: can hide the fact that data is missing

[...]

PR #805, opened 2026-05-22
Corrections to ramp tables

DominicDirkx:
This PR updates transmitting-frequency ramp handling to support the DSN ramp-table overlap and gap cases from #484.

Changes:
- Adds `FrequencyGapHandling` for ramp-table gaps, to decide what is done when ramp tables are interpolated outside their valid domain :
  - extrapolate silently (default)
  - throw
  - print every time
  - print once
- Updates `PiecewiseLinearFrequencyInterpolator` so later ramp definitions take precedence in overlaps by truncating the previous interval.
- Updates frequency integration so it can integrate across ramp boundaries and extrapolated gaps consistently.
- Adds unit tests for overlapping ramps, gap extrapolation, exceptions, and warning modes.


larshinueber:
@DominicDirkx I was just running the mro_tnf_residuals_analysis example with a locally compiled kernel for some testing and I believe I'm getting an error from the modified ramp table handling:

```
Traceback (most recent call last):
  File "/Users/lhinuber/miniconda3/envs/tudatpy-dev/lib/python3.11/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
                    ^^^^^^^^^^^^^^^^^^^
  File "/Users/lhinuber/miniconda3/envs/tudatpy-dev/lib/python3.11/multiprocessing/pool.py", line 48, in mapstar
    return list(map(*args))
           ^^^^^^^^^^^^^^^^
  File "/var/folders/f2/fdr93pcd65127g43prjyz2k5r9sq14/T/ipykernel_78065/4139909652.py", line 188, in perform_residuals_analysis
    tnfProcessor.set_tnf_information_in_bodies(bodies)
  File "/Users/lhinuber/miniconda3/envs/tudatpy-dev/lib/python3.11/site-packages/tudatpy/data/processTrk234/processor.py", line 146, in set_tnf_information_in_bodies
    frequencyInterpolator = PiecewiseLinearFrequencyInterpolator(
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Error when creating piecewise linear frequency interpolator: ramp end time (-946724401.000000) is not after ramp start time (404088516.000000).
```

which traces back to the `PiecewiseLinearFrequencyInterpolator::initialize( )` method. I will try to debug it tomorrow!

DominicDirkx:
Looks like those start/end times are not consistent, but throwing an error now causes an exception... Maybe just skip this entry and keep a record of failed ramps (with the option to print a warning)

larshinueber:
From what I can tell the error above is the consequence of the TNF file not having a record that ends the ramping explicit or any subsequent record, which leaves the end time of the ramp as `None`. This then gets implicitly converted to a Unix time epoch of 0 (1970-01-01).
From the TRK2-34 specification it sounds that this is expected for non-UPL-DTT antennas (e.g. DSS-63) but in the MRO example it also seems to happen for UPL-DTT antennas (e.g. DSS-25, 26, 65, ...):
image

I'm not sure how this would be handled best? Adding an artificial end time with an arbitrary small offset fixes the error      but then relies on the assumption of extrapolation, which might cause issues in some cases? Taking the last record of the TNF file as end time might interfere if the station has subsequent ramps from another spacecraft that is not part of the TNF processor (?)

PR #823, opened 2026-06-08
Feature/mcd integration

DominicDirkx:
Local version (from tudat-team) of https://github.com/tudat-team/tudatpy/pull/622 to allow repo secret to be used


DominicDirkx:
@nikemais All right, the CI finally seems to be happy (getting the Windows Fortran stuff to work was a pain). I made some changes so that the ClimateModel does not have the Body as a member object anymore (in line with all other environment models)

docs/tudatpy/source/conf.py
```diff
@@ -69,9 +71,23 @@ def filter_mcd_docs(app, docname, source):
         return
 
     text = source[0]
-    text = text.replace("\n   mcd\n", "\n")
+    text = text.replace("\n   mars_climate_database_climate_model\n", "\n")
```
DominicDirkx: Why is the conf.py manually adding MCD things?

PR #825, opened 2026-06-09
fixed candence gap handeling after filtering IFMS files

### LINKED ISSUE #824: [BUG]: IFMS reader fails when invalid-line filtering creates cadence gaps
### Summary

  Loading IFMS files with `remove_invalid_lines=true` can fail when invalid rows are removed from the
  middle of a file. The IFMS processor currently infers the Doppler integration time from the first two
  remaining UTC timestamps, then requires every consecutive timestamp difference to match within `0.01 s`.

  If an invalid row is filtered out in the middle of an otherwise regularly sampled file, the processed
  timestamps contain a gap and Tudat throws:

 ` RuntimeError: Error when getting integration time for processed file contents, step is not equal`

  ### Example

  Raw IFMS cadence:

  10:00:00 valid
  10:00:10 valid
  10:00:20 invalid -> removed
  10:00:30 valid

  After filtering:

  10:00:00
  10:00:10
  10:00:30

  The nominal cadence is still 10 s, but Tudat sees a 20 s jump and fails.

Load a IFMS file with a cadence gap.

### OS and Python version:

OS: maxOS 25.5.0 ARM64
Python: 3.13.5

### Expected behavior:

Tudat should tolerate cadence gaps caused by filtered invalid IFMS rows. The nominal `doppler_integration_time` should still be inferred as the regular single-sample cadence, while larger timestamp jumps should be treated as gaps.

### Current behavior:

  The processed file is rejected because not all consecutive timestamp differences are equal.

### Additional context:

_No response_

Markus-Reichel:
## Summary

  Fixes #824.

  This PR fixes IFMS Doppler processing when invalid-line filtering creates gaps in an otherwise regularly sampled
  file.

  Previously, IFMS files loaded with `remove_invalid_lines=true` could fail during observation collection creation.
  Invalid rows are filtered out first, and then `ProcessedTrackingTxtFileContents::getObservationTimeStep()`
  inferred the Doppler integration time from the first two remaining UTC timestamps and required every following
  timestamp difference to match within `0.01 s`.

  This fails when an invalid row occurs in the middle of a file:

```
  10:00:00 valid
  10:00:10 valid
  10:00:20 invalid -> removed
  10:00:30 valid
```

  After filtering, Tudat sees:

```
  10:00:00
  10:00:10
  10:00:30
```

  The nominal integration time is still 10 s, but the previous implementation treated the 20 s jump as fatal.

  ## Implementation

  ### Robust IFMS cadence inference

  `getObservationTimeStep()` now infers the nominal Doppler cadence from valid positive UTC time differences instead
  of requiring all consecutive filtered timestamps to be identical.

  The new behavior is:

  - require at least two observations
  - compute all consecutive UTC deltas
  - infer a nominal positive cadence
  - preserve previous gap-free behavior by using the first valid step unless the file clearly starts with a gap
  - classify larger-than-nominal jumps as cadence gaps
  - keep throwing for invalid timing cases such as non-positive steps, no inferable positive cadence, or steps
    smaller than the inferred cadence beyond tolerance.

  The inferred nominal cadence is still stored as doppler_integration_time.

  ### Warning for cadence gaps

  When cadence gaps are detected, Tudat prints a concise warning containing:

  - the tracking file name from TrackingDataType::file_name, if available;
  - a fallback label for unknown files;
  - number of cadence gaps;
  - inferred nominal cadence;
  - details for the first few gaps: index, previous UTC timestamp, next UTC timestamp, and observed delta.

  Gap-free files do not emit this warning.

  ### Run-based Doppler compression

  `compressDopplerData(...)` now avoids averaging observations across cadence gaps.

  Previously, compression used fixed index blocks:

  [0,1,2], [3,4,5], [6,7,8], ...

  If a gap fell inside one of those blocks, the entire block was skipped, and the fixed indexing could also shift
  later compression groups.

  The new behavior is:

  - convert original epochs to UTC as before
  - use `doppler_integration_time` as the nominal cadence
  - split observations internally into contiguous cadence runs
  - compress complete `compressionRatio` blocks only within each run
  - drop only leftover observations at the end of each run
  - never average across a cadence gap.

  For example, with nominal cadence 10 s and compression ratio 3:

  0, 10, 20, 30, gap, 60, 70, 80, 90, 100, 110

  The new compression keeps:

  [0, 10, 20]
  [60, 70, 80]
  [90, 100, 110]

  and drops only the leftover 30.

  For gap-free data, the behavior is preserved: same compression grouping, same averaged values, same averaged
  epochs, and the same compressed integration time:

`  compressed_integration_time = nominal_integration_time * compressionRatio`


DominicDirkx:
@copilot resolve the merge conflicts in this pull request

include/tudat/simulation/estimation_setup/processOdfFile.h
```diff
@@ -1340,6 +1339,12 @@ std::shared_ptr< observation_models::SingleObservationSet< ObservationScalarType
         const std::shared_ptr< observation_models::SingleObservationSet< ObservationScalarType, TimeType > > originalDopplerData,
         const unsigned int compressionRatio )
 {
+    if( compressionRatio == 0 )
+    {
+        throw std::runtime_error( "Error in Doppler data compression, compression ratio must be positive." );
+    }
+
+    const double cadenceTolerance = 0.01;
```
DominicDirkx, line 1350: What is the rationale for this value, and why is any value needed?

include/tudat/simulation/estimation_setup/processTrackingTxtFile.h
```diff
@@ -466,23 +473,101 @@ class ProcessedTrackingTxtFileContents
 private:
     double getObservationTimeStep( )
     {
+        const double cadenceTolerance = 0.01;
```
DominicDirkx, line 476: What is the rationale for this value, and why is any value needed?

include/tudat/simulation/estimation_setup/processTrackingTxtFile.h
```diff
@@ -466,23 +473,101 @@ class ProcessedTrackingTxtFileContents
...
+        std::vector< double > observationTimeSteps;
+        observationTimeSteps.reserve( observationTimesUtc_.size( ) - 1 );
+        double firstObservationTimeStep = std::numeric_limits< double >::infinity( );
+        double minimumObservationTimeStep = std::numeric_limits< double >::infinity( );
+
         for( unsigned int i = 1; i < observationTimesUtc_.size( ); i++ )
         {
-            double testObservationTimeStep = observationTimesUtc_.at( i ) - observationTimesUtc_.at( i - 1 );
-            if( std::fabs( observationTimeStep - testObservationTimeStep ) >
-                0.01 )  // 50.0 * std::numeric_limits< double >::epsilon( ) * observationTimesUtc_.at( i - 1 )  )
```
Markus-Reichel, line 478: The tolerance was already used before, I just put it into a named constant.

include/tudat/simulation/estimation_setup/processOdfFile.h
Markus-Reichel: In short, its used to check the time delta of each observation to check for uniform cadence. A strict `==` will not work, as the IFMS time stamps have a millisecond time jitter.
In this file I updated the tolerance to 0.01s in January already.
Before that, a tolerance of `10 * eps * t` was used, which was leading to issues with IFMS files, as the tolerance was too tight.

DominicDirkx:
@Markus-Reichel On Slack, there was the proposal to determine the integration time before doing the filtering. I think this would be much less invasive, and a much smaller difference in code, than what is implemented here (but perhaps I am wrong). What do you think?

Markus-Reichel:
> @Markus-Reichel On Slack, there was the proposal to determine the integration time before doing the filtering. I think this would be much less invasive, and a much smaller difference in code, than what is implemented here (but perhaps I am wrong). What do you think?

Determining the integration time before filtering is a valid smaller fix. However, it only fixes cadence inference. If compressed Doppler data are used, we still need a "gap-aware" compression or splitting, otherwise compression can skip/shift blocks of observations across missing samples. The current implementation fixes both issues. The failure of determining the integration time and the compression behaviour. 
So yes, its more intrusive, but covers more of the actual downstream issue. 

Also, if we determine the integration time before doing the filtering, that would mean we would need to move cadence determination into the reader because:

IFMS filtering happens inside `readIfmsFile()` before `ProcessedTrackingTxtFileContents` is build. With `filterInvalidLines=true`, invalid rows are skipped in `TrackingTxtFileContents::addLineToRawDataMap()` and never rach `getObservationTimeStep()` (`include/tudat/io/readTrackingTxtFile.h:459, src/tudat/io/
  readTrackingTxtFile.cpp:74, include/tudat/simulation/estimation_setup/processTrackingTxtFile.h:353`). 
  
  So, yes, that would be possible, but we would need to carry that information downstream.

Markus-Reichel:
@DominicDirkx have a look. I moved the integration time determination before the filtering.
Does that look better to you?

DominicDirkx:
Nice :) Could you also resolve the merge conflict? Then the CI can run.

DominicDirkx:
@larshinueber FYI, in case you hadn't seen this

PR #828, opened 2026-06-09
Feature/ constrained multi-arc estimation support

Markus-Reichel:
**Summary**

 Adds an opt-in soft continuity constraint between consecutive translational arcs of a multi-arc orbit determination, following the Orbit14 formulation of Lari et al. (2021), "Orbit determination methods for interplanetary missions: development and use of the Orbit14 software", §3.3 Eq. (27)–(28).

 The feature is off by default (empty constraint list ⇒ zero runtime cost) and is fully exposed in Python.

 **Math**

 For each constrained boundary `(k_left, k_right)` at connection epoch
 `t_c`:
```

 d  = x_right(t_c) − x_left(t_c)                      (6-vector, physical units)
 D  = M_right(t_c) − M_left(t_c)                      (6 × N_params, full multi-arc layout)
 W_d = (1 / (μ · m_d)) · C

```
 with `C` a user-supplied 6×6 PSD weight matrix (selects which components are constrained: position-only, velocity-only, full state, anisotropic, or dense), `μ > 0` a global penalty scalar (larger μ ⇒ weaker penalty, per paper convention), and **m_d = Σ rank(C_pair)** the global rank sum.

 The per-pair contribution is injected directly into the normal equations:

```
 H_constraint += D_normᵀ W_d D_norm
 g_constraint += − D_normᵀ W_d d
 cost          += (1 / (μ · m_d)) · dᵀ C d
```

`D` is column-normalized with the same `normalizationTerms` already applied to the observation design matrix, so the constraint lives in the same numerical conditioning as the observations. `d` stays in physical units. The cost term is added to the least-squares cost driving best-iteration selection; residual RMS is unchanged.

The only requirement is that each side arc has been propagated through t_c.

**Typical Call:**
```

 from tudatpy.estimation import estimation_analysis as ea

 constraint = ea.position_only_continuity(
     body="Rosetta",
     epochs=ocm_epochs,              # one per OCM boundary
     position_weight=1.0 / sigma_r**2,
     mu=1.0,                          # global penalty scaling
 )
 estimation_input.set_inter_arc_continuity_constraints([constraint])
```

 **Backward compatibility**

 - All new POD-input / output members default-construct to empty.
 - All new LSQ optional arguments default to empty matrices. existing
 call sites are unchanged.
 - A null/missing constraint list short-circuits the feature with no
 multi-arc casts, STM lookups, or extra allocations.


DominicDirkx:
Missed this one :) Could you also resolve the merge conflict? Then the CI can run.

DominicDirkx:
@Markus-Reichel This is really cool functionality (and great that it's already working well for you!)

A few questions/comments:
* If I understand correctly, this functionality acts to modify the cost function during the least squares estimation. Is this correct? If so, I think having it in the estimation_input is teh right place (like the a priori covariance)
* Why does this functionality modify the state transition matrix interface? 
* I think the paper you're referring to is Lari et al. (2021), which use the same functionality for this as Cicalo et al. (2016), Alessi et al. (2012) and Cicalò & Milani (2012). Is this correct?

Markus-Reichel:
> @Markus-Reichel This is really cool functionality (and great that it's already working well for you!)
> 
> A few questions/comments:
> 
>     * If I understand correctly, this functionality acts to modify the cost function during the least squares estimation. Is this correct? If so, I think having it in the estimation_input is teh right place (like the a priori covariance)
> 
>     * Why does this functionality modify the state transition matrix interface?
> 
>     * I think the paper you're referring to is Lari et al. (2021), which use the same functionality for this as Cicalo et al. (2016), Alessi et al. (2012) and Cicalò & Milani (2012). Is this correct?

1. Yes, exactly. This is a soft least-squares constraint: it adds an extra normal-matrix contribution to the LSQ solve, and the iteration cost is changed with the inter-arc continuity cost. 
I moved the implementation to `estimation_input`, as of your recommendation.

2. The constraint needs explicit left-arc and right-arc partials at the same connection epoch. As i understand it, the previously existing time-based multi-arc STM lookup cannot express that at a shared boundary. For one epoch it only selects one arc.

3. Yes, my bad. I used Lari et al. (2021) Eq. 28 for the implementation, which is based on Alessi et al. (2012)

DominicDirkx:
Ah ok, thanks for clarifying! So the issue is that for a given epoch, the code now chooses one specific arc, but what you need is to have it return the state transition/sensitivity matrix of two arcs at the same epoch (?) 

Then indeed something needs to be added here, but in the current implementation, part of the code in getCombinedStateTransitionAndSensitivityMatrix is now duplicated in the new function. Could you modify it so that the actual setup of the result happens in a single place, and is then called from the 'regular' and 'per arc' option?

Markus-Reichel:
Thats exactly right. the state transition/sensitivity matrix is needed two arcs `i` and `i+1` at the connection point, so at the same epoch.

I added a shared helper logic for the multi-arc combined STM matrix assembly. Have a look and let me know if thats fine with you:)

include/tudat/astro/propagators/stateTransitionMatrixInterface.h
```diff
@@ -482,48 +482,12 @@ class MultiArcCombinedStateTransitionAndSensitivityMatrixInterface : public Comb
...
-                            sensitivityMatrixSize /*) : stateTransitionMatrixSize*/;
-                    combinedStateTransitionMatrix.block(
-                            statePartialAdditionIndices_.at( currentArc ).at( i ).first, 0, 6, indicesToAdd ) +=
-                            combinedStateTransitionMatrix.block(
-                                    statePartialAdditionIndices_.at( currentArc ).at( i ).second, 0, 6, indicesToAdd );
-                }
-            }
-        }
-        return combinedStateTransitionMatrix;
+        return Eigen::MatrixXd::Zero( 0, 0 );
```
DominicDirkx, line 476: why is this a good default?

include/tudat/astro/propagators/stateTransitionMatrixInterface.h
```diff
@@ -482,48 +482,12 @@ class MultiArcCombinedStateTransitionAndSensitivityMatrixInterface : public Comb
             }
         }
 
-        int stateTransitionMatrixSize = 0;
-        int sensitivityMatrixSize = 0;
-
         if( currentArc >= 0 )
```
DominicDirkx, line 471: What happens if current arc <0?

DominicDirkx:
@tudat-team/radio-science-dev Any thoughts on this?

Markus-Reichel:
> What happens if current arc <0?

`currentArc` is the index of the arc that contains `evaluationTime`. If `evaluationTime` falls inside one of the arcs, `currentArc is 0, 1, 2, …` and a real state-transition matrix is returned.

But if the time falls in a gap between arcs (or before the first / after the last arc), `getCurrentArc` returns `-1`,so no arc covers this time. In that case the function has nothing meaningful to compute, so it returns the empty 0×0 matrix, which just means "there is nothing to compute".


> why is this a good default?

The size of a real matrix here depends on _which_ arc is looked into. Different arcs could estimate different parameters, so they could have different sizes (`arcWiseStateTransitionMatrixSize_[currentArc]`). When no arc applies, there would be no "correct" size to pic, so it just returns `0x0`.



@DominicDirkx Do you have suggestions how to better handle this?

Markus-Reichel:
One thing that I noticed, when an observation time falls in a gap between arcs, the estimation gets a zero matrix back and that observation simply contributes nothing, with no warning. So if the arc boundaries are set up slightly wrong, observations could be silently dropped. 
The question is, should that be up to the user to properly verify the arc boundaries, or should a warning be printed?
I guess it can be deliberate to exclude certain observations and then throwing an error would be kinda annoying, as you intentionally omit them.

include/tudat/astro/orbit_determination/podInputOutputTypes.h
```diff
@@ -750,12 +774,16 @@ struct CovarianceAnalysisOutput {
                               const Eigen::VectorXd& considerNormalizationFactors = Eigen::VectorXd::Zero( 0 ),
                               const Eigen::MatrixXd& considerCovarianceContribution = Eigen::MatrixXd::Zero( 0, 0 ),
                               const Eigen::MatrixXd& considerCovariance = Eigen::MatrixXd::Zero( 0, 0 ),
+                              const double interArcContinuityCost = 0.0,
+                              const std::vector< Eigen::Matrix< double, 6, 1 > >& interArcContinuityDiscrepancies =
```
DominicDirkx: This should not be a Vector6d, it should be generalized for any state type and number of bodies

include/tudat/simulation/estimation_setup/interArcContinuityConstraint.h
```diff
@@ -0,0 +1,482 @@
...
+    {
+        throw std::runtime_error( "Error in assembleInterArcContinuityContribution: full STM column count (" +
+                                  std::to_string( fullStmCols ) + ") does not match LSQ parameter size (" +
+                                  std::to_string( totalParameterSize ) +
+                                  "). Inter-arc continuity priors currently require a pure multi-arc parameter set "
+                                  "(spec section 9: hybrid-arc out of scope for v1)." );
+    }
+
+    // First pass: compute the global m_d_total = sum over every (settings, pair) of rank(C_pair).
+    int mDTotal = 0;
```
DominicDirkx: could you modify this to have a variable name that is descriptive (not mathematical symbols?)

include/tudat/simulation/estimation_setup/interArcContinuityConstraint.h
```diff
@@ -0,0 +1,482 @@
...
+    for( const auto& settings : constraintSettings )
+    {
+        for( std::size_t pairIndex = 0; pairIndex < settings->numberOfPairs( ); ++pairIndex )
+        {
+            mDTotal += detail::rankOf6x6PsdMatrix( settings->weightMatrixForPair( pairIndex ) );
+        }
+    }
+    if( mDTotal == 0 )
```
DominicDirkx: If the check is whether C is non-zero, why the calls to rankOf6x6PsdMatrix? What exactly is this part suppose to do?

include/tudat/simulation/estimation_setup/interArcContinuityConstraint.h
```diff
@@ -0,0 +1,482 @@
...
+
+    // Second pass: build D and accumulate contributions per (settings, pair).
+    for( const auto& settings : constraintSettings )
+    {
+        // Locate the body's arc-wise translational state parameter.
+        std::shared_ptr< estimatable_parameters::EstimatableParameter< Eigen::Matrix< ObservationScalarType, Eigen::Dynamic, 1 > > >
+                bodyParameter;
+        for( const auto& candidate : multiArcStateParameters )
+        {
+            if( candidate->getParameterName( ).second.first == settings->body( ) )
```
DominicDirkx: This will match *any* multi-arc state type (which is not how this is implemented, it requires translational?) It is also only valid for a single propagated body

include/tudat/simulation/estimation_setup/interArcContinuityConstraint.h
```diff
@@ -0,0 +1,482 @@
...
+            {
+                bodyParameter = candidate;
+                break;
```
DominicDirkx: Should also check if multiple parameters match (if so, an exception?)

include/tudat/simulation/estimation_setup/interArcContinuityConstraint.h
```diff
@@ -0,0 +1,482 @@
...
+            }
+        }
+        if( bodyParameter == nullptr )
+        {
+            throw std::runtime_error( "Error in assembleInterArcContinuityContribution: body \"" + settings->body( ) +
+                                      "\" referenced by an inter-arc continuity prior does not have an arc-wise "
+                                      "translational initial state parameter." );
+        }
+
+        const int nArcs = static_cast< int >( bodyParameter->getParameterSize( ) ) / 6;
```
DominicDirkx: hard-coded for single-body translational state

include/tudat/simulation/estimation_setup/interArcContinuityConstraint.h
```diff
@@ -0,0 +1,482 @@
...
+
+        const auto arcPairs = settings->arcPairs( );
+        const auto& epochs = settings->connectionEpochs( );
+
+        for( std::size_t pairIndex = 0; pairIndex < epochs.size( ); ++pairIndex )
```
DominicDirkx: Please add comments to describe the logic/math

include/tudat/simulation/estimation_setup/interArcContinuityConstraint.h
```diff
@@ -0,0 +1,482 @@
...
+            if( mLeft.rows( ) != stmInterface->getFullStateSize( ) || mLeft.cols( ) != totalParameterSize ||
+                mRight.rows( ) != stmInterface->getFullStateSize( ) || mRight.cols( ) != totalParameterSize )
+            {
+                throw std::runtime_error(
+                        "Error in assembleInterArcContinuityContribution: per-arc STM has unexpected "
+                        "shape (rows=" +
+                        std::to_string( mLeft.rows( ) ) + ", cols=" + std::to_string( mLeft.cols( ) ) + "), expected (" +
+                        std::to_string( stmInterface->getFullStateSize( ) ) + ", " + std::to_string( totalParameterSize ) + ")." );
+            }
+            Eigen::MatrixXd D = mRight.block( rightBodyRows.fullStateStart, 0, 6, totalParameterSize ) -
```
DominicDirkx: please use descriptive variable name (same in entire PR)

include/tudat/simulation/estimation_setup/orbitDeterminationManagerEstimationImplementation.h
```diff
@@ -125,6 +131,20 @@ OrbitDeterminationManager< ObservationScalarType, TimeType, Dummy >::estimatePar
         Eigen::MatrixXd normalizedInverseAprioriCovarianceMatrix = normalizeAprioriCovariance(
                 estimationInput->getInverseOfAprioriCovariance( numberEstimatedParameters_ ), normalizationTerms );
 
+        // Assemble soft inter-arc continuity-prior contribution for this iteration. The normalisation factors are the same
+        // ones just applied to the observation design matrix above.
+        InterArcConstraintContribution interArcContribution =
```
DominicDirkx: Is now always done (?) Code should check if any constraint is to be applied

include/tudat/simulation/estimation_setup/orbitDeterminationManagerCovarianceImplementation.h
```diff
@@ -87,6 +88,22 @@ OrbitDeterminationManager< ObservationScalarType, TimeType, Dummy >::computeCova
             constraintRightHandSide,
             estimationInput->getLimitConditionNumberForWarning( ) );
 
+    // Add the soft inter-arc continuity-prior normal-matrix contribution, when configured.
```
DominicDirkx: Is now always done (?) Code should check if any constraint is to be applied

src/tudatpy/estimation/estimation_analysis/expose_estimation_analysis.cpp
```diff
@@ -546,6 +546,28 @@ containing the data, see `user guide description <https://docs.tudat.space/en/la
...
+                  &tss::CovarianceAnalysisInput< STATE_SCALAR_TYPE, TIME_TYPE >::setInterArcContinuityConstraints,
+                  py::arg( "constraints" ),
+                  R"doc(
+
+         Attach soft inter-arc translational state continuity priors.
+
+         These priors add a regularizing normal-equation contribution
+         :math:`\mathbf{D}_\mathrm{norm}^{T}\mathbf{W}_{d}\mathbf{D}_\mathrm{norm}` in covariance analysis and,
+         during estimation, the matching right-hand-side contribution
+         :math:`-\mathbf{D}_\mathrm{norm}^{T}\mathbf{W}_{d}\mathbf{d}`. ``D_norm`` is the right-minus-left
```
DominicDirkx: This part should be more explicit on *exactly* what everything is

src/tudatpy/estimation/estimation_analysis/expose_estimation_analysis.cpp
```diff
@@ -546,6 +546,28 @@ containing the data, see `user guide description <https://docs.tudat.space/en/la
...
+         state-transition/sensitivity block after applying the estimator's column normalization. They are currently
+         supported only for pure multi-arc translational estimators. Pass an empty list to disable the feature.
+      )doc" )
```
DominicDirkx, line 570: The change in the covariance and parameter correction due to this new setting should be added to the docs of compute_covariance and perform_estimation

DominicDirkx:
@Markus-Reichel I've added a bunch of questions/comments to the code. I think it needs a small iteration, but it'll be really cool to add this to Tudat!

include/tudat/astro/orbit_determination/podInputOutputTypes.h
Markus-Reichel: So currently one constraint settings object is required per body. For example, in Python it looks like this:

  ```python
  estimation_input.set_inter_arc_continuity_constraints([
      estimation_analysis.position_only_continuity("Earth", epochs, position_weight=[...], mu=...),
      estimation_analysis.position_only_continuity("Mars", epochs, position_weight=[...], mu=...),
  ])
```

  So constraining multiple bodies is already supported. Here, mu is the scalar constraint scaling factor. I will also rename the internal variables/parameters to make this clearer.

  Do I understand correctly that you would prefer a more general design that can cover arbitrary state block sizes, for example multiple propagated bodies at the same arc boundary? For two translational bodies, full-state continuity would mean one discrepancy vector of size 12:

 ```
 d = [
      dx_Earth
      dy_Earth
      dz_Earth
      dvx_Earth
      dvy_Earth
      dvz_Earth
      dx_Mars
      dy_Mars
      dz_Mars
      dvx_Mars
      dvy_Mars
      dvz_Mars
  ]

```
  with a matching 12x12 constraint weight matrix C, instead of two independent 6x6 constraints.

  The cost term would then be:

```
  q = (1 / (mu * m_d)) *
      [d_Earth^T  d_Mars^T]
      C
      [d_Earth
       d_Mars]
```

  where mu is the scalar constraint scaling factor, and m_d is the number of constrained dimensions (rank(C)).

  If C is block diagonal,

```
  C = [ C_Earth   0
        0         C_Mars ]
```

  then the same global mu scales both body blocks, while the relative weighting between bodies is set through C_Earth and C_Mars.

  If we want users to choose a separate mu per body, then this is no longer a single combined constraint with one scalar mu.
  it becomes either:

  - independent per-body constraint contributions, which is close to the current design, or
  - a combined block weight matrix where the per-body scaling has already been absorbed into the individual blocks.

include/tudat/simulation/estimation_setup/orbitDeterminationManagerEstimationImplementation.h
Markus-Reichel: with no constraints, the contribution is empty/zero and the normal equations are not changed.
But you are correct, I will add another check at the call site

include/tudat/astro/orbit_determination/podInputOutputTypes.h
Markus-Reichel: use global mu

include/tudat/simulation/estimation_setup/interArcContinuityConstraint.h
Markus-Reichel: Mutli-Body support is now implemented, but at the moment only translational state type is supported, as the mutli-arc estimation only supports translational states

include/tudat/simulation/estimation_setup/interArcContinuityConstraint.h
Markus-Reichel: Now counts all matching **arc_wise_initial_body_state** parameters for the constrained body and throws if more than one exists

include/tudat/simulation/estimation_setup/interArcContinuityConstraint.h
Markus-Reichel: Now only supports translational state (can be extended to support other states in the future)

include/tudat/simulation/estimation_setup/interArcContinuityConstraint.h
Markus-Reichel: The rank is used to count how many state components are actually constrained by C, not just to check whether C is non-zero.

For example:

- position-only continuity has rank 3
- velocity-only continuity has rank 3
- full position+velocity continuity has rank 6
- a rank-one weight matrix constrains one linear combination

This count is then used in the scaling of the constraint weight (as is done in the Lari et al. paper). The zero-rank case is just the invalid case where C does not constrain anything.

Markus-Reichel:
@DominicDirkx I addressed all of your and copilots comments. 
I added a couple of replies to some of your comments. have a look if you're now fine with the implementation :)

include/tudat/simulation/estimation_setup/interArcContinuityConstraint.h
DominicDirkx: NIce!

include/tudat/simulation/estimation_setup/interArcContinuityConstraint.h
```diff
@@ -0,0 +1,568 @@
...
+    if( solver.info( ) != Eigen::Success )
+    {
+        throw std::runtime_error( "Error in computePsdMatrixRank: eigen-decomposition failed." );
+    }
+    const double largest = std::max( solver.eigenvalues( ).maxCoeff( ), 0.0 );
+    if( largest <= 0.0 )
+    {
+        return 0;
+    }
+    const double threshold = 1.0E-12 * largest;
```
DominicDirkx: why this specific threshold?

include/tudat/simulation/estimation_setup/interArcContinuityConstraint.h
```diff
@@ -0,0 +1,568 @@
...
+
+//! Aggregated normal-equation contribution of all soft inter-arc continuity priors.
+struct InterArcConstraintContribution {
+    Eigen::MatrixXd additionalNormalMatrix;
+    Eigen::VectorXd additionalRightHandSide;
+    double totalConstraintCost = 0.0;
+    std::vector< Eigen::VectorXd > perPairDiscrepancies;
+};
+
+namespace detail
```
DominicDirkx: Don't use the 'detail' namespaces

include/tudat/simulation/estimation_setup/interArcContinuityConstraint.h
```diff
@@ -0,0 +1,568 @@
...
+
+    // First pass: compute the total constraint dimension from the rank of every active weight matrix.
+    int totalConstraintDimension = 0;
+    for( const auto& settings : constraintSettings )
+    {
+        for( const auto& body : settings->bodies( ) )
+        {
+            for( std::size_t pairIndex = 0; pairIndex < settings->numberOfPairsForBody( body ); ++pairIndex )
+            {
+                totalConstraintDimension += detail::computePsdMatrixRank( settings->weightMatrixForBodyAndPair( body, pairIndex ) );
```
DominicDirkx: I don't understand this step. Is there a link to an equation in the source paper?

include/tudat/simulation/estimation_setup/interArcContinuityConstraint.h
```diff
@@ -0,0 +1,568 @@
...
+//! adds the resulting soft prior to the normal equations. Per Lari et al. (2021) Eq. 28, this is:
+//!   stateDiscrepancy = rightArcState(connectionEpoch) - leftArcState(connectionEpoch)
+//!   continuityDesignMatrix = rightArcVariationalRows - leftArcVariationalRows
+//!   constraintWeight = constraintWeightMatrix / (constraintScalingFactor * totalConstraintDimension)
+//!   normal += continuityDesignMatrix^T * constraintWeight * continuityDesignMatrix
+//!   rhs    -= continuityDesignMatrix^T * constraintWeight * stateDiscrepancy
+//! where totalConstraintDimension is the rank sum across every settings entry passed in. Column-normalization of
+//! continuityDesignMatrix uses the same factors that the OD loop applies to the observation design matrix.
+template< typename ObservationScalarType, typename TimeType >
+InterArcConstraintContribution assembleInterArcContinuityContribution(
```
DominicDirkx, line 378: This one function is very long. Can it be split up?

include/tudat/simulation/estimation_setup/interArcStateContinuityConstraintSettings.h
```diff
@@ -0,0 +1,189 @@
...
+    EpochMap connectionEpochsByBody_;
+    WeightMatrixMap weightMatricesByBody_;
+    double constraintScalingFactor_;
+    ArcPairMap arcPairsByBody_;
+};
+
+namespace detail
+{
+
+inline Eigen::MatrixXd diagonalWeight( double position, double velocity )
```
DominicDirkx: Rename after removing from detail namespace, also modify name of inputs (they are not a position or velocity)

src/tudatpy/estimation/estimation_analysis/expose_estimation_analysis.cpp
```diff
@@ -879,6 +926,24 @@ containing the data, see `user guide description <https://docs.tudat.space/en/la
...
 
          :type: numpy.ndarray[numpy.float64[m, 1]]
+      )doc" )
+            .def_property_readonly( "inter_arc_continuity_cost",
+                                    &tss::CovarianceAnalysisOutput< STATE_SCALAR_TYPE, TIME_TYPE >::getInterArcContinuityCost,
+                                    R"doc(
+
+         **read-only**
+
+         Soft inter-arc continuity-prior cost :math:`\sum d^T W_d d` at the covariance-analysis
```
DominicDirkx: Make explicit how the input relates to the math, and refer to place where the math is explained in more detail (same for all docstrings)

src/tudatpy/estimation/estimation_analysis/expose_estimation_analysis.cpp
```diff
@@ -879,6 +926,24 @@ containing the data, see `user guide description <https://docs.tudat.space/en/la
...
+         linearization point. This value is zero when no inter-arc continuity priors were attached.
```
DominicDirkx: Type should be specified (for all docs)

src/tudatpy/estimation/estimation_analysis/expose_inter_arc_constraints.cpp
```diff
@@ -0,0 +1,293 @@
...
+#include "tudat/simulation/estimation_setup/interArcStateContinuityConstraintSettings.h"
+
+namespace py = pybind11;
+namespace tss = tudat::simulation_setup;
+
+namespace
+{
+
+//! Build a 6x6 diagonal weight matrix from either a scalar or a length-3 sequence of per-component weights.
+Eigen::MatrixXd buildDiagonalWeight( const py::object& positionWeight, const py::object& velocityWeight )
```
DominicDirkx: It's not clear why these are here in the exposure rather than in the code, nor why the py::object is needed

src/tudatpy/estimation/estimation_analysis/expose_estimation_analysis.cpp
```diff
@@ -546,6 +546,28 @@ containing the data, see `user guide description <https://docs.tudat.space/en/la
          A-priori covariance matrix of the considered parameters.
 
          :type: numpy.ndarray[numpy.float64[n, n]]
+      )doc" )
+            .def( "set_inter_arc_continuity_constraints",
+                  &tss::CovarianceAnalysisInput< STATE_SCALAR_TYPE, TIME_TYPE >::setInterArcContinuityConstraints,
```
DominicDirkx, line 551: The definition of the constraints input should be made explicit in the docs (with symbol) and how it links to the equations that are provided should be made unambiguous. For each symbol used it should be fuly clear what it is, and how it relates to other functions in the API

PR #835, opened 2026-06-17
Fix TRK2-34 open ramp tables

larshinueber:
#805 introduced a check that the end time of a ramp table entry is after the start time of the ramp table. This causes issues for TNF files, as some ground stations only have starting records with no corresponding end record.
In this case, the end time is `None` and gets implicitly converted a 0 posix epoch, which now throws an error with the new behavior. 

This PR
- adds a check for open ramp intervals, depending on the user choice lets them close it artificially, print a warning or throw an exception
- checks if a frequency interpolator already exists for the ground station, if so, adds the new one to the existing one, instead of overwriting it

Instead of using an arbitrary epoch I tried to retrieve the epoch of the last SFDU in the processed TNF files, but couldn't get this to work reliably in the MRO example.

### DIFF
--- a/docs/tudatpy/source/data/processTrk234.rst
+++ b/docs/tudatpy/source/data/processTrk234.rst
@@ -8,6 +8,17 @@ This module provides tools to read, parse, and convert DSN radiometric tracking 
 (Doppler, range, and ramp data) from TNF/TRK files into Tudat observation formats.
 
 
+Enums
+-----
+.. currentmodule:: tudatpy.data.processTrk234
+
+.. autosummary::
+
+   converters.OpenRampHandling
+
+.. autoclass:: tudatpy.data.processTrk234.converters.OpenRampHandling
+   :members:
+
 Classes
 -------
 .. currentmodule:: tudatpy.data.processTrk234
--- a/include/tudat/astro/ground_stations/transmittingFrequencies.h
+++ b/include/tudat/astro/ground_stations/transmittingFrequencies.h
@@ -360,6 +360,12 @@ public:
         return startFrequencies_;
     }
 
+    //! Function to retrieve the gap handling option
+    FrequencyGapHandling getGapHandling( )
+    {
+        return gapHandling_;
+    }
+
     void addFrequencyInterpolator( const std::shared_ptr< PiecewiseLinearFrequencyInterpolator > rampsToAdd );
 
 private:
--- a/src/tudat/astro/ground_stations/transmittingFrequencies.cpp
+++ b/src/tudat/astro/ground_stations/transmittingFrequencies.cpp
@@ -132,6 +132,12 @@ void PiecewiseLinearFrequencyInterpolator::initialize( )
 void PiecewiseLinearFrequencyInterpolator::addFrequencyInterpolator(
         const std::shared_ptr< PiecewiseLinearFrequencyInterpolator > rampsToAdd )
 {
+    if( rampsToAdd->getGapHandling( ) != gapHandling_ )
+    {
+        std::cerr << "Warning: gap handling of ramp table to add (" << rampsToAdd->getGapHandling( )
+                  << ") differs from base ramp table gap handling (" << gapHandling_ << "). Using base gap handling." << std::endl;
+    }
+
     std::vector< Time > startTimesToAdd = rampsToAdd->getStartTimes( );
     std::vector< Time > endTimesToAdd = rampsToAdd->getEndTimes( );
     std::vector< double > rampRatesToAdd = rampsToAdd->getRampRates( );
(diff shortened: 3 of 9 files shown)


valeriof7:
I have noticed several times that the required ramp tables are included in files with earlier dates. Therefore, I would rather flag the measurements and warn the user than extrapolate and suggest loading another tracking file to get the correct ramp table. At the very least, we must inform the users and let them choose whether to extrapolate.

larshinueber:
Hi @valeriof7, thanks for the comments!
Maybe to clarify, the issue is that the TNF file provides a starting event (ramp event type 1) but no closing record event (event type 4/5). If you have subsequent TNF files that can be fine, as we implicitly close the previous ramp in that case, but if you dont have any subsequent files with the same ground station this ramp remains open (with the ramp end time being `None`, that gets converted to epoch 1970, which then throws an error in the frequency interpolator).

> At the very least, we must inform the users and let them choose whether to extrapolate.

Sounds good, we could make this an input option in the function 👍

DominicDirkx:
@larshinueber If I read this discussion correctly, a small addition will be made to this before merging right?

larshinueber:
This is ready for review now!
@DominicDirkx as discussed, for now I added options to throw an exception on open ramps, print a warning (once) or just close the interval artificially. I briefly tried to add an option to use the latest timestamp from the TNF file but couldnt get it to work reliably on all MRO TNF files from the example. I'll open an issue for that

DominicDirkx:
You recompiled the kernel and reran the python tests?

larshinueber:
@DominicDirkx Yes, all Python tests are passing 👍 
```
~/r/tudatpy fix/trk-open-ramp-tables *1 !1 ❯ pytest   19m 46s  tudatpy-dev
=========================== test session starts ============================
platform darwin -- Python 3.11.15, pytest-9.0.3, pluggy-1.6.0
rootdir: /Users/lhinuber/repos/tudatpy
configfile: pyproject.toml
collected 222 items                                                        

tests/test_tudatpy/test_constants.py .                               [  0%]
tests/test_tudatpy/test_data.py ...                                  [  1%]
tests/test_tudatpy/test_data_biases.py .....                         [  4%]
tests/test_tudatpy/test_data_horizons.py ........................... [ 16%]
..........................................                           [ 35%]
tests/test_tudatpy/test_data_mpc.py .........                        [ 39%]
tests/test_tudatpy/test_data_parser_80col.py ....................... [ 49%]
...................                                                  [ 58%]
tests/test_tudatpy/test_data_weights.py ........                     [ 61%]
tests/test_tudatpy/test_dependent_variable_dictionary.py .           [ 62%]
tests/test_tudatpy/test_discos.py ....                               [ 63%]
tests/test_tudatpy/test_dse_functions.py .                           [ 64%]
tests/test_tudatpy/test_hybrid_module_exposure.py .                  [ 64%]
tests/test_tudatpy/test_interface_spice.py ..                        [ 65%]
tests/test_tudatpy/test_nrlmsise00.py .                              [ 66%]
tests/test_tudatpy/test_processTrk234.py ...........                                    [ 71%]
tests/test_tudatpy/test_spacetrack.py ..........ss                                      [ 76%]
tests/test_tudatpy/test_time_conversions.py ........................ [ 87%]
............................                                                            [100%]

====================================== warnings summary =======================================
tests/test_tudatpy/test_data_biases.py: 5 warnings
tests/test_tudatpy/test_data_weights.py: 8 warnings
  /Users/lhinuber/miniconda3/envs/tudatpy-dev/lib/python3.11/site-packages/tudatpy/data/mpc/mpc.py:140: PerformanceWarning: DataFrame is highly fragmented.  This is usually the result of calling `frame.insert` many times, which has poor performance.  Consider joining all columns at once using pd.concat(axis=1) instead. To get a de-fragmented frame, use `newframe = frame.copy()`
    bias_dataframe[["aa", "bb", "cc", "dd"]] = unknown_columns

tests/test_tudatpy/test_data_mpc.py: 9 warnings
tests/test_tudatpy/test_data_parser_80col.py: 3 warnings
tests/test_tudatpy/test_data_weights.py: 8 warnings
  /Users/lhinuber/miniconda3/envs/tudatpy-dev/lib/python3.11/site-packages/tudatpy/data/mpc/mpc.py:988: Pandas4Warning: 'future.no_silent_downcasting' is deprecated, please refrain from using it.
    pd.set_option("future.no_silent_downcasting", True)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
======================== 220 passed, 2 skipped, 33 warnings in 49.69s =========================
~/r/tudatpy fix/trk-open-ramp-tables *1 !1 ❯ 
```

DominicDirkx:
Great!

PR #836, opened 2026-06-18
DSN custom seasonal troposphere calibration input

larshinueber:
This PR adds the option to load a custom seasonal troposphere calibration file from the python interface. The current default troposphere seasonal correction model only contains the trigonometric coefficients, while the default file from the PDS (https://pds-geosciences.wustl.edu/radiosciencedocs/urn-nasa-pds-jpl_dsn_mmm/tro/1972_001_2048_001_tro.csp) contains additional constant terms per single DSN station. The constant corrections are in the order of cm to mm for all active stations, for backwards compatibility the default behavior is unchanged.

The current implementation checks that only one adjustment is added per time window. This check is removed, since the constant corrections are specified per time window and are added on top of the trigonometric corrections, therefore multiple corrections are active at the same time and summed up.

### DIFF
--- a/include/tudat/astro/observation_models/corrections/atmosphereCorrection.h
+++ b/include/tudat/astro/observation_models/corrections/atmosphereCorrection.h
@@ -237,7 +237,7 @@ public:
     /*!
      * Constructor.
      */
-    TabulatedMediaReferenceCorrectionManager( ): isLookupSchemeUpdated_( false ) {}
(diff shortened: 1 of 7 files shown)


DominicDirkx:
Nice! Thanks for adding. Could you also add a (small) unit test for this fuctionality?

DominicDirkx:
Nice!

PR #839, opened 2026-06-21
feature/ Expose panel material parameters

Markus-Reichel:
Adds estimation support for panel material properties beyond specular/diffuse reflectivity.

  Changes:

  - Adds estimatable parameters for:
     - energy_accomodation_coefficient
      - normal_accomodation_coefficient
      - tangential_accomodation_coefficient
      - normal_velocity_at_wall_ratio

  - Reuses the existing panel-group estimation pattern used by specular/diffuse reflectivity.
  - Adds Python parameters_setup bindings for the new parameter settings.
  - Extends aerodynamic acceleration partials to compute sensitivities via central finite difference on the material parameter value.


DominicDirkx:
Nice to have this functionality! I'm guessing it's still in draft because unit tests are missing (?) 

For the partials, i think these can be analytical rather than numerical, to make sure they're robust in any regime

Markus-Reichel:
exactly. I just did a quick implementation but still need to check if its working properly. I am currently running the first tests!

But one thing that I noticed, the macro-model is reeeaaalllly slow, especially when self-shadowing is enabled.
Maybe I will have an AI look over the implementation to see if some code changes could speed it up a bit.

DominicDirkx:
Indeed, the macromodel is very slow, and scales quadratically with the number of panels. I think you could probably simplify your model to 10-20 panels, how many are there now? Also, you can play with the number of pixels/rays to make it faster. Of course, suggestions to improve computational efficiency would be much appreciated (but let's do this in a different issue)!

PR #840, opened 2026-06-21
Add offline mode and listing cache to mission data downloader

Markus-Reichel:
LoadPDS previously hit the network on every run to list archive directories, even when all files were already on disk: repeated periods were slow and offline use failed outright.

Add a JSON listing cache (<local_folder>/.mission_data_cache.json) and route all directory listings through a single _fetch_listing() helper:
- cache-first: reuse a resolved listing with no HTTP call
- offline=True: never touch the network (use cache + on-disk files)
- auto-fallback: on a network error, use the cache if available

### DIFF
--- a/src/tudatpy/data/mission_data_downloader/mission_data_downloader.py
+++ b/src/tudatpy/data/mission_data_downloader/mission_data_downloader.py
@@ -570,7 +570,24 @@ class LoadPDS:
 
         try:
             reqs = requests.get(url, timeout=_REQUEST_TIMEOUT)
-            reqs.raise_for_status()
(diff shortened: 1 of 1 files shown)


DominicDirkx:
@luigigisolfi Any thoughts on this?

DominicDirkx:
@Markus-Reichel @luigigisolfi  I'm closing this PR for now. I think we will need to do a thorough redesign of the mission data downloader, and the changes proposed here can inform this redesign. But, I don't think it's worth the time to do a thorough review of this rather large PR only to have it be completely redesigned soon(ish)

PR #841, opened 2026-06-21
Feature/custom number density atmosphere

Markus-Reichel:
Summary

  Adds a CustomNumberDensityAtmosphere model that lets users define an atmosphere from a user-supplied number density function, rather than a mass density function.

PR #848, opened 2026-06-24
fix/nrlmsise00-override-warnings

Markus-Reichel:
Fixing compiler warnings.


DominicDirkx:
@Markus-Reichel Any idea why these corrections are needed in this atmosphere model but not others?

Markus-Reichel:
It was only a warning caused by an inconsistency in the style in `NRLMSISE00Atmosphere`.
In PR #841, I added `getTotalNumberDensity` to several atmosphere models. For the new custom number density atmosphere model, I used the modern style and marked the overridden functions with `override`. When I added `getTotalNumberDensity` to the `NRLMSISE00Atmosphere` model, I copied that function including override.
The already existing functions however didnt use `override`. So that classes mixed both styles, triggering the warning.

larshinueber:
Thanks for fixing, I got the same warnings just now :)

PR #850, opened 2026-06-24
Fix/ NaN in aerodynamic partials when a wind model is present

Markus-Reichel:
**The bug**

Running an estimation with drag + a wind model crashed with:

`Error: Lagrange interpolator cannot identify zero entry.`

The NaN came from this line in `propagationOutput.h`, which computes the wind velocity in the output frame:

`return rotationMatrixFunction() * windVelocityCorotatingFunction();`

Eigen returns this product as an expression that points at the two temporary values from the function calls. Those temporaries die when the line returns, so the result reads dead memory → NaN. 

**The fix**

Store the result in a real vector before returning, so nothing points at dead memory:

```
Eigen::Vector3d transformedWindVelocity =
    rotationMatrixFunction() * windVelocityCorotatingFunction();
return transformedWindVelocity;
```

**Additions**

- Some additional clean-up for coma model 
- using `getAerodynamicAngleCalculator() ` in `propagationOutput.h` for readability

### DIFF
--- a/include/tudat/astro/aerodynamics/comaModel.h
+++ b/include/tudat/astro/aerodynamics/comaModel.h
@@ -170,10 +170,10 @@ private:
     // ========== Hot path: Frequently accessed cached values (grouped for cache locality) ==========
 
     //! Cached solar longitude value to avoid redundant state function calls [rad]
-    double cachedSolarLongitude_;
+    mutable double cachedSolarLongitude_;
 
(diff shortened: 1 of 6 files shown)


DominicDirkx:
@Markus-Reichel Looks good to me! If you could run the auto-formatting (you can also set it up to do this automatically), I think this culd be merged!

PR #852, opened 2026-06-25
Save partials as dependent variables

### LINKED ISSUE #674: [F&E]: Jacobian of the state time derivative
### Description:

Tudatpy provides functionality to compute the state transition matrix which requires the Jacobian of the state time derivative WRT the states. However, tudatpy does not expose the state time derivative Jacobian to the user. A function to compute the state time derivative Jacobian would be nice to have too!

_No response_

### Additional context:

_No response_

DominicDirkx:
Closes #674

### DIFF
--- a/tests/test_tudat/src/astro/propagators/unitTestDependentVariableOutput.cpp
+++ b/tests/test_tudat/src/astro/propagators/unitTestDependentVariableOutput.cpp
@@ -1345,6 +1345,233 @@ BOOST_AUTO_TEST_CASE( test_AccelerationPartialSaving )
     }
 }
 
+//! Unit test to check if acceleration derivatives w.r.t. estimatable parameters are correctly stored.
+BOOST_AUTO_TEST_CASE( test_AccelerationParameterPartialSaving )
+{
+    spice_interface::loadStandardSpiceKernels( );
+
+    std::vector< std::string > bodyNames;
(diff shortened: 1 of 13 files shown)


tests/test_tudat/src/astro/propagators/unitTestDependentVariableOutput.cpp
```diff
@@ -1345,6 +1345,233 @@ BOOST_AUTO_TEST_CASE( test_AccelerationPartialSaving )
...
+        const Eigen::MatrixXd midSensitivity = sensitivitySolution.at( dependentVariableIteratorMid->first );
+
+        Eigen::MatrixXd earthCosinePartial;
+        getOutputVectorInMatrixRepresentation(
+                dependentVariableIteratorMid->second.segment( earthCosineCoefficientPartialStart, earthCosineCoefficientPartialSize ),
+                earthCosinePartial,
+                3,
+                earthCosineParameterIndices.at( 0 ).first.second );
+        Eigen::MatrixXd expectedEarthCosinePartial =
+                sensitivityDerivative.block( 3, earthCosineSensitivityStart, 3, earthCosineParameterIndices.at( 0 ).first.second ) -
```
DominicDirkx: This is a poor method of testing. The StateDerivativeModel should be updated and the partial object queried to retrieve the real value

PR #854, opened 2026-06-25
Moved transponder delay to vehicle systems

### LINKED ISSUE #477: Add transponder delay via VehicleSystem object
PR tudat-team/tudat#268 introduced the possibility to set the transponder delay with the new method of the `ObservationCollection` object named `setTransponderDelay` (see `include/tudat/simulation/estimation_setup/observations.h`). However, since this is a spacecraft attribute, it makes sense to set it via the `VehicleSystem` object with a method like the already implemented `setTransponderTurnaroundRatio` (see `include/tudat/astro/system_models/vehicleSystems.h`).

DominicDirkx:
Closes #477

### DIFF
--- a/include/tudat/astro/system_models/vehicleSystems.h
+++ b/include/tudat/astro/system_models/vehicleSystems.h
@@ -45,7 +45,9 @@ public:
      * Constructor
      * \param dryMass Total dry mass of the vehicle (not defined; NaN by default).
      */
-    VehicleSystems( const double dryMass = TUDAT_NAN ): currentOrientationTime_( TUDAT_NAN ), dryMass_( dryMass ) {}
+    VehicleSystems( const double dryMass = TUDAT_NAN ):
+        currentOrientationTime_( TUDAT_NAN ), dryMass_( dryMass ), transponderDelay_( TUDAT_NAN )
+    {}
 
     //! Destructor
     ~VehicleSystems( ) {}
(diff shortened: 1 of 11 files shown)


include/tudat/astro/system_models/vehicleSystems.h
```diff
@@ -352,6 +354,25 @@ class VehicleSystems
...
     }
 
+    void setTransponderDelay( const double transponderDelay )
+    {
+        transponderDelay_ = transponderDelay;
+    }
+
+    bool isTransponderDelayDefined( ) const
+    {
+        return !( transponderDelay_ != transponderDelay_ );
```
larshinueber: From my limited understanding of C++, couldnt this also be turned into `transponderDelay_ == transponderDelay_` or `!std::isnan(transponderDelay_)`?

src/tudatpy/estimation/observations/expose_observations.cpp
```diff
@@ -1584,23 +1584,6 @@ residuals_per_parser : dict[ObservationCollectionParser, np.ndarray]
              Link end type to which the reference point should be applied.
          observation_parser : tudatpy.estimation.observations.observations_processing.ObservationCollectionParser, optional
              Parser to select the observation sets.
-     )doc" )
-            .def( "set_transponder_delay",
```
larshinueber, line 1588: Removing this would break backwards compatibility for (I suspect) many real data analysis, maybe we can add a deprecation warning first and then removing in a major version bump?

include/tudat/astro/system_models/vehicleSystems.h
DominicDirkx: Indeed :)

src/tudatpy/estimation/observations/expose_observations.cpp
DominicDirkx: I forgot to put this back in, thanks for the reminder!

PR #857, opened 2026-06-27
fix/residual wrapping for periodic observables

PascalChristiaanse:
adds:
- wrapping feature for periodic observables by applying 
 ```
r_wrapped = r - P * round(r/P), with P the observables period.
```
  this wraps the residual to [-P/2, P/2]. Its applied to:
- `angular_position` — Component 0 (RA), period $2\pi$
- `relative_angular_position` — Component 0, period $2\pi$
- `azimuth_elevation_angle` — Component 0 (azimuth), period $2\pi$
- `euler_angle_313_observable` — All 3 components, period $2\pi$

### DIFF
--- a/src/tudat/astro/observation_models/observableTypes.cpp
+++ b/src/tudat/astro/observation_models/observableTypes.cpp
@@ -1745,6 +1745,53 @@ std::map< LinkEndType, int > getSingleLinkStateEntryIndices( const ObservableTyp
     return singleLinkStateEntries;
 }
 
+//! Function to check if an observable type requires residual wrapping.
+bool isResidualWrappingRequired( const ObservableType observableType )
+{
+    bool wrappingRequired = false;
+    switch( observableType )
+    {
+        case angular_position:
+        case relative_angular_position:
+        case azimuth_elevation_angle:
+        case euler_angle_313_observable:
+            wrappingRequired = true;
+            break;
+        default:
+            break;
+    }
+    return wrappingRequired;
+}
+
+//! Function to get the wrapping ranges per component for an observable type.
+std::vector< ResidualWrappingRange > getResidualWrappingRanges( const ObservableType observableType )
+{
+    std::vector< ResidualWrappingRange > wrappingRanges;
+    switch( observableType )
+    {
+        case angular_position:
+        case relative_angular_position:
+        case azimuth_elevation_angle: {
+            wrappingRanges.resize( 2 );
+            // Component 0 (RA / azimuth): residual wraps to [-pi, pi]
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by DominicDirkx** [comment]
Nice! This is a much needed addition. 

Could you make changes such that:
* There is a definition of the list of observableTypes for which the wrapping is needed, so that the function implemented is only called for these observableTypes
* Add a specification per observableType that needs to be wrapped what the range is to which it is to be wrapper (e.g. [0,2*pi]; [-p1,[i])
* Add a unit test to the new functionality

**Comment by PascalChristiaanse** [comment]
>There is a definition of the list of observableTypes for which the wrapping is needed, so that the function implemented is only called for these observableTypes

>Add a specification per observableType that needs to be wrapped what the range is to which it is to be wrapper (e.g. [0,2*pi]; [-p1,[i])

im guessing a nice place would be observableTypes.h?

**Comment by DominicDirkx** [comment]
That seems like a good place for it yes

File: src/tudat/astro/observation_models/observableTypes.cpp [hunk 1]
```diff
@@ -1745,6 +1745,53 @@ std::map< LinkEndType, int > getSingleLinkStateEntryIndices( const ObservableTyp
...
+    switch( observableType )
+    {
+        case angular_position:
+        case relative_angular_position:
+        case azimuth_elevation_angle: {
+            wrappingRanges.resize( 2 );
+            // Component 0 (RA / azimuth): residual wraps to [-pi, pi]
+            wrappingRanges[ 0 ] = ResidualWrappingRange( -mathematical_constants::PI, mathematical_constants::PI );
+            // Component 1 (DEC / elevation): residual wraps to [-pi/2, pi/2]
+            wrappingRanges[ 1 ] = ResidualWrappingRange( -0.5 * mathematical_constants::PI, 0.5 * mathematical_constants::PI );
```
PascalChristiaanse, line 1779: @DominicDirkx This is the line that deals with Dec wrapping, lets discuss this during summer camp.

PascalChristiaanse (commented):
Ready for review

DominicDirkx:
The new wrapObservationResiduals function should now replace checkObservationResidualDiscontinuities right? In two out of three instances where it is used they are now called both. And then I think the checkObservationResidualDiscontinuities can be removed. I also think that, for the angular position, only the right ascension should wrap, not the declination. For the declination, the residuals can be between -180 and 180 degrees (observed at north pole and computed at south pole, and vice versa. Unlike for the right ascension, there is no discontinuity in the declination itself. A small change in position will always lead to a small change in declination.

PR #599, opened 2026-01-07
Unified data library reader

PascalChristiaanse:
feature/unified data library

This PR adds support for data in the UnifiedDataLIbray (UDL) format. This format is a JSON based, vendor specialized dataformat used in the pride group.

The PR introduces a common interface for vendors to implement, allowing easy data ingress of their data schemes.

This initial version adds vendor support for the University of Tasmania.

Feel free to provide feedback! Itll be a few weeks before i can implement suggestions (exams etc...)

PR #858, opened 2026-06-27
READY TO MERGE ON feature/data-refactor - Feature/udl reader

PascalChristiaanse:
continues #599 but with a clean slate


PascalChristiaanse:
alright ive checked prefit residuals with this new module against some data i still had from my internship in australia; and they look very familiar. Ive added testing and reviewed everything, so i think its ready for review @DominicDirkx

PR #861, opened 2026-07-01
Add frequency interpolators docs

larshinueber:
Add API docs for frequency interpolators, move exposures around such that C++ objects are resolved before being referenced in later methods


DominicDirkx:
I like the name of this branch :)

PR #864, opened 2026-07-03
Fix/mission data downloader

luigigisolfi:
This PR aims at improving mission data downloader. Steps implemented:

1) Separation of concerns. moving from monolithic version of the script into a proper library
2) Addressed Backwards Compatibility
3) Removed IONEX/VMF downloaders as a better version was recently pushed in `data/ancillary` (thanks @kimonito98!)

Next steps:
1) fix unused variables 
2) consider simplifying grail logic (merge grail-a and grail-b logic if possible)
3) add proper logging of errors (for now, only bare prints are used)
4) use the robust `_download_file` method where possible (avoid bare (and duplicated) "`urlretrieve`")
5) write unit tests.


DominicDirkx:
@luigigisolfi I've merged these changes into the data module refactor branch. Could you open an issue to track modifications that need to be made to the data downloader?

PR #865, opened 2026-07-04
fix/leap seconds fails on dates before 1972 for GCRS to ITRF conversion

PascalChristiaanse:
Adds:
- Change sofaTimeConversions.cpp to only throw if a _*negative*_ leap second is thrown; indicating an actual error. Leap seconds value can be 0 as the first leap second was in 1972, so dates before that will have 
```
deltaAt =0
```


DominicDirkx:
@PascalChristiaanse Could you add a link to a script that throws an error?

PR #870, opened 2026-07-06
Add interface for Gaia astrometry and orbital data

yverhulst:
This work-in-progress PR adds functionality to obtain Gaia astrometry and Gaia-derived orbit data for ~150,000 solar system objects. The data can be retrieved either online through astroquery or by loading a local version of the archives (downloadable as CSV files from the Gaia archive website).


yverhulst:
@DominicDirkx the gaia interface and correction PRs are almost done. The documentation and unit tests work. So you can review them if you want :)

Only thing that probably still needs some work is the documentation in some places, and I still need to add an example workflow for on the website

src/tudatpy/data/gaia/gaia.py
```diff
@@ -0,0 +1,841 @@
...
+
+
+    @property
+    def table(self) -> pd.DataFrame:
+        """Read-only copy of the astrometry table."""
+        return self._table.copy()
+
+
+    def table_for_mpc(self, mpc_number: int)-> pd.DataFrame:
+        """Return a copy of the observations and metadata filtered for one asteroid by MPC number"""
```
DominicDirkx, line 67: Input parameters should be documented for all user-facing functions, and the return argument should be very clearly specified (type and contents)

src/tudatpy/data/gaia/gaia.py
```diff
@@ -0,0 +1,841 @@
...
+        self._table = observations_and_metadata
+
+
+    @property
+    def table(self) -> pd.DataFrame:
+        """Read-only copy of the astrometry table."""
+        return self._table.copy()
+
+
+    def table_for_mpc(self, mpc_number: int)-> pd.DataFrame:
```
DominicDirkx, line 66: 'for_mpc' sounds like it is for the MPC, not for a single object by mpc number. How about 'table_for_single_object' (that it's by mpc number is implicit in the input argument)

src/tudatpy/data/gaia/gaia.py
```diff
@@ -0,0 +1,841 @@
...
+
+        if any(m not in self.mpc_numbers for m in mpc_numbers):
+            raise ValueError('Requested MPC numbers are not in loaded observation table')
+
+        new = self.copy()
+        new._table = new._table.query('number_mp in @mpc_numbers').reset_index(drop=True)
+
+        return new
+
+    def get_observation_covariance_matrix(self,
```
DominicDirkx, line 118: For now, this is fine, but having a single SingleObservationSet for all transits will initialize a rather large matrix. Would it not be better to instantiate a single SingleObservationSet per transit per object? (not blocking for the merge, but an improvement for after the merge

src/tudatpy/data/gaia/gaia.py
```diff
@@ -0,0 +1,841 @@
...
+            observation_set_list
+        )
+
+        return observation_collection
+
+    def correct_observations(self,
+                             bodies: SystemOfBodies,
+                             light_deflection_bodies: Iterable | None = ('Sun',),
+                             photocenter_correction_model: str | None = None,
+                             diameters: dict | None = None,
```
DominicDirkx, line 234: diameters/sizes should  not come from an input here, but from the SystemOfBodies to ensure consistency between correction and environment (not a blocking point for a merge, but an enhancement to be made)

src/tudatpy/data/gaia/gaia.py
```diff
@@ -0,0 +1,841 @@
...
+                    perturbing_bodies_list=list(light_deflection_bodies),
+                )
+                self._table.loc[mask, light_deflection_columns] = light_bending_corrections
+                self._table.loc[mask, ['ra', 'dec']] += light_bending_corrections
+
+        # Wrap RA
+        self._table['ra'] = (self._table['ra'] + np.pi) % (2 * np.pi) - np.pi
+
+    @classmethod
+    def load_from_astroquery(cls,
```
DominicDirkx, line 324: what is cls?

src/tudatpy/data/gaia/gaia.py
```diff
@@ -0,0 +1,841 @@
...
+        )
+        return settings
+
+
+class GaiaAsteroids:
+    """
+    Class that retrieves Gaia-derived asteroid orbits and covariance data, found in the sso_source tables on the Gaia archives
+    """
+    def __init__(self,
+                 asteroid_orbits_and_covariance: pd.DataFrame) -> None:
```
DominicDirkx, line 585: where do you get the orbits/covariances from? are these not retrieved automatically?

yverhulst: they are retrieved with the classmethods, similar as for the GaiaAstrometry class. So you create the object with "GaiaAsteroids.load_from_astroquery()" or "GaiaAsteroids.load_from_local_archive()" and that gives you an object with the data loaded

DominicDirkx: Thanks, my C++-brain sometimes gets confused about Python :)

DominicDirkx:
I've updated the feature/off_diagonal_weights to be up-to-date with develop, and have run the CI:
https://github.com/tudat-team/tudatpy/actions/runs/30621672565/job/91166197202
It seems to work well, there was an unrelated Windows test failure (nrlmsise) so I retriggered it to make sure. The CI is for the observation_corrections branch, I also just triggered the one for this PR:
https://github.com/tudat-team/tudatpy/actions/runs/30637864429
With the recent update it will also run all the Python tests. Note that there may be failures on due to a connection timeout, these are not related to your implementation (most likely)

src/tudatpy/data/gaia/gaia.py
yverhulst: a reference to the class itself, for classmethod functions

src/tudatpy/data/gaia/gaia.py
yverhulst: but there is no option to create an ellipsoid shape at the moment. so it would be an inconsistency if one type of shape should be in bodies, and the other should be provided separately

DominicDirkx:
@yverhulst I see a lot of commits to the code in this PR :) Can you confirm that it's ready for another review iteration and (as far as you're concerned) merging?

yverhulst:
> @yverhulst I see a lot of commits to the code in this PR :) Can you confirm that it's ready for another review iteration and (as far as you're concerned) merging?

I have not yet addressed all the comments/feedback yet (for this branch or the corrections branch), so not yet. I hope to have that fixed by next week!

yverhulst:
@DominicDirkx I made a number of changes since last time. I think it is now in a state where it's ready to merge. Can you please re-review it? 

Only thing I did not fix is the weight matrix, which is currently still assigned on all observations instead of per transit. I left it like this because I figured this part would be rewritten anyway when adopting the new ObservationDatasets. let me know if I should still change this!

DominicDirkx:
Really cool! Some minor corrections made, and now merging :)

Issue #875, opened 2026-07-07
Implement resource installation as part of main tudatpy


suvayu:
- For conda, a post installation step invokes that code to download the resources
- For non-conda, users invoke the same recipe manually
- The implementation should avoid redundant downloads
- Download location:
  - default: `$HOME/.tudat/resource` (same as now)
  - `TUDAT_RESOURCE_DIR` environment variable, takes precedence if this is set

HannoSpreeuw:
Suggestion from @jhidding : use the [Pooch Python package](https://pypi.org/project/pooch/).

HannoSpreeuw:
> Suggestion from [@jhidding](https://github.com/jhidding) : use the [Pooch Python package](https://pypi.org/project/pooch/).

> Pooch is a Python library that can manage data by downloading files from a server (only when needed) and storing them locally in a data cache (a folder on your computer).

This should then also cover #881

HannoSpreeuw:
Suggestion from @DominicDirkx  : Add a progress  bar for downloading.

suvayu:
I was looking at the `875-resource-installer` branch.  There's been a bit of misunderstanding on the organisation.  I see it's been implemented as standalone scripts at the top-level.  This does not work well with packaging.  I guess the confusion happened because everyone keeps calling it a "script".

Everything should be moved to an executable module under tudat proper:
`resource_{installer,data_registry}.py` -> `src/tudatpy/resource_{installer,data_registry}.py`.
And it should be run as `python -m tudat.resource_installer [options]`.

I would also prefer moving the contents of `resource_data_registry.py` to a data file like CSV, or simply, we could just have the `manifest.txt` here as the canonical source (without the DOI line of course).  It is after all a catalog, not code.  In the long term, this will be much easier to maintain; there are 150+ entries!

Also, catalog is my preferred name over registry.

DominicDirkx:
For my clarification, the only change here:

> 
> Everything should be moved to an executable module under tudat proper:
> resource_{installer,data_registry}.py -> src/tudatpy/resource_{installer,data_registry}.py.
> And it should be run as python -m tudat.resource_installer [options].
> 

is where the file would be right?

> I would also prefer moving the contents of resource_data_registry.py to a data file like CSV, or simply, we could just have the manifest.txt here as the canonical source (without the DOI line of course). It is after all a catalog, not code. In the long term, this will be much easier to maintain; there are 150+ entries!

[...]

Issue #874, opened 2026-07-07
Support customisable location


suvayu:
Current implementation looks for data files in `$HOME/.tudat/`.  Support customisable location, probably through an environment variable
  - Storing large datasets in the home directory on an HPC is often inconvenient

Issue #867, opened 2026-07-06
[F&E]: Remove the need for `tudat_resources`


suvayu:
### Description:

Maintaining `tudat_resources` and downloading it repeatedly is inconvenient, let's get rid of it.

### User's perspective 
- Downloading from original sources is preferred over maintaining our own download.
  - #873 
  - #877
- #875
- Add functionality where the user can provide their own data files for a specific subset, e.g. ephemerides.
- Decide how to host it.
  - Zenodo?
  - Make the same choice as the snapshot for testing?

### Developer's perspective
Unit tests need almost all of the resources, however downloading from original sources is inconsistent as different providers have different reliability.  So our proposal is we maintain a **snapshot** of the resources, updated as needed, e.g. when a new feature test requires a new data file, we can update the test tarball.

- #866
- #874 
- In CI, download the resources in a separate step of the CI job before running `ctest`.
- Possible solution for CI: cache the download, any past successful download will be cached.  Check cache size and number of files quota.

PR #876, opened 2026-07-07
Remove tudat_resources as a separate package

suvayu:
Addresses:
- #867
- #874 

Merge this PR only after #875 (code to download datasets) is implemented

### Summary
The test suite still requires the resource files.  Downloading them on CI is not very reliable.  Specially I don't think Zenodo finds it acceptable to download records in CI.  Additionally, in the last 2 weeks [Zenodo](https://blog.zenodo.org/2026/09/15/2026-09-15-stability-and-performance-updates/) has been struggling with too much automated traffic, including AI crawlers.  We don't want to be bad citizens and add to the problem.  So, this PR also adds a few mechanisms to deal with that.
- Caching of the resource files in between CI runs.
- Identify the subset of files required by the test suite.  This is included in the `tests/` directory as a resource catalog.  The cache is created/invalidated based on this list.
- For the larger files, identify the data blocks required by the tests, and trim them.  Final total size achieved: <600MB.
- The trimmed resources are hosted in the `tudat-resources` repository, see: https://github.com/tudat-escience/tudat-resources/tree/updated_resource_directory/test-data.  This bypasses the zenodo issue.

#### Remarks
- Hosting the test related resources on another repo is not ideal for various reasons.  But this is the simplest solution I could find given limited time.  Ideally, this should be hosted using `git-lfs` within the main repo.
- The trimming exercise was conservative since I'm not a domain expert.  You could probably be more aggressive to reduce this further.  Note that this also requires more developer discipline.


HannoSpreeuw (commented):
Where is the support for a customisable location?

HannoSpreeuw:
I guess you would also want to remove
>   - tudat-resources

from the [environment.yaml file](https://github.com/tudat-team/tudatpy/blob/e0a384b25ddb700346e36319f86156edcee42a55/environment.yaml#L10).

suvayu:
https://github.com/tudat-team/tudatpy/pull/876/changes#diff-9bcd77d39e9d808da5c75b1acfe7df40701b3f655f401f8cbc11a7e8fac51c13R50-R62

This is the configurable location code.

And I fixed the environment file in the latest commit.  Thanks for noticing :)

include/tudat/resource/resource.h
HannoSpreeuw, line 1: Why do we need this new file?

suvayu: These functions are what the C++ code uses to find the datasets.  It's virtually identical with what the tudat_resources package provided.  It adds a feature, looks for the resources at `TUDATPY_RESOURCE_DIR`, before defaulting to regular location.

suvayu:
@DominicDirkx @yiquintero 

Things to review are:
- my data trimming (see the `tudat-resources` fork mentioned in the description).
- CI caching
- Windows build failure that is unrelated to any of the changes I made.  Maybe this goes away after a rebase.

I'll wait for your reviews before rebasing, don't want to rebase more times than necessary.

PR #887, opened 2026-07-08
Separate ground station logic, Doppler observation collection compression, DSS-65 updates

larshinueber:
Minor changes:

- make ground station locations static variables to avoid repeated definitions
- Add explicit argument to Doppler observation collection compression functions for approximate ground station locations
- Move Doppler observation collection compression functionality to dedicated header file
- Use piece-wise constant ground station motion for DSS-65 as it has been moved in 2005, see https://naif.jpl.nasa.gov/pub/naif/generic_kernels/spk/stations/a_old_versions/earthstns_itrf93_050714.cmt
- Only build explicit double double instantiations from build.py script if tests are build
- Fix compile commands export


larshinueber:
@DominicDirkx this is ready for review and merge now! It's a bunch of small changes that would be good to have before adding more changes in the data refactor.
In particular, functions unrelated to odf processing are moved out of the processOdfFile.h header (such that this can be safely deprecated once we have implemented the new data flow). The default ground station state functions are generalized for the station complexes and the positions are marked static const to avoid repeated definitions.

DominicDirkx:
Cool, thanks!

PR #896, opened 2026-07-11
Feature/data refactor

luigigisolfi:
The updated MPC estimation example was previously failing due to what I believe to be a bug in the **ancillarysettings** creation. More specifically, it looks like the `getAncillarySettingsFromTrackingData` was handing back a non-null pointer even though I had no ancillary settings in my case. This would then trigger the following line:

```
if( ancillarySetings != nullptr )
{
    throw std::runtime_error( "Error, calling angular position observable with ancillary settings, but none are supported." );
}
```

Adding these lines to `createObservationCollection.h`'s `getAncillarySettingsFromTrackingData`before returnning the ancillarySettings:

```
    if( trackingData->getAncillarySettingsDouble( ).empty( ) &&
    trackingData->getAncillarySettingsDoubleVector( ).empty( ) &&
    trackingData->getAncillarySettingsStringVector( ).empty( ) &&
    trackingData->getAncillarySettingsString( ).empty( ) )
    {
        return nullptr;
    }

    return ancillarySettings;
}
```


DominicDirkx:
Draft PR to check mergeability of mpc module

PR #906, opened 2026-07-21
DSN N-Way range partials

larshinueber:
First draft of DSN n-way range partials.
The partials for the _DSN_ n-way range model are implemented as scaling factors to the partials of the n-way range model, which already implements the partials of the precise light time. The scaling factors then account for the conversion from meters in the n-way range partials to Range Units in the DSN n-way range model, according to Moyer (2003), eq. 13-128.
The unit test follows the DSN n-way doppler unit test but needs more checking for the tolerance settings.

### DIFF
--- a/cmake/tudat/TudatLinkLibraries.cmake
+++ b/cmake/tudat/TudatLinkLibraries.cmake
@@ -51,6 +51,7 @@ endif ()
 
 list(APPEND Tudat_PROPAGATION_LIBRARIES
         Tudat::tudat_propagation_setup
+        Tudat::tudat_estimatable_parameters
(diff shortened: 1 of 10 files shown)


DominicDirkx:
Good to see this added :) @larshinueber It looks like the build is failing on some linking error.

larshinueber:
Thanks for the ping @DominicDirkx! I moved some stuff to the .cpp file but then forgot to add it in the CMakeList 😅

DominicDirkx:
I did a sweep to modify the perturbations and allow a more stringent tolerance. Locally I was able to reduce the tolerance by a factor 200. If this also works out in the CI, I suggest we merge this.

PR #911, opened 2026-07-28
Clean stubs build

### LINKED ISSUE #288: [BUG]: Missing types during stub generation

DominicDirkx:
This PR fixes the missing and incorrectly formatted type information in the generated Python stubs.

## Changes

- Reorders the bindings so classes and enums are exposed before they are used in function signatures.
- Splits type declarations from the remaining bindings where dependencies cross modules.
- Replaces invalid C++ default-value representations with valid Python equivalents
- Uses public factory functions for default settings where they produce the same result  (and modifies parameter description accordingly).
- Adds the enum mappings needed by `pybind11-stubgen`.
- Skips non-Python directories when creating the stub directory structure.
- Corrects affected default values and parameter documentation.
- Fixes the documentation warnings found while validating the changes.
- Ensures local documentation builds use the package from the current build directory.

## Validation

- `python build.py -j8` completes without compiler or stub-generation warnings.
- Stub generation completes successfully.
- The Sphinx documentation builds with zero warnings or errors.

Closes #288


DominicDirkx:
@larshinueber @alfonsoSR May be of interest to you :)

build.py
```diff
@@ -594,8 +637,12 @@ def __create_stubs_directory_structure(self) -> None:
 
         for item in self.python_source_dir.rglob("*"):
 
-            # Skip if not a directory or if it is a cache directory
-            if not item.is_dir() or item.name == "__pycache__":
+            # Only Python packages require a stub directory.
```
alfonsoSR, line 640: Do we have non-python packages?

DominicDirkx: No, not in the tracked source tree. This prevents (generated or temporary) directories without an __init__.py from being treated as Python packages.

Issue #877, opened 2026-07-07
Document preprocessing of data files in resources that are created by Tudat


suvayu:
Some files in the tudat resources are created by Tudat after some preprocessing.  These preprocessing steps need to be documented (preferably in code) before we can implement semi-automatic updates.

### TODOs
from: https://github.com/tudat-team/tudatpy/issues/873#issuecomment-4904940408

- [ ] atmosphere_tables
- [ ] earth_deformation
- [ ] earth_orientation
- [ ] ephemeris
- [ ] gravity_models (partially)
  - [ ] .gfc -> Earth/GOCO05c.txt
  - [ ] Earth/egm96.txt (source moved)
  - [ ] Earth/ggm02c.txt and Earth/ggm02s.txt
  - [ ] Mars/jgmro120d.txt (no .txt in download)
  - [ ] Moon/glgm3150.txt and Moon/lpe200.txt (no .txt in download)
- [ ] quadrature
- [ ] star_catalog_biases
- [x] station_locations

HannoSpreeuw:
After discussion with @suvayu : these preprocessing steps can, most likely, not be found in a GH repo somewhere. 
Preprocessing is probably perfomed through a script or a couple of scripts, but we do not know where those scripts are stored.

luigigisolfi:
@HannoSpreeuw for the optical stations locations, see #878, where I also attached the code I used

Issue #873, opened 2026-07-07
Identify data created by Tudat


suvayu:

yiquintero:
The size (on disk) of the [resource directory](https://github.com/tudat-team/tudat-resources/tree/master/resource) of the `tudat-resources` repo is **763MB**.
The size (on disk) of the [tudat-resources v2.4 release](https://github.com/tudat-team/tudat-resources/releases/tag/v2.4) is **2.1GB**
There is a difference of **1.3694GB**.

This table lists the difference in sizes between the release and the repository folder.

| folder name | size in repo | size in release v2.4 | difference |
|----------|-----------|----------|-----------|
| atmosphere_tables | 376M | 387M | 11M |
| earth_deformation | 104K | 104K | - |
| earth_orientation | 6.5M | 6.5M | - |
| ephemeris | 8K | 24K | 16K |
| gravity_models | 117M | 117M | - |
| quadrature | 76K | 76K| - |
| space_weather | 3.2M | 3.1M | -0.1M |
| spice_kernels | 222M | 1.5G | 1278M |
| star_catalog_biases | 37M | 37M | - |
| station_locations | 572K | 572K | - |

The **1.3694GB** difference in sizes is due to the following files and directories which are contained in the release, but not stored in the repo:

| directory | type | name | size | present in |
|------|------|------|------|------------|
|`atmosphere_tables` | file | `nrlmsise00_validation_data.pkl` | 11M | release | 
|`atmosphere_tables` | dir (8 files)  | `vmf3` | 96K | release |
|`ephemeris`         | file | `eros_obs.txt` | 16K | release |
|`space_weather`     | file | `sw19571001.txt` | - | both, 3.2M in repo, 3.1M in release |
|`spice_kernels`     | file | `NOE-4-2020.bsp` | 233M | release | 
|`spice_kernels`     | file | `NOE-5-2021.bsp` | 434M | release |
|`spice_kernels`     | file | `NOE-6-2018-MAIN-v2.bsp` | 212M | release |
|`spice_kernels`     | file | `de440.bsp` | 114M | release |
|`spice_kernels`     | dir (28 files) | `inpop19a_TCB_m100_p100_asc` | 180M | release |
|`spice_kernels`     | dir (28 files) | `inpop19a_TDB_m100_p100_asc` | 180M | release |
|`spice_kernels`     | file | `inpop19a_TDB_m100_p100_spice.bpc` | 2.3M | release |

kgzenk:
[...]

PR #915, opened 2026-07-29
Remove the need for a separate "tudat-resources" repo.

### LINKED ISSUE #867: [F&E]: Remove the need for `tudat_resources`

### LINKED ISSUE #874: Support customisable location

### LINKED ISSUE #875: Implement resource installation as part of main tudatpy

### LINKED ISSUE #881: [F&E]: Caching updateable resources
### Description:

Related to #867, following a question from @luigigisolfi : Can we cache updateable resources?
That would be convenient when an update is interrupted.

So perhaps not a subissue of #867 , but an extension, or enhancement?

_No response_

### Additional context:

_No response_

HannoSpreeuw:
Closes #875

this includes addressing [one](https://github.com/tudat-team/tudatpy/issues/875#issuecomment-4927912026) of @suvayu 's comments.

Closes #881

Closes #874

Closes #867 

I will need to check if this also addresses #873 completely, e.g. the [very descriptive error message](https://github.com/tudat-team/tudatpy/issues/873#issuecomment-4913115722) might not yet be as descriptive as we would like.

#877 remains open.


HannoSpreeuw:
After this we can proceed with PR #876

DominicDirkx:
@HannoSpreeuw Great! Could you give a summary of the functionality this implements, and how it relates to the rest of the work to get rid of the tudat-resources dependency? I also see there's no test of any functionality here. Will the testing of this be part of the next PR (or, will getting rid of resources and testing if everything still works be the effective test)?

HannoSpreeuw:
> @HannoSpreeuw Great! Could you give a summary of the functionality this implements

Right. This PR introduces four new files:  `resource_installer.py`, `resource_hashes.json`, `resource_catalog.py` and `resource_catalog.csv`

- resource_installer.py implements the command-line tool for downloading, caching, extracting, updating, and optionally SHA-256-verifying Tudat resource files.
- resource_hashes.json contains the expected SHA-256 checksums for the downloadable resources and archives, allowing the installer to verify file integrity.
- resource_catalog.py provides the logic for loading the packaged resource catalogue and refreshing it from the authoritative Zenodo manifests.
- resource_catalog.csv is the packaged default catalogue that maps each resource path to its source archive or download URL, together with provenance and modification metadata.

`resource_catalog.csv` will be updated from our `manifest.txt` files as part of our two Zenodo records by issuing `python -m tudatpy.resource_installer --mode manifest`.
The `resource_catalog.csv` file is convenient when requesting a list of all  `tudat-resources` files by issuing
`python -m tudatpy.resource_installer --mode list` which will not require an internet connection.

Please note that any change in our Zenodo files will need to go along with a corresponding change in the `manifest.txt` files and in `resource_hashes.json`.
The `TUDATPY_RESOURCE_DIR` and `TUDATPY_RESOURCE_CACHE` env vars set the paths for downloading and caching the `tudat-resources` data that is now stored in two Zenodo records, with default paths equal to `~/.tudat/resource` and `~/.cache/tudatpy_resources`, respectively.

To make sure these four files are packaged, i.e. copied to a build directory, we added [a few lines to the PythonExtension.cmake file](https://github.com/tudat-team/tudatpy/blob/ae495cfbbf99d47797669f59770ad2f1d551b75a/cmake/tudatpy/PythonExtension.cmake#L109-L120)


> and how it relates to the rest of the work to get rid of the tudat-resources dependency? 

This PR together with PR #876 should do exactly that; e.g. [PR #876 includes removing `tudat-resources` from the environment.yaml file](https://github.com/tudat-team/tudatpy/pull/876/changes/736f40cfb6824f94a68ee0a75fedd24fb8dabb63).

Now what is pending is the automatic execution of `python -m tudatpy.resource_installer --mode missing` as part of the Tudatpy install process, i.e. as part of `conda env create -f environment.yaml`. Most likely this has to be added to a Conda recipe and something similar for the conda-free build. I will discuss with @suvayu whether this automatic execution should become part of this PR or of #876.


> I also see there's no test of any functionality here.  Will the testing of this be part of the next PR (or, will getting rid of resources and testing if everything still works be the effective test)?

I am inclined to argue that this does not need new unit tests, since these two PRs constitute a replacement - of `tudat-resources` - rather than adding a feature.

DominicDirkx:
Thanks, that's very clear. So the test of this will be that the whole process of getting the resources does indeed still work when it is used instead of the resources package. Then indeed no new test needed.

In terms of incorporating this into the installation, with the conda package a change in the post-link script should do it (calling this file instead of downloading the resources manually as is done now)

src/tudatpy/resource_catalog.py
```diff
@@ -0,0 +1,116 @@
...
+
+    for line_number, row in enumerate(rows, start=2):
+        if len(row) != 3:
+            raise ValueError(f"Invalid manifest row {line_number} from {source}: {row}")
+        path, modified, source_url = (value.strip() for value in row)
+        if not source_url:
+            # The static manifest currently has a malformed star-catalog
+            # archive descriptor. The record exposes the canonical name.
+            archive = "star_catalog_biases.tar.gz" if path == ".tar.gz_catalog_biases.tar" else path
+            continue
```
suvayu, line 75: Is there an error in the zenodo record?  We shouldn't do this kind of workaround, instead the record should be fixed.

EDIT: Okay, I checked.  Somehow typo snuck in, my bad.

suvayu:
Hi @DominicDirkx I was still reviewing this, I don't think this should be merged yet.  I have comments that need addressing.

src/tudatpy/resource_installer.py
```diff
@@ -0,0 +1,473 @@
...
+            matches = {
+                catalog_key: url
+                for catalog_key, url in RESOURCE_CATALOG.items()
+                if catalog_key.startswith(key) or key in catalog_key
+            }
+            if not matches:
+                raise ValueError(f"No catalog entries match '{key}'")
+            resolved.update(matches)
+
+    return resolved
```
suvayu, line 97: I don't see a difference between `keys` and `search`.  Both match by substring match (i.e. `'<token>' in key`), so the complexity of `resolve_catalog_keys` is unnecessary, including the distinction between the two arguments.  These two are equivalent:
```
resolve_catalog_keys(search="<search_str>")
resolve_catalog_keys(keys=["<search_str>"])
```
I would change the implementation to something like this:
```
def resolve_catalog_keys(keys: list[str]):
    resolved = {}
    for k in keys:
        resolved.update(find_in_catalog(k))
    return resolved
```

src/tudatpy/resource_installer.py
```diff
@@ -0,0 +1,473 @@
...
+        "--cache-dir",
+        default=str(DEFAULT_CACHE),
+        help="Cache directory for downloads and tarballs.",
+    )
+    parser.add_argument(
+        "--files", nargs="+", help="Exact catalog keys or prefixes for update mode."
+    )
+    parser.add_argument(
+        "--search", help="Substring search to select catalog files for update or list."
+    )
```
suvayu, line 388: If you update `resolve_catalog_keys` as suggested, then these should be merged.  Maybe something like this?
```suggestion
    parser.add_argument(
        "--files", nargs="+", help="Restrict update to catalog entries that partially string match the provided list"
    )
```
No change in functionality, and simpler interface.

src/tudatpy/resource_installer.py
```diff
@@ -0,0 +1,473 @@
...
+from tudatpy.resource_catalog import (
+    USER_RESOURCE_CATALOG,
+    load_resource_catalog,
+    update_resource_catalog,
+)
+
+try:
+    import pooch
+except ImportError:  # pragma: no cover
+    pooch = None
```
suvayu, line 24: Why is this optional?  Either use it or don't.  I had no strong opinion either way, but since `_download_with_requests` is already implemented, we might as well drop the dependency.

src/tudatpy/resource_installer.py
```diff
@@ -0,0 +1,473 @@
...
+            raise
+        print(f"Verified SHA256 for {dest_path}")
+    return True
+
+
+def _download_tarball(
+    url: str, cache_dir: Path, force: bool = False, expected_hash: Optional[str] = None
+) -> Path:
+    cache_dir.mkdir(parents=True, exist_ok=True)
+    tar_path = cache_dir / _tarball_cache_name(url)
```
suvayu, line 201: I don't think caching the downloaded tarballs is a good idea since there is no way to determine when to invalidate the cache.  If the intention is to reuse it during one execution, then you could store it as a temporary file.  You can get a platform agnostic temporary directory using [`tempfile.TemporaryDirectory`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory).

HannoSpreeuw:
> Hi @DominicDirkx I was still reviewing this, I don't think this should be merged yet. I have comments that need addressing.

Okay, it should not be too hard to undo the merge.

src/tudatpy/resource_installer.py
```diff
@@ -0,0 +1,473 @@
...
+        for chunk in iter(lambda: f.read(8192), b""):
+            h.update(chunk)
+    return h.hexdigest()
+
+
+def _verify_sha256(path: Path, expected_hash: str) -> None:
+    """Verify a file against an expected SHA256 digest."""
+    got = _sha256_file(path)
+    if got.lower() != expected_hash.lower():
+        raise RuntimeError(f"SHA256 mismatch for {path}: expected {expected_hash}, got {got}")
```
suvayu, line 170: Instead of raising, return `bool`.  Return values are much simpler to act upon, also hash mismatch is not necessarily an error, e.g. when you are updating a file.

src/tudatpy/resource_installer.py
```diff
@@ -0,0 +1,473 @@
...
+) -> int:
+    installed = 0
+    verified = 0
+    print(f"Extracting {len(targets)} resources from {tar_path.name}")
+    with tarfile.open(tar_path, mode="r:*") as tar:
+        for target in targets:
+            dest_path = dest_root / target
+            if dest_path.exists() and not force:
+                if hashes and (expected_hash := hashes.get(target)):
+                    _verify_sha256(dest_path, expected_hash)
```
suvayu, line 254: Explicitly check if verification passed or failed.  If you switch the exception to `bool` return value, much easier to do.  The `continue` statement should be conditioned on if verification passes.

src/tudatpy/resource_installer.py
```diff
@@ -0,0 +1,473 @@
...
+            extracted = tar.extractfile(member)
+            if extracted is None:
+                raise RuntimeError(f"Cannot extract '{member.name}' from {tar_path}")
+            dest_path.write_bytes(extracted.read())
+            if hashes and (expected_hash := hashes.get(target)):
+                try:
+                    _verify_sha256(dest_path, expected_hash)
+                except RuntimeError:
+                    dest_path.unlink(missing_ok=True)
+                    raise
```
suvayu, line 268: If verification fails, you should issue a warning and continue, because there are other `target`s to process.  Better to try all and report, rather than fail one by one.

src/tudatpy/resource_installer.py
```diff
@@ -0,0 +1,473 @@
...
+        _download_with_requests(url, tar_path)
+
+    if expected_hash:
+        try:
+            _verify_sha256(tar_path, expected_hash)
+        except RuntimeError:
+            tar_path.unlink(missing_ok=True)
+            raise
+        print(f"Verified SHA256 for archive {tar_path.name}")
+    return tar_path
```
suvayu, line 220: If you remove the caching behaviour, then there's no difference between `_download_tarball` and `_download_file`.

src/tudatpy/resource_installer.py
```diff
@@ -0,0 +1,473 @@
...
+    dest_path = dest_path.expanduser()
+    if dest_path.is_dir():
+        dest_path = dest_path / Path(url).name
+    dest_path.parent.mkdir(parents=True, exist_ok=True)
+    expected = None
+    if hashes:
+        expected = hashes.get(url) or hashes.get(Path(url).name)
+    if _download_file(url, dest_path, force=force, expected_hash=expected):
+        return dest_path
+    return dest_path
```
suvayu, line 333: IIUC, this is for the user to add a resource of their own?  If so, I would also update the local manifest file, add a line corresponding to the file that was downloaded (path, timestamp, source_url).  If the checksum file is also there, adding an entry there would also be good.

src/tudatpy/resource_installer.py
```diff
@@ -0,0 +1,473 @@
...
+        "--search", help="Substring search to select catalog files for update or list."
+    )
+    parser.add_argument("--extra-url", help="URL of an extra file to download in extra mode.")
+    parser.add_argument(
+        "--extra-dest", help="Relative or absolute destination path for extra mode."
+    )
+    parser.add_argument("--force", action="store_true", help="Overwrite existing files.")
+    parser.add_argument(
+        "--list-search", help="Search string to filter catalog output in list mode."
+    )
```
suvayu, line 396: Similar to `--files` and `--search`, this is also essentially the same.  There is no need for this extra flag.  You already know the mode from `--mode`, so passing the argument from `--files` to `list_catalog` would work.

src/tudatpy/resource_installer.py
```diff
@@ -0,0 +1,473 @@
...
+    return dest_path
+
+
+def list_catalog(search: Optional[str] = None) -> None:
+    """Print catalog entries, optionally filtered by search string."""
+    entries = RESOURCE_CATALOG
+    if search:
+        entries = find_in_catalog(search)
+    for key in sorted(entries):
+        print(key)
```
suvayu, line 342: If you switch to `resolve_catalog_keys` (after my suggestion to merge `--files` and `--search`) then this can list multiple paths.  The argument has to change to `list[str]`.

suvayu (commented):
I'm not sure this structure of embedding the resource catalog and the hashes in the repo is correct.  Basically updating a file will now require updating these in two places.  And the convention is these kind of files are always distributed **with** the data.  So the ones in the repo should go.

The use case is offline operations, like list/verify.  But these are always after first download.  So I think it is reasonable to say, the installer always starts by downloading the manifest and checksum files from zenodo first (one of the reasons they are not in an archive).

There is one issue though, both manifest & checksum files are named identically.  So downloading blindly will overwrite the previous one.  The simplest solution is to rename one of them.  I suggest for the "rolling" space weather dataset, we rename:
- `manifest.txt -> manifest-rolling.txt`, and
- `SHA256SUM -> SHA256SUM-rolling`.

If you agree, I'll go about fixing the the datasets on zenodo.

src/tudatpy/resource_installer.py
HannoSpreeuw: Nice catch.
Addressed through [e02048e](https://github.com/tudat-team/tudatpy/commit/e02048e6bc895b622728582a3b633979126ac1e9).

src/tudatpy/resource_installer.py
HannoSpreeuw: Also addressed through [e02048e](https://github.com/tudat-team/tudatpy/commit/e02048e6bc895b622728582a3b633979126ac1e9).

src/tudatpy/resource_installer.py
HannoSpreeuw: Addressed through commit [ee1de56](https://github.com/tudat-team/tudatpy/commit/ee1de5613b03abbf78c795e08039824137cabe62).

src/tudatpy/resource_installer.py
HannoSpreeuw: We can remove caching, but we added caching at the request of  @luigigisolfi , see #881 .
As far as I can remember the intention was to resume an interrupted download.

src/tudatpy/resource_catalog.py
HannoSpreeuw: For the record: we fixed that together.

src/tudatpy/resource_installer.py
suvayu: Okay, that should be kept in temp.  The assumption is, if you find an existing "cache", that's an interrupted download from a recent attempt.  In that case, using `tempfile.TemporaryDirectory` is probably not correct since it creates a unique path every time.  The path should be predictable, but unique for a given user.  So maybe something like this: 

```python
import tempfile
import getpass

TMP = tempfile.gettempdir()
USER = getpass.getuser()

DOWNLOAD_DIR = f"{TMP}/tudat_resource_installer_{USER}/"
```

Then you can download any files under `DOWNLOAD_DIR` as you were doing `file.ext.part` -> `file.ext`, and move to final location.

That said, I would hesitate to call it cache.  There is no invalidation logic, maybe just call it resumable/partial downloads, or download parts?

HannoSpreeuw: > There is no invalidation logic, maybe just call it resumable/partial downloads, or download parts?

I think we can reasonably assume that a user has first run
`python -m tudatpy.resource_installer --mode manifest`

Wouldn't that mean that the hashes of the tar files have also been updated and are known?

After all, Zenodo record 21261530 contains `SHA256SUM-rolling`, which includes the hash of `resource-rolling.tar.gz`.

Similarly, `SHA256SUM-static` in Zenodo record 21277280 contains the hashes of all the tar files.

So shouldn't it be possible to invalidate the cache based on those hashes, or am I missing something?

src/tudatpy/resource_installer.py
HannoSpreeuw: Addressed through commit [3377e97](https://github.com/tudat-team/tudatpy/commit/3377e97179d438965b51006aea9206285971ab94)

src/tudatpy/resource_installer.py
HannoSpreeuw: > So shouldn't it be possible to invalidate the cache based on those hashes

Addressed through commit [3377e97](https://github.com/tudat-team/tudatpy/commit/3377e97179d438965b51006aea9206285971ab94).

src/tudatpy/resource_installer.py
HannoSpreeuw: Addressed through commit [d66cadc](https://github.com/tudat-team/tudatpy/commit/d66cadca8b0e1bb56791d93073744db09a7ed675).

src/tudatpy/resource_installer.py
HannoSpreeuw: I thought we could invalidate caches through commit [3377e97](https://github.com/tudat-team/tudatpy/commit/3377e97179d438965b51006aea9206285971ab94), as mentioned [here](https://github.com/tudat-team/tudatpy/pull/915/changes#r3740777354), and in that way retain the caching behaviour.

src/tudatpy/resource_installer.py
HannoSpreeuw: Addressed through commit [d66cadc](https://github.com/tudat-team/tudatpy/commit/d66cadca8b0e1bb56791d93073744db09a7ed675).

src/tudatpy/resource_installer.py
HannoSpreeuw: Yes, this mode allows users to download an arbitrary resource that is not part of the Tudat resource catalog. However, I intended it as a one-off download rather than as a way to register a persistent user-managed resource.

src/tudatpy/resource_installer.py
HannoSpreeuw: Addressed through commit [3ace8c5](https://github.com/tudat-team/tudatpy/commit/3ace8c5a4176268889b392aec5791333fd82e7c2).

src/tudatpy/resource_installer.py
HannoSpreeuw: Addressed through commit [d08f781](https://github.com/tudat-team/tudatpy/commit/d08f78159bd187d3266d57854eb1ab8821115e10).

src/tudatpy/resource_installer.py
HannoSpreeuw: Commit [b0bf6aa](https://github.com/tudat-team/tudatpy/commit/b0bf6aa5aac089dddd324a460821308a73916045) adds a feature to the
```
python -m tudatpy.resource_installer --mode list --catalog 
    path/to/metadata/resource_catalog.csv --dest path/to/resources
````
command, i.e. it highlights files - by annotating with `[not in manifest]` - in the `path/to/resources` directory that are not in `path/to/metadata/resource_catalog.csv`.

Btw, `path/to/metadata/resource_catalog.csv` is created from
```
python -m tudatpy.resource_installer \
    --mode manifest \
    --catalog path/to/metadata/resource_catalog.csv
```

PR #930, opened 2026-08-16
Add round-trip light-time observation dependent variable

### LINKED ISSUE #927: [F&E]: Round-trip light time as observation dependent variable
### Description:

It would be useful to be able to retrieve the round-trip light time of a signal in an observation. Ideally, it should be configurable for which part of a link geometry this light time is computed for (e.g., in a two-way observation it should be possible to retrieve the light time of the up-, down-, or up- and downlink portion).

Without having looked at the implementation of observation dependent variables in detail, I would imagine the implementation could be similar to the implementation of the `target_range` dependent variable: https://github.com/tudat-team/tudatpy/blob/8879d010ff19fa86cad8ad83ee77b8e4bbe66443/src/tudat/simulation/estimation_setup/observationOutput.cpp#L318

### Additional context:

_No response_

Strtus:
## Summary

- Add `light_time_between_link_ends_dependent_variable` for supported linear observation chains.
- Read each selected leg's solved light time directly, preserving configured light-time corrections while excluding link-end delays.
- Support uplink, downlink, and complete-path selections, with clear errors for ambiguous or reversed selections.
- Add C++ and Python regression coverage for late epochs, n-way paths, retransmission delays, and integrated observables.

## Validation

- `ctest --test-dir ../build-codex-review -R 'ObservationDependentVariables|LightTimeCorrectionComponents|LightTimeDependentVariable' --output-on-failure`
- `ctest --test-dir ../build-codex-review -R '^test_observation_models_NWayRangeObservationModel$' --output-on-failure`
- `PYTHONPATH=../build-codex-review/src python -m pytest -q tests/test_tudatpy/test_observation_light_time_dependent_variable.py`
- `git diff --check`

Closes #927


.github/workflows/inform.yml
```diff
@@ -10,11 +10,12 @@ on:
 
 jobs:
   dispacher:
+    if: github.repository == 'tudat-team/tudatpy'
```
DominicDirkx: why is this file modified?

include/tudat/simulation/estimation_setup/observationOutputSettings.h
```diff
@@ -39,7 +40,8 @@ enum ObservationDependentVariables {
     integration_time_dependent_variable,
     retransmission_delays_dependent_variable,
     link_end_epochs_dependent_variable,
-    light_time_correction_components
+    light_time_correction_components,
```
DominicDirkx, line 43: could you add 'dependent_variable' to the name (to prevent name clashes with other enums)?

tests/test_tudatpy/test_observation_light_time_dependent_variable.py
DominicDirkx, line 1: Is a separate Python test for this specific functionality needed?

DominicDirkx:
@larshinueber Could you also have a look at this?

DominicDirkx:
@Strtus Many thanks for another nice PR! I'm just back from leave and am quickly going through everything that happened while I was away. I left a few small notes in the PR, a more in-depth review will follow soon

src/tudat/simulation/estimation_setup/observationOutput.cpp
```diff
@@ -924,6 +1088,20 @@ void ObservationDependentVariableCalculator::addDependentVariableFunction(
         const int currentIndex,
         const int parameterSize )
 {
+    // `light_time` is read from the final per-leg light-time solutions rather than reconstructed
```
DominicDirkx: I think we should move this (and the if( variableSettings->variableType_ == light_time_correction_components ) in a way to not be treated as special cases. I think the better approach is to add the legLightTimeCalculators_ as input to the getObservationVectorDependentVariableFunction function

Strtus:
Thanks for taking a look @DominicDirkx  I’ve seen the comments — there seems to be quite a bit to work through, and I agree with the direction you’re suggesting. I’m still at the lab at the moment, so I’ll go through everything properly later and push an update

.github/workflows/inform.yml
Strtus: this is unrelated to the light-time changes , it was an extra workflow cleanup that ended up in this PR
The intent was to avoid running the repository-dispatch job from forks and to update the action version, but it shouldn’t be part of #930. I’ll revert this change when I get back to the PR later

Strtus: Reverted now — this file is no longer part of the PR

include/tudat/simulation/estimation_setup/observationOutputSettings.h
Strtus: Yep, renamed it to `light_time_dependent_variable`

tests/test_tudatpy/test_observation_light_time_dependent_variable.py
Strtus: I dropped it

src/tudat/simulation/estimation_setup/observationOutput.cpp
Strtus: Agreed — I’ve moved both through `getObservationVectorDependentVariableFunction` and pass the leg calculator map into the factory now. I left only the size-resolution step for the correction components outside, since that still depends on the available calculators

PR #934, opened 2026-08-18
Complement and Expose TLE Fitting Functionality

luigigisolfi:
- Add TLE raw lines writer upon fitting
- Add optional Tle Fit settings to set the norad id and intldes of an object
- While I was working on it, I also added getter methods and exposures to get the international designator from the `environment.Tle` object


DominicDirkx:
Thanks! This looks good, but build on #936 which should be merged before this one

PR #936, opened 2026-08-18
Improvements to crashing `to_python_datetime` edge cases

luigigisolfi:
@valeriof7 could you check this out?

Changes motivated by the following error:

Input
```
t_utc = 1723456789.9999999
DateTime.from_epoch(t_utc).to_python_datetime(`
```

Output
```
ValueError: microsecond must be in 0..999999
```

Improve `to_python_datetime`: 1) improved performance by caching the static `py_datetime` import 2) avoid microsecond overflow crash when converting to python datetimes 3) for similar reasons, avoid leap second crash (checks if sec_int >=60 and clamps it to 59)

### DIFF
--- a/src/tudatpy/astro/time_representation/expose_time_representation.cpp
+++ b/src/tudatpy/astro/time_representation/expose_time_representation.cpp
@@ -830,10 +830,10 @@ In this example, the calendar date corresponding to when 122 days have passed in
                             microsecond = 0;
                         }
 
-                        // Handle potential leap second
+                        // Python datetime cannot represent leap seconds.
                         if( sec_int >= 60 )
                         {
-                            sec_int = 59;
+                            throw py::value_error( "Cannot convert a Tudat DateTime containing a leap second to Python datetime." );
                         }
 
                         // Call the cached constructor
--- a/src/tudatpy/astro/time_representation/expose_time_representation.cpp
+++ b/src/tudatpy/astro/time_representation/expose_time_representation.cpp
@@ -815,17 +815,35 @@ In this example, the calendar date corresponding to when 122 days have passed in
             .def(
                     "to_python_datetime",
                     []( const tba::DateTime& self ) {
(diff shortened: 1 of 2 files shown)


DominicDirkx:
@luigigisolfi Since this is Python-only, is this covered by the (smallish) unit test coverage? Also, could you add a unit test for this new fix?

luigigisolfi:
@DominicDirkx I have added a small unit test, which should suffice. 
It tests that these dates:

```
        DateTime(
            2024, 12, 31, 23, 59, 59.9999996
        ),  # probe (with microsecond overflow, last day of year)

        DateTime(
            2024, 12, 30, 22, 51, 59.9999996
        ),  # probe (with microsecond overflow, random day of year)
        
```
        are converted into:
`                datetime(2025, 1, 1, 0, 0, 0, 0),  # expected result`
`                datetime(2024, 12, 30, 22, 52, 0, 0),`

src/tudatpy/astro/time_representation/expose_time_representation.cpp
```diff
@@ -815,17 +815,35 @@ In this example, the calendar date corresponding to when 122 days have passed in
...
+
+                        if( microsecond == 1000000 )
+                        {
+                            normalizedDateTime = normalizedDateTime.addSecondsToDateTime( 1.0L );
+                            sec_int = static_cast< int >( normalizedDateTime.getSeconds( ) );
+                            microsecond = 0;
+                        }
+
+                        // Handle potential leap second
+                        if( sec_int >= 60 )
```
DominicDirkx, line 834: This is here because Python datetime cannot handle leap seconds right? I'd recommend we throw an error if this happens, since then we cannot be confident in the conversionn (datetime with leap second to one without)

DominicDirkx:
@luigigisolfi This closes https://github.com/tudat-team/tudatpy/issues/943 right?

PR #938, opened 2026-08-20
Add LOS conversion for DSN range and Doppler residuals

### LINKED ISSUE #872: [F&E]: Simplify retrieving radiometric residuals in state space
### Description:

The observation collection stores the residuals in the dimension of the observation themselves (i.e., range units and Hertz for range and Doppler), which can directly be retrieved from the observation collection through e.g. the `get_concatenated_residuals` method.
In order to retrieve the residuals in the state space of the spacecraft (e.g. position and velocity in line of sight direction) the user has to iterate over the single observation sets, retrieve the range unit to meter conversion factor from the ancillary settings and then apply the conversion manually. In the case of Doppler observations, the conversion factor cannot be retrieved directly but needs to be computed manually at the moment.

_No response_

### Additional context:

See https://deepspace.jpl.nasa.gov/dsndocs/810-005/202/202E.pdf section 2.2 for the conversion from frequency to velocity for one- and n-way Doppler

Strtus:
Closes #872

DSN range residuals are stored in range units and DSN Doppler in Hz. Getting those into metres / m/s currently means looping over each `SingleObservationSet`, reading `range_conversion_factor` from the ancillary settings, and (for Doppler) computing the scale by hand.

This stores a `doppler_conversion_factor` next to the existing range one, using the one-way / two-way scaling from DSN 810-005 202E §2.2 (`c/f_C` and `c/(2 f_C)`). For n-way averaged Doppler, `f_C` is the downlink reference carrier `M_{2,R} f_ref` already available from the ODF ancillary data.

`ObservationCollection` then has `get_residuals_in_state_space` / `get_concatenated_residuals_in_state_space`, same parser behaviour as the existing residual getters. Native-unit residuals used in the estimation are unchanged.

First cut is limited to types where the units are unambiguous: `dsn_n_way_range`, `dsn_n_way_averaged_doppler`, and measured-frequency Doppler when a set-level factor or reference frequency is present. `one_way_doppler` / `two_way_doppler` are left alone (their units depend on `normalizeWithSpeedOfLight`). Mixed collections with an unsupported set raise rather than converting only part of the vector.

## Test plan

- [x] `test_observation_models_RadiometricStateSpaceResiduals` (range, n-way Doppler, one-way frequency Doppler, unsupported types, collection order)
- [x] `tests/test_tudatpy/test_radiometric_state_space_residuals.py`


DominicDirkx:
Similar to https://github.com/tudat-team/tudatpy/pull/939 I think the solution is too invasive for the issue we're solving. Specifically, adding a function at the top-level (to the ObservationCollection) to support a single data type conversion does not seem to be the obviously best solution.  Let's first discuss in the issue before going to implementation. A better alternative (I think) would be to have a function that converts the range-unit residuals to meter-residuals, without changes to top-level interfaces

PR #900, opened 2026-07-15
Refactoring of data input layer of Tudat

DominicDirkx:
Closes Tudatpy side of https://github.com/tudat-team/tudatpy/issues/882

PR #939, opened 2026-08-20
Allow choosing transponder vs ancillary link-end delays

### LINKED ISSUE #937: Allow for more granular choice of link-end/transponder delays
### Description:

#854 added functionality to set the transponder delay in the VehicleSystems, as the preferred way over setting it in the ObservationCollection (through the link end delays). If link end delays are however present in the observation model through the ancillary settings, they will take precedence over the transponder delay, regardless of their value.
This is an issue, since in the current interfaces for ODF and TNF files, the link end delays are always taken and defined from the file, even if they are 0.0. This results in a warning ("Warning when computing observation: transponder delay functions are present in the observation model, but retransmission delays are provided through ancillary settings. The ancillary settings will be used. This warning is printed only once for this light-time calculator.") and a discarding of the transponder delay.

Since, with the current implementation of the interfaces, the link end delays will always be defined, regardless of their actual values, it should be possible for the user to choose which delay is used.

As a temporary fix, the old `.set_transponder_delay` method of the observation collection works fine if the transponder delay should be used instead of the link end delay at the spacecraft that is read from the file.

_No response_

### Additional context:

_No response_

Strtus:
Closes #937

ODF/TNF observations always carry a full `link_ends_delays` vector, so the VehicleSystems transponder delay from #854 never gets used, even when the spacecraft entry is just 0. Station delays from the file should still be applied.

This keeps those station delays and lets the vehicle-systems delay be used at the spacecraft instead, either from the body (`transponder_delay_overrides_ancillary_settings`) or per observation set (`set_transponder_delay_overrides_ancillary_settings`).

### DIFF
--- a/include/tudat/astro/observation_models/lightTimeSolution.h
+++ b/include/tudat/astro/observation_models/lightTimeSolution.h
@@ -19,6 +19,7 @@
 #include <iomanip>
 #include <map>
 #include <numeric>
+#include <string>
 #include <vector>
 
 #include "tudat/basics/basicTypedefs.h"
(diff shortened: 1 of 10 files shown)


Strtus:
The Windows job was red on `ConnectTimeout` to `minorplanetcenter.net` in the `remote_data` MPC tests, not on the delay/transponder changes (Linux, macOS, and the C++ tests were already green).

I pushed a small follow-up so those `remote_data` tests skip on connection timeouts instead of failing the job. Assertion failures and HTTP errors from an actual response are still treated as failures.

DominicDirkx:
To me it seems like too large of a change to the code for the issue that is being addressed (?) I think a better solution would be to:
* Set the transponder delay in the vehicle systems before loading the ODF/TNF data. 
* When loading the ODF/TNF data and creating the ancillary settings, deciding whether there is any useful delay data that is in the ODF/TNF files (e.g. are any non-zero), and deciding then whether to create the link end delays in the ancillary settings at all.

@larshinueber what do you think?

@Strtus We greatly appreciate your contributions to the code! But, I would recommend that, for issues like this where the solution to the issue being tackled is not obvious and multiple fundamentally different designs are possible, to first start a discussion in the issue before opening a PR.

DominicDirkx:
I think this issue should be tackled in a different way that is less invasive in the models once #900 is merged. I am closing this PR

PR #941, opened 2026-08-21
Add OpikEncounter and EphemerisOpikEncounter classes

luigigisolfi:
Relates to issue #1038.

This PR adds two new Tudatpy classes: `OpikEncounter` and `EphemerisOpikEncounter`.
It is an implementation of the [Opik formalism for **close** encounters. ](https://ui.adsabs.harvard.edu/abs/1951PRIA...54..165O/abstract)

Main assumptions in Opik theory:

1) Asteroid's orbit is a heliocentric ellipse
2) at the encounter time, asteroid's moves on a hyperbolic orbit
3) if encounter is close, assume Earth and asteroid are found at the same point at the encounter time, i.e. MOID = 0 . Then the asteroid can be assumed to move along a straight line (basically, the incoming hyperbolic asymptote)
4) most particles in the solar system with undergo multiple close encounters during their lifetime, this effectively leads to randomization of the relative velocity directions, thus a statistical reasoning can be applied if specific encounter geometry is not known.

Main result based on such assumptions:

1) post-encounter parameters [a,e,i] can be derived from the pre-encounter ones. 
2) a surface of variations can be computed, assuming all possible encounter geometries (see [Carusi et al. 1990](https://link.springer.com/article/10.1007/BF00050709))

If, on the other hand, an ephemeris is available, the actual values can be computed through the same theory. This is what, starting from Tudat's ephemeris, `EphemerisOpikEncounter` allows to compute.

### DIFF
--- a/src/tudatpy/astro/close_encounters/opik_encounters.py
+++ b/src/tudatpy/astro/close_encounters/opik_encounters.py
@@ -1,0 +1,362 @@
+import numpy as np
+import math
+from tudatpy import constants
+
+
+class OpikEncounter:
+    """
+    Idealized Opik encounter. Given (a, e, i), the planet's mass/radius/orbit,
+    and this encounter's b and psi, everything else -- theta, phi, U, gamma,
+    theta', phi', a', e', i' -- follows analytically. No ephemerides needed.
+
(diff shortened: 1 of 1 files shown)


luigigisolfi:
This PR is in draft because I will write some unit tests

PR #942, opened 2026-08-22
Allow custom interpolator settings for estimation ephemerides

### LINKED ISSUE #932: [F&E]: Change the interpolation order of ephemerides for an estimation
### Description:

By default, the state history is interpolated using a 6th order Lagrange interpolator to get the states at observation times, to compute the observables. For my use case, this interpolation is a substantial source of error if I choose a standard integrator like RK7(8) or DP, even with the tightest tolerances. Interpolation error can be at the level of a kilometer of more, which translates to more than a miliarcsecond at 1 AU (so above Gaia astrometry level for example). This forces me to use a lower order integrator with a very tight tolerance or small step sizes, that is significantly slower.

So I propose that an option is added where you can specify the interpolation order that is used in the interpolation, to reduce the interpolation error during estimation. Or if this is not possible, to at least increase the current interpolation order to 8 or 10 for example.

One option could be to add an optional argument for the interpolation settings to the EstimationInput object.

### Additional context:

_No response_

Strtus:
Closes #932

Integrated states are interpolated to observation times with a 6th-order Lagrange interpolator. For RK7(8)/DP that interpolation can be the dominant error (km-level), which is well above e.g. Gaia-level astrometry.

`EstimationInput` and `CovarianceAnalysisInput` now take optional `interpolator_settings` (same object as elsewhere, e.g. `lagrange_interpolation(8)`). Leave it unset and the old 6th-order interpolator is used.

The same setting is on `processing_settings`, since that is what actually writes the integrated states into the environment.


DominicDirkx:
@Strtus Thanks for the PR! I'm making a couple of small changes
*  PropagatorProcessingSettings is now the single source of truth for the interpolator settings. The previous sixth-order Lagrange behavior is defined there as the default. The low-level interpolator factory now only constructs the requested interpolator and no longer contains its own fallback policy. This makes behaviour more transparent for the yser

 * The duplicate setting has been removed from CovarianceAnalysisInput and EstimationInput, together with the corresponding Python API and the forwarding code through the orbit-determination manager,
  variational-equations solvers, and dynamics simulator. Estimation therefore uses the interpolator configured directly on the associated propagation processing settings.

*  The Python documentation and unit tests have also been updated. The tests verify the exact default configuration and confirm that non-default settings reach the generated environment ephemerides after:

  - single-arc propagation and estimation reintegration;
  - multi-arc propagation;
  - both the single- and multi-arc portions of hybrid-arc propagation.

I have not run all modified unit tests locally. Let's see what the CI does

include/tudat/simulation/propagation_setup/propagationProcessingSettings.h
```diff
@@ -42,7 +42,13 @@ class PropagatorProcessingSettings
                                   const bool setIntegratedVariationalResult = true ):
         clearNumericalSolutions_( clearNumericalSolutions ), setIntegratedResult_( setIntegratedResult ),
         createStateProcessors_( setIntegratedResult ), updateDependentVariableInterpolator_( updateDependentVariableInterpolator ),
-        setIntegratedVariationalResult_( setIntegratedVariationalResult )
+        setIntegratedVariationalResult_( setIntegratedVariationalResult ),
+        interpolatorSettings_( std::make_shared< interpolators::LagrangeInterpolatorSettings >(
```
Strtus, line 48: Ah yeah,much clearer than having the fallback buried down in the interpolator creation

tests/test_tudat/include/tudat/simulation/estimation_setup/executeEarthOrbiterParameterEstimationTestCase.h
```diff
@@ -40,7 +41,9 @@ Eigen::VectorXd executeEarthOrbiterParameterEstimation(
         const int numberOfDaysOfData = 3,
         const int numberOfIterations = 5,
         const bool useFullParameterSet = true,
-        const bool saveDesignMatrix = true )
+        const bool saveDesignMatrix = true,
+        const std::shared_ptr< interpolators::InterpolatorSettings > integratedStateInterpolatorSettings = nullptr,
+        std::pair< int, int >* integratedStateInterpolatorOrders = nullptr )
```
DominicDirkx: why is this a raw pointer? If it's needed as secondary return, make it a reference instead

tests/test_tudat/src/astro/propagators/unitTestIntegratedStateInterpolator.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "tudat/simulation/propagation_setup/setNumericallyIntegratedStates.h"
+
+namespace tudat
+{
+namespace unit_tests
+{
+
+BOOST_AUTO_TEST_SUITE( test_integrated_state_interpolator )
+
+BOOST_AUTO_TEST_CASE( testCreateStateInterpolatorOrder )
```
DominicDirkx, line 29: For each test case, add a comment describing what is being tested, and also add a comment per (group of) test statements to make the specific test clear (for all tests in this PR)

DominicDirkx:
It's almost ready, I think! One small stylistic code point, and some comments should be added to the tests. After that, I think this is ready

Strtus:
Was tied up with some lab work the past couple of days and didn't get to these in time — thanks for taking care of them.

DominicDirkx:
No worries! Thanks for initiating this PR and taking care of the bulk of it!

PR #914, opened 2026-07-29
Implement quadruple precision in C++

DominicDirkx:
This PR adds optional high-precision state support to Tudat’s C++ interfaces through the definition of a new HighPrecisionStateScalar, using Boost cpp_bin_float_quad by default, with long double as an alternative. Closes C++ part of https://github.com/tudat-team/tudatpy/issues/1031

The changes update propagation, ephemerides, interpolation, integrators, and observation models to support the configured scalar. Runge–Kutta coefficients are constructed at the selected precision, and time conversions and DSN Doppler calculations better preserve small intervals.

 Unit tests cover state storage and interpolation, integrator convergence, orbit propagation, range and DSN Doppler observations, and conversion of small negative time intervals. 

 High-precision support remains optional. Python scalar exposure is unchanged (added here: https://github.com/tudat-team/tudatpy/pull/945), and Time retains its split representation with a long double fractional component.

PR #945, opened 2026-08-24
Python exposure of quad precision

### LINKED ISSUE #1031: [F&E]: Implement support for better-than-double precision in Tudat
### Description:

At the moment, all computations in Tudat are done in double precision. There is support for long double, but on modern Macs and Windows, this is equivalent to double (64 bit) and only only Linux and older Macs does it use extended precision (80 bit)

For a number of applications, including high-accuracy solar-system scale applications and low compression-time Doppler observables, the use of double precision can become a fundamentally limiting factor.

While higher-than-double precision should be optional (since it does incur an overhead), it should be supported in an easy manner

Use boost multiprecision's quadruple precision (128 bit) to allow the following quantities to be in higher precision:
* State vectors during propagation
* States stored in ephemerides and returned from interpolators
* Values computed from observation models
covering the same quantities that now have long double support. I propose we leave the Time class as it is. It has sub-ps resolution over long time scales

Note that the proposal leaves accelerations and other quantities in the state derivative in double precision, but this is typically not the source of the noise floor in the current set up 

### Additional context:

_No response_

DominicDirkx:
Python exposure to complement #914, closes https://github.com/tudat-team/tudatpy/issues/1031.

The setup allows the user to select whether to use double or quadruple precision at the top of their Python script. Everything on the Python side remains in double precision, but choosing quadruple precision toggles this in the core calculations in C++, significantly reducing the numerical noise floor.

In the current implementation, the Python exposure is built twice: once with double and once with quadruple precision. The user can choose (at the top of their script) which one to use. This means that the user interfaces for double and quadruple precision are identical.

The heavy price to pay for this is the fact that this leads to a significant overhead for the compilation: compiling the kernel with quad precision is even slower than for double precision.

I would propose to disabe this option for the standard CI, and only run the quad precision CI weekly (when bumping the version?). I also propose we hold off on adding quad precision to the conda packages until everything is migrated to Github, and we can revisit this choice then


include/tudat/astro/observation_models/dsnNWayAveragedDopplerObservationModel.h
```diff
@@ -207,7 +205,7 @@ class DsnNWayAveragedDopplerObservationModel : public ObservationModel< 1, Obser
         std::vector< double > arcEndLinkEndTimes;
         std::vector< Eigen::Matrix< double, 6, 1 > > arcEndLinkEndStates;
 
-        TimeType integrationTime;
+        double integrationTime;
```
DominicDirkx: why is this turned into a double?

include/tudat/astro/observation_models/dsnNWayAveragedDopplerObservationModel.h
```diff
@@ -237,7 +235,8 @@ class DsnNWayAveragedDopplerObservationModel : public ObservationModel< 1, Obser
         FrequencyBands downlinkBand = frequencyBands.at( 1 );
 
         // Set approximate up- and down-link frequencies.
-        ObservationScalarType currentTurnAroundRatio = static_cast< ObservationScalarType >( turnaroundRatio_( uplinkBand, downlinkBand ) );
```
DominicDirkx, line 240: why is the turn around ratio not computed directly as high-prcision? This is very fishy

include/tudat/astro/observation_models/dsnNWayAveragedDopplerObservationModel.h
```diff
@@ -306,14 +305,38 @@ class DsnNWayAveragedDopplerObservationModel : public ObservationModel< 1, Obser
         ObservationScalarType transmitterFrequencyIntegral =
                 frequencyInterpolator_->template getTemplatedFrequencyIntegral< ObservationScalarType, TimeType >( transmissionUtcStartTime,
                                                                                                                    transmissionUtcEndTime );
+        const ObservationScalarType integrationTimeScalar = static_cast< ObservationScalarType >( integrationTime );
```
DominicDirkx: I dont understand why this added part is needed

include/tudat/astro/observation_models/dsnNWayAveragedDopplerObservationModel.h
```diff
@@ -237,7 +235,8 @@ class DsnNWayAveragedDopplerObservationModel : public ObservationModel< 1, Obser
         FrequencyBands downlinkBand = frequencyBands.at( 1 );
 
         // Set approximate up- and down-link frequencies.
-        ObservationScalarType currentTurnAroundRatio = static_cast< ObservationScalarType >( turnaroundRatio_( uplinkBand, downlinkBand ) );
+        const double currentTurnAroundRatioAsDouble = turnaroundRatio_( uplinkBand, downlinkBand );
+        ObservationScalarType currentTurnAroundRatio = static_cast< ObservationScalarType >( currentTurnAroundRatioAsDouble );
         ObservationScalarType currentReferenceTurnAroundRatio =
```
DominicDirkx, line 241: why is this computed as double? this was also done originally, but it is not right. the turn around ratio is an exact ratio of integers and can be computed in high precision

include/tudat/astro/observation_models/lightTimeSolution.h
```diff
@@ -189,6 +193,53 @@ bool isMultiLegLightTimeSolutionConverged( const std::shared_ptr< LightTimeConve
     return isToleranceReached;
 }
 
+namespace light_time_solution_detail
+{
+template< typename TimeType >
+struct EpochIntervalArithmetic {
+    template< typename ScalarType >
+    static TimeType add( const TimeType& epoch, const ScalarType& timeInterval )
```
DominicDirkx: These are things that should be added to the Time class, I think. Not here (whole block)

PR #953, opened 2026-08-25
Add batch SBDB catalogue queries

### LINKED ISSUE #940: [F&E]: Make JPL SBDB interface suitable for large batches of objects
### Description:

Currently the SBDBquery functionality is quite slow and often throws a timeout error when attempting to get the data for multiple asteroids/comets. This could be fixed quite easily by loading the entire SBDB database up front through the SBDB Query API (https://ssd-api.jpl.nasa.gov/doc/sbdb_query.html) and filtering the requested rows from that. For large amounts of objects, this is significantly faster and more robust

Add a new class alongside SBDBquery called "SBDBbatch", for example, that loads all minor body data into a dataframe and reads from this. There could also be an option to load from a local CSV version of the table, so that no internet connection is required.

### Additional context:

_No response_

DominicDirkx:
## Summary

  Add `SBDBbatch` for downloading, filtering, saving, and reloading large JPL SBDB catalogues.

  ## Validation

  - Added tests for pagination, filtering, and CSV loading
  - 2 tests passed

  Closes #940.


DominicDirkx:
This PR should be modified so that it matches the new data input layer (#900)

PR #955, opened 2026-08-26
Avoid unnecessary resource archive downloads

HannoSpreeuw:
# Summary
Improve `resource_installer --mode missing` by checking installed resources before obtaining their corresponding archives.

Before this change, `--mode missing` already reused valid archives from `TUDATPY_RESOURCE_CACHE`, avoiding network downloads when the archive cache was populated. However, the installer obtained, verified, and opened each archive before checking its independently verifiable installed resource files.

Because every installed resource has its own SHA-256 checksum, archives whose resources are already present and valid can be skipped entirely.

# Changes

- Check installed resources before obtaining their corresponding archive.
- Consider a resource to require installation when it:
  - is missing; or
  - fails its SHA-256 verification.
- Skip an archive when all associated resources are present and valid.
- Obtain an archive from the local cache or Zenodo only when at least one associated resource requires installation.
- Extract only the resources that require installation.
- Preserve the existing behaviour of scratch and update modes.

# Behaviour before this change
With a valid, populated archive cache:

- archives were reused without network downloads;
- each archive was still verified and opened before the installed resources were checked.

Without a cached archive:

- the complete archive was downloaded before the installed resources were checked;
- this could result in large downloads even when every resource from that archive was already installed.

# Behaviour after this change
The installer first verifies the installed resource files using their individual SHA-256 checksums.

- If all resources associated with an archive are present and valid, the archive is skipped completely.
- If one or more resources are missing or invalid, the installer obtains the archive from the local cache or Zenodo, verifies it, and extracts only the required resources.
- A complete archive must still be downloaded when at least one of its resources is required and no valid cached archive is available.

This is particularly beneficial for large archives such as spice_kernels.tar.gz, which is approximately 1.2 GB.

# Testing
Added regression tests covering:

- skipping an archive when all associated resources are present and valid;
- extracting only missing resources from a required archive;
- obtaining an archive when an installed resource fails checksum verification.

All relevant resource-installer and resource-catalog tests pass:
> 23 passed

### DIFF
--- a/src/tudatpy/resource_installer.py
+++ b/src/tudatpy/resource_installer.py
@@ -309,6 +309,22 @@ def _extract_tarball_members(
     return installed, failures
 
 
+def _required_tarball_targets(
+    targets: List[str], dest_root: Path, hashes: Optional[Dict[str, str]] = None
+) -> List[str]:
+    """Return archive members that are absent or fail their checksum."""
+    required = []
+    for target in targets:
+        dest_path = dest_root / target
+        if not dest_path.exists():
+            required.append(target)
+            continue
+        if hashes and (expected_hash := hashes.get(target)):
+            if not _verify_sha256(dest_path, expected_hash):
+                required.append(target)
+    return required
+
+
 def install_files(
     files: Dict[str, str],
     dest_path: Path,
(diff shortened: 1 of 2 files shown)


DominicDirkx:
@HannoSpreeuw I can't see any issues in this addition, but I also don't have the whole setup of these scripts properly in my head. But, I don't see any risk or issue in merging this. One question: could you add some more comments to teh code in this script so that the logic will still be clear for other developers in the future?

HannoSpreeuw:
> One question: could you add some more comments to teh code in this script so that the logic will still be clear for other developers in the future?

All right, I think that commit [1841838](https://github.com/tudat-team/tudatpy/pull/955/commits/1841838f51f68a0e9324d7072b23e953441ff750) addresses that.

PR #956, opened 2026-08-26
Add new constrainedTranslationalStateParameter

luigigisolfi:
With this PR (written with the help of AI to navigate the architecture of Tudat) we add a template for a `constrainedTranslationalStateParameter` that inherits from the `initialTranslationalState` and allows to trigger a least squares solver with a constraint of the type:

`G \DeltaX = d` 

where `G` is a 1x6 vector, and `d = 0`.

the least squares detects the constraint and modifies the original (unconstrained) system of equations according to Lagrange Multiplier method:

```
N \Delta x + G^T\lambda
G\Delta X = 0
```
The above system is solved for both `\lambda` and `\Delta x` and, at convergence, yields a correction to x, $$\Delta x$$, which is orthogonal to the vector G. 

If G is assumed to be the weak direction vector of the covariance of an unconstrained Least Squares Nominal Solutions, then this parameter allows to find points along the Line of Variations, as explained in [Milani et al, 2005](https://www.researchgate.net/publication/222652291_Nonlinear_impact_monitoring_Line_of_variation_searches_for_impactors).

**Note 1:** since the new parameter inherits from `initialTranslationalState`, a flag was added to the constructor of  `initialTranslationalState` to define the correct identifier for the parameter, as `initialTranslationalState` had a hardcoded one that would make it impossible for tudat to recognize our new parameter (and this would potentially apply to any other defined parameter that inherited from `initialTranslationalState`), see https://github.com/tudat-team/tudatpy/pull/956/commits/52ad9a0408d4ab7f784941918496dbe11a427b75
**Note 2:** some equality checks needed to be able to estimate the new parameter, see https://github.com/tudat-team/tudatpy/pull/956/commits/4f21905216a7043a843cc64b6343cff10c1fc5dd

### DIFF
--- a/include/tudat/astro/orbit_determination/estimatable_parameters/constrainedTranslationalState.h
+++ b/include/tudat/astro/orbit_determination/estimatable_parameters/constrainedTranslationalState.h
@@ -24,8 +24,26 @@ template< typename InitialStateParameterType = double >
 class ConstrainedTranslationalStateParameter : public InitialTranslationalStateParameter< InitialStateParameterType >
 {
 public:
-    //! Inherit constructors from the base class
-    using InitialTranslationalStateParameter< InitialStateParameterType >::InitialTranslationalStateParameter;
+    //! Constructor, sets initial value of translational state.
+    /*!
+     * Constructor, sets initial value of translational state. Explicitly forwards to the base class constructor with
+     * parameterType set to constrained_initial_body_state, so that this parameter correctly identifies itself as distinct
+     * from a plain InitialTranslationalStateParameter (e.g. via getParameterName()).
+     * \param associatedBody Body for which initial state is to be estimated.
+     * \param initialTranslationalState Current value of initial state (w.r.t. centralBody)
+     * \param centralBody Body w.r.t. which the initial state is to be estimated.
+     * \param frameOrientation Orientation of the frame in which the state is defined.
+     */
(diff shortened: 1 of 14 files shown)


luigigisolfi:
@DominicDirkx got some improvements before tomorrow :)

luigigisolfi:
C++ tests unaffected, even upon normalization of constraint vector (G), although tests still need to be written for this functionality.

**Note (based on some examples I wrote for Eros):**
All test pass, but I suppose the quaternion example was passing WITHOUT any normalization term because the quaternions are already O(1), since they are all dimensionless by definition. However, when dealing with a physical constraint such as the one along the LOV, the weak direction has mixed units (e.g. position, meters and velocity, m/s), and it can happen that one of these components of G differs by the other by orders of magnitude. As a result, G is not necessarily O(1) and we need to normalize it properly with the same normalization terms used to normalize the design matrix, or the problem will mix dimensions. 

In other words, the quaternion test is unaffected because the quaternion components of the constraint are the same physical quantity, and the normalization rescales them more or less in the same way. If for any reason one of the quaternions were to scale differently from the others, we would have likely found a bug. 


# This closes #958

DominicDirkx:
@luigigisolfi Rather than creating a new initial state parameter (which could have unintended side effects) I think it would be a more robust approach to add an option to more easily add the right kind of constraint to the existing single-arc initial state parameter. What do you think?

Issue #893, opened 2026-07-10
Commented-out unit tests


larshinueber:
There is a number of commented out unit tests in the CMakeLists.txt without apparent reason.
These include, for example, DsnOdfGrail, DsnOdfGrailStatistics, EstimationFromDsnData, EstimationFromSpiceData, HybridArcMultiBodyEstimation and number of others.

These should be re-enabled and either fixed or assigned an issue why they cannot be re-enabled at the moment.

PR #957, opened 2026-08-26
Feature/issue 893 multi hybrid tests

DominicDirkx:
### Summary

  Re-enable and strengthen four multi-/hybrid-arc regression tests with numerical checks for states, STMs, sensitivities, interface composition, parameter recovery, and residuals.

  Also remove obsolete disabled tests and the unused, outdated full-propagation comparison helpers.

  ### Validation

  - Four strengthened regression tests pass.


DominicDirkx:
Closes #893

PR #965, opened 2026-08-27
Feature/is it dark at ground station

luigigisolfi:
This PR implements a geometric check informing the user about whether a given target is visible from a ground station.
This function can be used outside of a Tudat simulation (unlike what already exists for the elevation angle viability settings, which allows to do the same but only upon simulating observations). 

In particular, `isTargetVisibleFromGroundStation` uses the `pointingAnglesCalculator` to check whether the target is above some minimum elevation. 

When the target is the Sun, and the elevation is - say - the nautical twilight (i.e. 6 degrees below horizon), this function allows to check whether the ground station is in daylight or not (useful for SSA simulations, for which angles-only measurements from optical telescopes are usually available).

Users can also use `isItDarkAtGroundStation`, which is a dedicated wrapper allowing to do exactly that, without the need to extract the positions from the Sun and Ground Station ephemeris.


luigigisolfi:
@DominicDirkx do you think it is feasible to now add this to the viability settings?

DominicDirkx:
@luigigisolfi Yes, definitely! Do you need input on adding it as a viability setting option?

luigigisolfi:
I have never done it so I’ll have to take a look at how it’s done first, but if you have any specific ideas already on things like nomenclature, architecture, etc just let me know! 

Two questions about the existing code in this PR: 

1) do the input arguments look good to you? 
2) does the dedicated function to assess daylight/dark conditions make sense? (i.e for instance hardcoding the target to be the sun, and having “Sun” as hardcoded string corresponding, indeed, to the sun itself)

for instance the second function does not take inertial states directly, but it computes them inside the function. I think it’s good but wanted to double check with you before proceeding.

include/tudat/astro/ground_stations/pointingAnglesCalculator.h
```diff
@@ -149,6 +149,26 @@ double calculateGroundStationAzimuthAngle( const std::shared_ptr< PointingAngles
...
+ * target and negating the result (or comparing against a maximum instead) yields a day/night check at the station.
+ * \param groundStationPosition Inertial position of the ground station.
+ * \param targetPosition Inertial position of the target.
+ * \param timeTdb Time (TDB seconds since J2000) at which the check is to be performed.
+ * \param minimumElevationAngle Elevation angle [rad] at or above which the target is considered visible.
+ * \param pointingAnglesCalculator Object used to compute the elevation angle at the ground station (as retrieved from
+ * GroundStation::getPointingAnglesCalculator).
+ * \return True if the target's elevation angle at the ground station is at or above minimumElevationAngle, false otherwise.
+ */
+bool isTargetVisibleFromGroundStation( const Eigen::Vector3d& groundStationPosition,
```
DominicDirkx, line 166: The implementation of this is identical to that of the elevation angle viability condition, but then with the Sun as target rather than the spacecraft (?) Perhaps using the existing code for this would be easier than redefining the calculations (finding a clever way to swap out the target with the Sun)

EDIT: to comment on my previous point: these functions are useful to have regardless of whether they're implemented as viability conditions. They can be called directoy by the user like this, which is useful

luigigisolfi: Indeed, it's basically the same functionality but applied to an extra target, not to the main target. I dont know whether this fits what you have in mind, but I think we should implement this viability setting in such a way that the distinction is clear between the target and the "extra-target" we want to observe within a simulation.

PR #988, opened 2026-09-02
Add TDM, OEM, OMM Parser/writer + TDM to tracking data conversion

luigigisolfi:
This PR builds on the data-refactor branch to provide TDM , OEM, OMM Support. 
Additionally, TDMs can be converted into trackingData objects.


luigigisolfi:
Not fully convinced about this PR. Using an already existing library (like [ccsds-ndm-py](https://github.com/JochimMaene/ccsds-ndm)) would be better and allow us to get parsers for any ccsds standard for free, including xml <-> kvn conversions, while at the same time minimizing our code. Working on it.

DominicDirkx:
I think using such an existing library would indeed be a better approach. I'll close this PR since it would then be superseded (you can reopen if I am incorrect)

luigigisolfi:
@DominicDirkx considering their Licenses, which one looks better suited for tudat to you?

https://github.com/egemenimre/ccsds-ndm (GNU General Public License v3.0)
https://github.com/JochimMaene/ccsds-ndm (MPL-2.0-1)

DominicDirkx:
Definitely the second one (if I understand the differences correctly) since the GPL would require us to make all of tudat gpl

PR #989, opened 2026-09-02
Update viability settings for (shadow function)

luigigisolfi:
As discussed with @DominicDirkx , the condition for occultation in the new  `body_in_sunlight_viability` settings was updated to use `computeShadowFunction`. This allows to define a (user-defined) `minimum_shadow_function_value` below which the satellite is visible. 

If  `minimum_shadow_function_value` = 0.5, we are allowing the taerget to be visible up to when the 50% of the Sun's area (sun disk) is occulted by Earth.

Default `minimum_shadow_function_value` is set to 1, namely observations are retained only when the object is fully illuminated.


DominicDirkx:
Nice, thanks!

PR #990, opened 2026-09-04
Add interface for NASA ATDF files

larshinueber:
This PR adds an interface to parse NASA ATDF files (TRK-2-25) into TrackingData/TrackingSupplementaryData objects.
It is based on the `atdf2ascii` tool by Ashok Verma (see https://github.com/ashokverma24/atdf2ascii and the corresponding paper: https://doi.org/10.1016/j.softx.2022.101190), which is used to convert the binary ATDF files into ASCII tables, that are then read and converted to TrackingData objects.
For the moment I added my fork of atdf2ascii as dependency, since I had to make a few minor packaging changes to be able to install the tool as a Python package through pip. A PR on the original repository is open (https://github.com/ashokverma24/atdf2ascii/pull/1), once that is merged it should be used as dependency in the environment.yaml.

TODO before merging:
- [x] add atdf2ascii dependency in tudatpy-feedstock: https://github.com/tudat-team/tudatpy-feedstock/pull/37
- [ ] add option to silence atdf2ascii parsing output
- [x] add API reference examples for how to use the convenience method (read_atdf_data) and the Processor class (e.g. to go from ASCII directly to TrackingData without having to reparse the binary)
- [x] add paper citation in docstrings

For some comparisons of the plots published in https://doi.org/10.1016/j.softx.2022.101190 and pre-fits obtained with Tudat using the new interface, see the attached [ATDF Reader Tudat Integration.pdf](https://github.com/user-attachments/files/31837228/ATDF.Reader.Tudat.Integration.pdf).
Comments/reviews are appreciated @tudat-team/radio-science-dev :)


larshinueber:
@DominicDirkx I fixed all open points on the tudatpy side except for the silencing of the `atdf2ascii` output.
Since the tool uses multiprocessing and writes to the output from within the subprocesses it cannot simply be redirected from stdout. In my opinion it is not critical for this PR and would make sense to make it configurable upstream, so I would leave it as an open point for the future.
From my point of view this is ready to review now :)

A PR on the feedstock is open: https://github.com/tudat-team/tudatpy-feedstock/pull/37

src/tudatpy/data_input/tracking_data/atdf/_converters.py
```diff
@@ -0,0 +1,321 @@
...
+            tracking_data.add_double_ancillary_setting(
+                "Doppler observable integration time", float(ct)
+            )
+            tracking_data.add_double_vector_ancillary_setting("link ends time delays", list(ttd))
+            tracking_data_list.append(tracking_data)
+
+        return tracking_data_list
+
+
+class AtdfRampConverter:
```
DominicDirkx, line 279: question: how do you deal with multiple transmission frequencies (so multiple ramps) that overlap in time? (for instance S- and X-band)? At teh moment we don't yet support this in Tudat of course, but we should make sure that the input to Tudat does not merge multiple bands

tests/test_tudatpy/test_data_input_atdf.py
```diff
@@ -0,0 +1,689 @@
...
+}
+MSR_ROW_COUNT = sum(MSR_ROWS_PER_DATA_TYPE.values())
+NWAY_DOPPLER_ROW_COUNT = (
+    MSR_ROWS_PER_DATA_TYPE["2-Way-Doppler"] + MSR_ROWS_PER_DATA_TYPE["3-Way-Doppler"]
+)
+NWAY_RANGE_ROW_COUNT = MSR_ROWS_PER_DATA_TYPE["2-Way-Range"]
+RAMP_ROWS_PER_STATION = {"DSS-34": 746, "DSS-45": 629, "DSS-15": 179, "DSS-54": 10}
+RAMP_ROW_COUNT = sum(RAMP_ROWS_PER_STATION.values())
+
+FIXTURE_COUNT_TIME = 60.0
```
DominicDirkx, line 52: These are all quantities that you have taken from the files and hard-code here for the test?

larshinueber: Correct 👍

src/tudatpy/data_input/tracking_data/atdf/_converters.py
larshinueber: Good question, at the moment it is not dealt with at all and all ramps are just appended to the ramp histories, so for multiple transmitting frequencies later ramp records would just terminate the previous ones early. From what I can see that's the same for TNF and ODFs tho, right?

I think to make this work we would need a way for the frequency interpolator to keep track of multiple ramp tables for different bands at the same time (or have multiple frequency interpolators at the ground station level) and then request the frequency of a certain band from the observation model?

DominicDirkx: Indeed, this is an issue that is similarly present in the ODF and TNF file readers. So, I'd propose that we introduce it similarly in the ATDF file reader, and open an issue to make the frequency transmission objects 'band-aware' so that you can query transmission frequencies at different bands, and then patch up the file readers accordingly.

In short: for this PR we keep it as it is now :)

larshinueber: Sounds good, I'll open an issue :)

DominicDirkx:
All right! Merged :)

PR #994, opened 2026-09-10
Publish pip wheels by repacking the official conda package

kovar:
Repack conda `tudatpy` into pip wheels. No extra C++ build. Conda stays the recommended install.

**Does not run on commit / PR / push.** Triggers:

- `workflow_dispatch` — artifacts only, nothing published
- `release` published — builds and attaches `.whl` + `index.html` to that release

**First 1.0.0 wheels:** merge (or otherwise have `wheels.yml` on the tagged commit) → Releases → Draft a new release → tag **`v1.0.0`** → Publish. CI installs conda `tudatpy=1.0.0` (pinned in the workflow) and uploads the wheels onto that release. Merging this PR by itself does not publish anything.

```bash
pip install tudatpy --find-links https://github.com/tudat-team/tudatpy/releases/download/v1.0.0/index.html
```

`--find-links` needs a public repo (or `gh release download` if private).

Matrix: linux-64, osx-arm64 (`macos-14`), osx-64 (`macos-15-intel`), win-64 × cp310–312.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -1,0 +1,131 @@
+name: Wheels
+
+# Publishes pip/uv wheels by *repacking* the official conda tudatpy builds.
+# Does not compile C++.
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      only-os:
+        description: "If set, only this runner (ubuntu-latest, macos-14, macos-15-intel, windows-latest)"
+        required: false
+        default: ""
+      only-python:
+        description: "If set, only this Python (3.10, 3.11, 3.12)"
+        required: false
+        default: ""
+
+permissions:
+  contents: read
+
+jobs:
+  matrix:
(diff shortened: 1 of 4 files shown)


DominicDirkx:
@kovar Thanks a lot for adding this! We will need to discuss internally if we will want to support pip packages in this manner, so it may be a week or two until we get back to you.

DominicDirkx:
@kovar I'm afraid we'll have to decline the pull request. While we are very much in favour of the idea of having pip packages, when we do this we want to make sure we can support and maintain it. The code in this pull request in not clear enough (to us) for us to commit to maintaining this. If you want, you can open an issue on this, and propose design solutions on how to proceed on this. Then, we mihght be able to converge to a solution that the existing team can suport. I hope this does not discourage you, we're very happy to be seeing external contributors! But since this is a PR that might well involve a significant maintaneance burden, we are not able to merge it.

PR #995, opened 2026-09-11
Add regression test for barycentric Lagrange interpolation

Markus-Reichel:
The barycentric Lagrange interpolator introduced in #628 replaced an exact-equality guard on the independent variables with a tolerance-based check on the difference of every stencil point, cast to the scalar type of the interpolator. That change was not covered by a test.

Add a test case that constructs the situation the previous implementation failed on: a requested Time that differs from a tabulated Time by less than one ULP of its seconds-into-full-period member. The two epochs are not equal, but the normalization of their difference rounds the remainder to exactly one full period, so that the difference cast to the scalar type is exactly zero. The previous implementation evaluated the interpolating polynomial with a vanishing denominator and returned NaN; the barycentric implementation returns the tabulated value.

The test fails on the implementation preceding #628 and passes on the current one.

### DIFF
--- a/tests/test_tudat/src/math/interpolators/unitTestLagrangeInterpolators.cpp
+++ b/tests/test_tudat/src/math/interpolators/unitTestLagrangeInterpolators.cpp
@@ -11,9 +11,12 @@
 
 #define BOOST_TEST_MAIN
 
+#include <cmath>
+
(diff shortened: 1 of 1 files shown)


Markus-Reichel:
@DominicDirkx sorry I wanted to open a draft pull request, because I still want to add another test!

PR #997, opened 2026-09-11
Test/barycentric lagrange regression

Markus-Reichel:
**Add regression tests for the barycentric Lagrange interpolator**

#628 replaced the Lagrange interpolator with its barycentric form to fix the banding of #612, but added no test. This adds two to `unitTestLagrangeInterpolators.cpp`. One checks that a constant is reproduced bitwise exactly, the other that a `Time` query one ULP below a node no longer returns NaN.

Both fail on the code preceding #628  and pass on the current implementation.


DominicDirkx:
Looks good :) I found the comments a little too verbose (to teh point that they got a little confusing to understand what was really going on) and I slightly reduced them

PR #996, opened 2026-09-11
Test CI

DominicDirkx:
Test only: run CI for the PR #900 review fixes.

PR #1000, opened 2026-09-12
Minor corrections to data input refactor

DominicDirkx:
These changes are based on an AI-performed review of PR #900. The review points addressed here are those retained after iteration and discussion.

The retained findings are grouped into four initial commits, preserving the original review numbering:

**1. Container consistency, validation, and numerical precision — `274117696`**

- **F5:** Optical observation metadata was interpreted as simulation ancillary settings, causing collection conversion to throw. Separate observation metadata from simulation settings, and return no simulation-settings object when only metadata remains.
- **F10:** The frequency-ramp container discarded extended ODF epoch precision. Preserve ramp epochs as `Time` values.
- **F11:** Invalid weight or correction replacements could erase existing values. Validate all dimensions before replacing state.
- **F12:** Removing observations left per-row metadata misaligned. Remove the corresponding registered metadata entries.
- **F13:** The C++ vector-weight setter rejected valid vector observables. Validate one correctly sized weight vector per observation.

This commit also includes the follow-up ownership fix: Python frequency-ramp records are independent copies, so appending to the container cannot invalidate previously retrieved records.

**2. Observation values and weights — `a655eca26`**

- **F1:** Optical catalog biases were applied with the opposite sign. Store negative biases under the additive correction contract.
- **F2:** Requesting VFCC17 weights could discard required technique and catalog metadata. Retain these inputs whenever weights are requested.
- **F8:** Applying corrections rejected uncorrected inputs in mixed collections. Apply available corrections and accept inputs without corrections.

**3. Environment setup and object identity — `bd38daaba`**

- **F3:** Multi-file PSF conversion retained only the first camera history. Merge compatible histories and reject calibration or attitude conflicts.
- **F6:** Optical `custom_name` did not reach observation links. Resolve target names consistently while preserving original identifiers as metadata and MPC target order.
- **F9:** Piecewise-constant frequency supplementary data was ignored. Install it for stations and vehicles through the existing frequency calculator using zero-rate ramps.

**4. Backwards compatibility — `fc47a0017`**

- **F4:** Compatibility aliases did not preserve legacy methods, return types, or environment setup. Restore the MPC, Horizons, TNF, and OMMUtils behavior, including legacy TNF converters and station-dependent time conversion. The TLE helper methods belonged to `OMMUtils`; their original attribution to `SpaceTrackQuery` was corrected during review.

**Excluded after iteration — F7:** Nightly weighting can depend on how inputs are partitioned across batches. This limitation predates PR #900; an introduced regression was not established, so the proposed correction was reverted and is not included.

Replaces #996 with a revised commit history targeting `feature/data-refactor`, the branch used by PR #900. The tree after the four initial commits is identical to #996. Subsequent review follow-ups make deprecation warnings consistent, document removal of the compatibility layer, clarify test coverage, and fix metadata-only optical simulation (F5).

The catalogue-sign regression (F1) now uses 111 frozen MPC Eros observations in two intervals, with geometric JPL Horizons ephemerides and no dynamics integration. The 1996 McDonald series has a 0.613 arcsec correction RMS: residual RMS is 0.726 raw, 0.414 with subtraction, and 1.278 with addition. The 2016 ATLAS series gives 0.219, 0.140, and 0.320 arcsec respectively. Catalogue-coded 1898/1900 counterexamples and other candidate intervals are documented rather than omitted.

Validation: built this branch's Python kernel and passed 229 targeted optical, frequency, compatibility, and Eros tests. The Eros model uses IAU 2006/IERS Earth orientation, station time conversion, solid/pole tides, and solar light-time delay; ocean loading tides are omitted. Tudat/Horizons observer-angle agreement is better than 0.000015 arcsec. Reproduction scripts, frozen observations/state tables, model-sensitivity results, and concise descriptions of each test/assertion group are included.


include/tudat/io/trackingData.h
```diff
@@ -265,14 +294,6 @@ class TrackingData
     //! Set observation weights to the tracking data object (optional)
     void setObservationWeights( const std::vector< Eigen::Matrix< double, Eigen::Dynamic, 1 > >& observationWeights )
     {
-        // Check if observation weights already existed and overwrite them if they did (+throw a warning)
-        if( !weights_.empty( ) )
```
DominicDirkx, line 269: Removing this warning is not safe: users should get a warning when their data is overwritten (similar blocks exist below)

PR #1011, opened 2026-09-16
Support observation-only estimation with null propagator settings

### LINKED ISSUE #1006: [BUG]: Estimating only observation parameters is not possible
### Describe the issue:

`OrbitDeterminationManager` looks like it supports estimation runs where no dynamics is estimated. `propagatorSettings = nullptr` is passed and only parameters that affect observations should be estimated. There is an explicit else `if( propagatorSettings == nullptr )` branch for it.

That path cannot currently run. Three separate places reject the null settings object, and the third one is the function that the `nullptr `branch itself calls. So the branch throws on its own first line of work.

I want to add a per-image camera pointing correction: a small 3-vector rotation of a camera, estimated from optical landmark (pixel) observations. It is a pure observation-model parameter and it changes where a landmark lands on the image, and has no direct effect on the spacecraft trajectory. 
As a test I want to use a fixed SPICE trajectory and estimate only the camera pointing. This needs an estimation run with no dynamical parameters and no propagator (`propagatorSettings = nullptr`)

**What happens**
Constructing an `OrbitDeterminationManager` with a parameter set that contains no initial dynamical state, and `propagatorSettings = nullptr` throws:
```
Error when defining setIntegratedResult, dynamics type not recognized
```

Now this leads to a couple of issues:

1. `toggleIntegratedResultSettings` `simulation/propagation_setup/propagationSettings.h`, called from `initializeOrbitDeterminationManager` asks is this single or multi arc using `dynamic_pointer_cast`. A null pointer falls into the `else` and throws "dynamics type not recognized"
2. an unguarded dereference in `simulation/estimation_setup/orbitDeterminationManagerConstructionImplementation.h`
```
propagatorSettings->getOutputSettingsBase( )->setUpdateDependentVariableInterpolator( true );
```
which runs unconditionally a couple lines above the `nullptr` branch

3. `createStateTransitionAndSensitivityMatrixInterface` in `simulation/estimation_setup/createNumericalSimulator.h`
This is the function the `nullptr` branch calls:
```
else if( propagatorSettings == nullptr )
{
    stateTransitionAndSensitivityMatrixInterface_ =
            createStateTransitionAndSensitivityMatrixInterface< ... >(
                    propagatorSettings, parametersToEstimate_, 0, totalNumberParameters_ );
}
```
When `propagatorSettings = nullptr` it thows "Error when creating state transition interface, dynamics type not recognized".

**How to fix?**
The question becomes how to handle this, because it needs to be decided what an empty state transition interface means when there is no dynamics at all.

- which interface type should be build when there is no propagator?
- is a single-arc interface with `dynamicalStateSize = 0` enough?
- should the `nullptr` still be accepted or should we handle the "estimation of observations only" in a different way? We could make it an explicit supported mode in `OrbitDeterminationManager`

```
SystemOfBodies bodies( "SSB", "J2000" );
bodies.createEmptyBody< double, double >( "Earth", false );
bodies.at( "Earth" )->setGravityFieldModel(
        std::make_shared< gravitation::GravityFieldModel >( 3.986004418E14 ) );

// One non-dynamical parameter: no dynamics estimated, so no propagator needed.
const std::vector< std::shared_ptr< EstimatableParameterSettings > > parameterNames = {
    std::make_shared< EstimatableParameterSettings >( "Earth", gravitational_parameter )
};
const auto parametersToEstimate = createParametersToEstimate( parameterNames, bodies );

// Throws: "Error when defining setIntegratedResult, dynamics type not recognized"
OrbitDeterminationManager< double, double > orbitDeterminationManager(
        bodies,
        parametersToEstimate,
        std::vector< std::shared_ptr< observation_models::ObservationModelSettings > >( ),
        nullptr );
```

### OS and Python version:

OS: macOS ARM

### Expected behavior:

_No response_

### Current behavior:

_No response_

### Additional context:

_No response_

Markus-Reichel:
Allow OrbitDeterminationManager to run against a fixed trajectory, estimating only parameters that have no dynamical state, by passing null propagator settings.

Fixes #1006


src/tudatpy/estimation/estimation_analysis/expose_estimation_analysis_estimator.cpp
```diff
@@ -60,7 +60,7 @@ void expose_estimation_analysis_estimator( py::module& m )
                   py::arg( "bodies" ),
                   py::arg( "estimated_parameters" ),
                   py::arg( "observation_settings" ),
-                  py::arg( "propagator_settings" ),
+                  py::arg_v( "propagator_settings", std::shared_ptr< tp::PropagatorSettings< STATE_SCALAR_TYPE > >( ), "None" ),
```
DominicDirkx: default should not be None

tests/test_tudat/src/astro/orbit_determination/unitTestEstimationInputWeightsAndConvergence.cpp
```diff
@@ -29,6 +32,127 @@ namespace unit_tests
 {
 BOOST_AUTO_TEST_SUITE( test_estimation_input_output )
 
+BOOST_AUTO_TEST_CASE( test_ObservationOnlyEstimatorConstruction )
```
DominicDirkx: add comment to test case and (group of) test statements to explain the overall test and the specific test statements

tests/test_tudat/src/astro/orbit_determination/unitTestEstimationInputWeightsAndConvergence.cpp
```diff
@@ -29,6 +32,127 @@ namespace unit_tests
 {
 BOOST_AUTO_TEST_SUITE( test_estimation_input_output )
```
DominicDirkx, line 34: I don't think all these separate test cases are needed. Some can be combined

tests/test_tudat/src/astro/orbit_determination/unitTestEstimationInputWeightsAndConvergence.cpp
DominicDirkx: This test step can be combined with others, this test case is redundant

tests/test_tudat/src/astro/orbit_determination/unitTestEstimationInputWeightsAndConvergence.cpp
```diff
@@ -29,6 +32,127 @@ namespace unit_tests
...
+    std::shared_ptr< CombinedStateTransitionAndSensitivityMatrixInterface > interfaceFromManager =
+            orbitDeterminationManager.getStateTransitionAndSensitivityMatrixInterface( );
+    BOOST_REQUIRE( interfaceFromManager != nullptr );
+    BOOST_CHECK_EQUAL( interfaceFromManager->getStateTransitionMatrixSize( ), 0 );
+    BOOST_CHECK_EQUAL( interfaceFromManager->getSensitivityMatrixSize( ), 1 );
+    BOOST_CHECK_EQUAL( interfaceFromManager->getFullParameterVectorSize( ), 1 );
+    BOOST_CHECK( orbitDeterminationManager.getVariationalEquationsSolver( ) == nullptr );
+}
+
+BOOST_AUTO_TEST_CASE( test_ObservationOnlyEstimatorRejectsInitialStateParameters )
```
DominicDirkx: This and the next test can be combined in a loop.

tests/test_tudat/src/astro/orbit_determination/unitTestEstimationInputWeightsAndConvergence.cpp
```diff
@@ -29,6 +32,127 @@ namespace unit_tests
...
+    // confirms the rejection above is specific to the time bias, not to estimating biases as such.
+    const std::vector< std::shared_ptr< EstimatableParameterSettings > > additiveBiasParameterNames = {
+        std::make_shared< ConstantObservationBiasEstimatableParameterSettings >( linkEnds, one_way_range, true )
+    };
+    const std::shared_ptr< EstimatableParameterSet< double > > additiveBiasParameters =
+            createParametersToEstimate< double, double >( additiveBiasParameterNames, bodies );
+    BOOST_REQUIRE_EQUAL( additiveBiasParameters->getEstimatedParameterSetSize( ), 1 );
+    BOOST_CHECK_NO_THROW( ( OrbitDeterminationManager< double, double >(
+            bodies, additiveBiasParameters, observationSettingsList, nullPropagatorSettings ) ) );
+}
```
DominicDirkx, line 149: There is no test of the full functionality in an estimation, this is an important step. Also the values in the ObservationOnlyStateTransitionAndSensitivityMatrixInterface should be tested

PR #1012, opened 2026-09-17
CI compiler warnings as errors

### LINKED ISSUE #967: Treat compiler warnings as errors
At the moment compiler warnings are simply ignored by the default build configuration. This makes it quite tedious to build from develop, as the compiler output may be obscured and diluted with new warnings when pulling and rebuilding. It also takes regular attention from maintainers at the project level instead of the PR level (e.g. #772, #775, #848, #960)
If not enabled by default in the CMake configuration, I think it should at least be enabled in the build and test GH action workflow, such that no new warnings are introduced at PR level.
Thoughts @tudat-team/dev-ops?

larshinueber:
Fixes #967.

This adds a flag to the C++ compilation step of the GH action build and test workflow to treat compiler warnings as errors.
It only affects CI runs and should reduce the chances of warnings in local builds from develop.


DominicDirkx:
@larshinueber The problem is that this terminates the build once it finds an error, which could slow down development *a lot* since every next possible warning will not occur until the previous wanring is corrected and the build retriggered.

larshinueber:
@DominicDirkx I think this should be fixed by the `- k` flag, since the build would continue with independent targets but still be marked as failed in CI.
From the ninja docs: "-k N     keep going until N jobs fail (0 means infinity) [default=1]"

I would also expect this to be a smaller issue once established (similar to the pre-commit thing) since the build output should be without any warnings from the beginning, and only newly added code could potentially introduce warnings, which would (ideally) already be resolved locally.
That being said, I'm happy to discuss this on the project-level with @tudat-team/dev-ops (either in #967 or one of the next meetings) :)

Issue #1008, opened 2026-09-15
Issues in the Mars High Accuracy Rotation Model


JelteB:
Hi again 👋,

I bring more comments I have encountered during my Tudat adventures, for you to deal with at some point...

Firstly, the baked-in `mars_high_accuracy` is listed as using Konopliv et al. (2016) parameters, while this is true for most parameters, it is NOT for the default polar motion coefficients. These are not available from Konopliv until the 2020 paper. After a long time searching I think you guys took these values from Van den Acker et al., 2002 paper: should be stated in the documentation.

Secondly, the parameters taken from Van den Acker et al., 2002, contain a small typo in the code (line 359 under tudatpy/src/tudat/simulation/environment_setup/defaultBodies.cpp):
in `yPolarMotionCoefficients[ 2.0 ] = std::make_pair( 3.9 * milliArcSecondToRadian * std::sin( convertDegreesToRadians( 172.5 ) ), 3.9 * milliArcSecondToRadian * std::cos( convertDegreesToRadians( 118.7 ) ) );`, BOTH the angles should be 172.5: I think the 118.7 angle is a copy-paste typo from the `yPolarMotionCoefficients[ 1.0 ]`.

Thirdly, as a feature request it would be nice if some of the other parameters from Konopliv's rotation model e.g., `angleIAtEpoch`, `angleIRateAtEpoch`, `anglePhiAtEpoch`, `anglePhiRateAtEpoch`, `coreFactor `and `freeCoreNutationRate` would also be fully custom in `mars_high_accuracy_full_custom` as these estimates do also vary from Konopliv's parameters in papers from 2006, 2016 and 2020.

PR #1015, opened 2026-09-20
Fix y polar-motion semi-annual cosine phase and state its source

shaneraphel:
Related to #1008 (parts 1 and 2).

The y[2.0] polar-motion term paired sin(172.5 deg) with cos(118.7 deg). Every other nonzero term (x[1.0], x[2.0], y[1.0]) uses one shared phase across its sine/cosine pair, and 118.7 deg is the y[1.0] phase, so this reads as a copy slip. This gives the y[2.0] cosine the 172.5 deg phase. At 3.9 mas amplitude the cosine coefficient moves from -1.87 mas to -3.87 mas, a shift of about 2.0 mas:

```
python3 -c "import math; print(3.9*(math.cos(math.radians(118.7))-math.cos(math.radians(172.5))))"
```

The polar-motion defaults are Van den Acker et al. (2002), not Konopliv et al. (2016) as the surrounding documentation implies, so the block comment now says that. Konopliv 2016 carries no polar motion (per the issue, those values only appear from the 2020 paper).

Part 3 of the issue (more full-custom angles) is intentionally left out; that is a feature request for the maintainers.


DominicDirkx:
We appreciate the PR, but we will give the whole mars high-accuracy rotation model a thorough review and prefer to handle all points of the issue in one go.

shaneraphel:
tanks for review!

PR #1021, opened 2026-09-21
fix/Increase Rosetta RSI mapping timeout.

Markus-Reichel:
Increase timeout for Rosetta mission data downloader.


DominicDirkx:
Overall I think we should leave the mission data downloader along until we can thoroughly revise it, but this seems like a harmless change @luigigisolfi  what do you think?

luigigisolfi:
> Overall I think we should leave the mission data downloader along until we can thoroughly revise it, but this seems like a harmless change @luigigisolfi what do you think?

I think so, looks like we can merge it.

Issue #1023, opened 2026-09-22
[F&E]: Additional properties retrievable from the SBDB


mrevellino:
### Description:

Adding the possibility of retrieving cometary non-gravs parameters (A1, A2, A3 and Dt) from the SBDB; also add the time of first and last observations, time of perihelion and perihelion distance

### Suggestions for implementation:

Add additional properties to the SBDB class

PR #1024, opened 2026-09-22
Feature/sbdb additions

mrevellino:
Addition of following properties to the SBDBquery class: 
- nongrav_params
- Dt
- first_obs
- last_obs
- perihelion
- time_perihelion

Addresses #1023


DominicDirkx:
@mrevellino Good to have these additions :) Could you add a very small unit test (calling the new functions and checking if the values returned are as expected)?

mrevellino:
@DominicDirkx Unit tests should be there now :) 
Let me know if they are ok (not super sure how things are normally done)!

DominicDirkx:
Looks good to me :) @larshinueber or @luigigisolfi Could you give this a quick lookover?

DominicDirkx:
One point (which we'll handle separately from this PR) is making sure this code also makes its way into the refactored setup in #900  (where the module it is in will change)

src/tudatpy/data/sbdb/sbdb.py
```diff
@@ -106,7 +109,87 @@ def diameter(self):
...
+            raise ValueError(f"Diameter is not available for object {self.name}")
+
+    @property
+    def nongrav_params(self):
+        """Returns cometary non-gravitational model parameters for the small body in [m/s^2].
+        If one or more parameter is unavailable a corresponding value of 0 is returned in the array
+        """
+        try:
+            A1 = (
+                (self.query["orbit"]["model_pars"]["A1"].value * u.au / u.day**2)
```
larshinueber: Since `query["orbit"]["model_pars"]["A1"]` is already a astropy quantity, couldn't you just directly cast it to SI units (instead of retrieving the quantity's value, multiplying it by the unit and then converting it to SI).
(Same in other places)

src/tudatpy/data/sbdb/sbdb.py
```diff
@@ -106,7 +109,87 @@ def diameter(self):
...
+                (DT * u.day).to(u.s).value
+            )  # a positive Dt will need to evaluate the position at (t-Dt)
+        except Exception as _:
+            raise ValueError(f"Asymmetry parameter DT is not available for object {self.name}")
+
+    @property
+    def first_obs(self):
+        """If available, returns date of the first observation used for the orbit estimation of the small body in [JD]"""
+        try:
+            first_obs = [int(el) for el in self.query["orbit"]["first_obs"].split("-")]
```
larshinueber: Minor stylistic comment, but instead of splitting manually and then assembling the datetime, you can also parse it directly into a datetime object using something like
```python
datetime.strptime(first_obs, "%Y-%m-%d")
```

src/tudatpy/data/sbdb/sbdb.py
```diff
@@ -106,7 +109,87 @@ def diameter(self):
...
+        try:
+            DT = self.query["orbit"]["model_pars"]["DT"].value
+            return (
+                (DT * u.day).to(u.s).value
+            )  # a positive Dt will need to evaluate the position at (t-Dt)
+        except Exception as _:
+            raise ValueError(f"Asymmetry parameter DT is not available for object {self.name}")
+
+    @property
+    def first_obs(self):
```
larshinueber, line 144: I'm not familiar with the conventions here, but is there a particular relevance for the date to be in Julian Days instead of ephemeris time/seconds since J2000 as is commonly used in Tudat?

src/tudatpy/data/sbdb/sbdb.py
```diff
@@ -106,7 +109,87 @@ def diameter(self):
...
+        """If available, returns date of the first observation used for the orbit estimation of the small body in [JD]"""
+        try:
+            first_obs = [int(el) for el in self.query["orbit"]["first_obs"].split("-")]
+            observation_start = datetime.datetime(first_obs[0], first_obs[1], first_obs[2])
+            return Time(observation_start).jd
```
larshinueber: I think we're not fully consistent with this in other modules, but you can also do the conversion using the `tudatpy.astro.time_representation.DateTime` class:

```python
DateTime.from_python_datetime(observation_start).to_julian_day()
```

larshinueber (commented):
@mrevellino just a few minor comments, besides the astropy quantity conversion it's only stylistic and not blocking from my side! Nice :)

mrevellino:
Thanks for the comments @larshinueber :)
I should have implemented all of them now

larshinueber (approved):
Nice, looks all good to me :)

DominicDirkx:
Nice! Congratulations on your first PR @mrevellino :)

Issue #1033, opened 2026-09-23
[F&E]: Add support for radar data and space astrometry


DominicDirkx:
### Description:

The current MPC interface throws away radar data and optical observations from space astrometry. Support for these data should be added. Similarly, support for the radar data from JPL should be added.

DominicDirkx:
@mrevellino I mentioned this a couple of times, and thought it was already in #900, but apparently I had forgotten to add it... I'm making the branch I have up-to-date with the current state of the tracking-data refactor and will open a PR

PR #1034, opened 2026-09-23
Support for radar data and space astrometry

DominicDirkx:
Closes #1033


DominicDirkx:
@mrevellino Good to be aware of :) I did a local merge of the latest feature/data-refactor so I probably messed things up a little. The CI will show which builds/tests fail. Some of the code may need a little cleaning, but it's already working well (I think).

DominicDirkx:
I think when creating this PR, I worked from an old version of my implementation. Will come back to it this afternoon

src/tudatpy/data_input/tracking_data/jpl_radar/jpl_radar.py
```diff
@@ -0,0 +1,179 @@
...
+from tudatpy.astro import time_representation
+from tudatpy.data_input.tracking_data.radar_utilities import (
+    empty_radar_table,
+    filter_radar_data,
+    radar_data_from_raw,
+    radar_data_to_tracking_data,
+)
+
+_API_URL = "https://ssd-api.jpl.nasa.gov/sb_radar.api"
+_JPL_TO_MPC_STATION = {
```
DominicDirkx: Check issue that MPC and JPL station positions are slightly different

src/tudatpy/data_input/tracking_data/jpl_radar/jpl_radar.py
```diff
@@ -0,0 +1,179 @@
...
+    "-1": "251",
+    "-2": "254",
+    "-9": "256",
+    "-13": "252",
+    "-14": "253",
+    "-25": "257",
+    "-38": "255",
+    "-73": "259",
+}
+_ALTITUDE_UNITS = {"m": 1.0, "km": 1.0e3, "ft": 0.3048}
```
DominicDirkx: Unlikely that these units are not implemented in some package. do not hardcode the conversaion values here

src/tudatpy/data_input/tracking_data/mpc/mpc.py
```diff
@@ -432,39 +488,248 @@ def __add__(self, other):
...
+                edge_order=2 if len(epochs) > 2 else 1,
+            )
+        else:
+            velocities = np.zeros_like(positions)
+
+        states = np.hstack((positions, velocities))
+        return {float(epoch): state for epoch, state in zip(epochs, states)}
+
+    @staticmethod
+    def satellite_state_history_to_ephemeris_settings(
```
DominicDirkx: This must not be here. The data_input must be completely independent of anything in tudatpy.dynamics or tudatpy.estimation (apply to full PR)

src/tudatpy/data_input/tracking_data/obs_80_cols/parsers.py
```diff
@@ -1,10 +1,154 @@
+import os
 import re
```
DominicDirkx: Determine why all these changes to this file are needed.Do not change things, but explain why it was done

tests/test_tudatpy/conftest.py
```diff
@@ -12,6 +12,58 @@
...
 
+def _implicit_decimal_field(value, integer_width, fraction_width=4, signed=False):
+    """Return a fixed-width MPC field with an implied decimal point."""
+    if value is None:
+        return " " * (integer_width + fraction_width)
+    width = integer_width + fraction_width + 1
+    return f"{value:{'+' if signed else ''}0{width}.{fraction_width}f}".replace(".", "")
+
+
+def _mpc_radar_pair(
```
DominicDirkx: conftest seems like the wrong place for this.

src/tudatpy/data_input/tracking_data/jpl_radar/jpl_radar.py
DominicDirkx: I checked the JPL coordinates against `ground_station.optical_telescope_stations()` using the WGS84 ellipsoid. The horizontal/vertical differences were: 251: 95.7 m / -53.9 m; 252: 38.4 m / -4.7 m; 253: 1.6 m / -13.8 m; 254: 47.2 m / -42.0 m. Since several exceed 5 m, I kept a single station source in the observation setup (the MPC catalogue) instead of silently overriding selected stations with a second hard-coded table. The JPL coordinates remain available through `station_geodetic_positions()`. The appropriate follow-up is to correct the affected entries in the upstream `mpc.sit` resource.

src/tudatpy/data_input/tracking_data/jpl_radar/jpl_radar.py
DominicDirkx: Fixed in c98f3c398a. Altitudes are now converted with `astropy.units.Unit(...).to_value(u.m)` instead of a hard-coded conversion dictionary. The mocked API response now also exercises the real `km` unit returned by the JPL endpoint.

src/tudatpy/data_input/tracking_data/mpc/mpc.py
DominicDirkx: Fixed in c98f3c398a. I removed the ephemeris-settings helpers and all `tudatpy.dynamics`/`tudatpy.estimation` imports from `src/tudatpy/data_input`. Passive-reflector model setup now lives on the test/observation-model side. A source-wide import check under `src/tudatpy/data_input` is clean.

src/tudatpy/data_input/tracking_data/obs_80_cols/parsers.py
DominicDirkx: The parser changes are needed for the two new MPC record families while preserving the existing optical output: (1) Astroquery can concatenate paired `S`/`s` 80-column records, so they must first be split; (2) `R`/`r` radar pairs must be extracted before optical-field validation, with delay and Doppler read from their distinct fixed columns and radar epochs rounded to the integral second required by the MPC format; (3) `s` parallax records use type-dependent km/AU scaling and may contain spaces between signs and digits, then must be joined to the preceding `S` observation; and (4) radar-only input must be valid even though it has no optical rows, so the canonical radar table is carried in `Table.meta`. The optical-output construction was moved to a helper to keep this mixed-record path separate; its data model is unchanged.

tests/test_tudatpy/conftest.py
DominicDirkx: Fixed in c98f3c398a. The synthetic MPC radar-pair factory and fixture are now local to `test_data_radar.py`; `conftest.py` is left for genuinely shared test infrastructure.

PR #1036, opened 2026-09-23
Windows compilation instructions

larshinueber:
Draft PR to add instructions how compile tudatpy locally on Windows.

Changes include:
- match CI configuration in build.py script by explicitly stating `cl` compiler
- Add installation of VS 2022, v17 in pre-requisites for Windows users
- Remove obscure characters in docstrings since this would give errors in the stub generation process, specify utf-8 encoding explicitly in stub reading/writing


DominicDirkx:
@larshinueber Nice! Great (and surprising!) that it wasn't that hard. From what I understand this can come out of draft and be 'ready for review' right?

larshinueber:
@DominicDirkx Yes, I left it as draft because I would like someone on Windows to go through the instructions and try compiling before merging, if possible 👍

DominicDirkx:
@larshinueber I think that's a good idea, but I don't think this needs to delay merging this. The changes all look good, and if someone finds an issue with the Windows build (and at some point: someone will!) we can open a new issue.

